PR #17, opened 2019-11-21
feat(dom): add support for animation frames

AdeAttwood:
This adds animation frame support adding `requestAnimationFrame` and `cancelAnimationFrame` functions to the window

ref #9

### DIFF
--- a/src/Window.ts
+++ b/src/Window.ts
@@ -105,4 +105,20 @@ export default class Window extends EventTarget {
 	 * Disposes the window.
 	 */
 	public dispose(): void {}
+
+	/**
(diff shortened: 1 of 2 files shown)


capricorn86:
Thanks for your contribution @AdeAttwood! :)

I made some minor changes so that it works in AsyncWindow as well. Also I changed so that ID of the timeout is returned by requestAnimationFrame(), so that it can be canceled by cancelAnimationFrame().

Your change is available in v.0.7.0 of happy-dom and v.0.0.28 of jest-environment-happy-dom.

PR #30, opened 2020-01-27
Reduce possible namespace conflicts

DaSchTour:
I ran into a problem while trying to integrate this library into angular SSR. There are compile errors because Response object is taken from fetch namespace and not node fetch as DOM library needs to be available for typescript compilation. Changing this to explizit named imports avoids naming conflicts.

### DIFF
--- a/src/AsyncWindow.ts
+++ b/src/AsyncWindow.ts
@@ -1,4 +1,4 @@
-import NodeFetch from 'node-fetch';
+import { Response, fetch } from 'node-fetch';
 import Window from './Window';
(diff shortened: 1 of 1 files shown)


capricorn86:
Hi @DaSchTour!

Thank you for providing a fix. I hope you get Angular to work with server side rendering.

Your fix is available in v0.10.7 of happy-dom and v0.0.38 of jest-environment-happy-dom.

DaSchTour:
@capricorn86 not completly yet. I have some problems small issues with figuring out where exactly to attach happy-dom in SSR to not conflict with @angular/universal. But I'm working on it.

PR #39, opened 2020-05-09
sets target on event when dispatched

allain:
event.target was always null, but I think it ought to be the EventTarget itself.

Rather than just assign target to the EventTarget on the event, I'm creating a new object using Object.create and assigning target to it.

I'm not sure what the spec says should be done in this case but it felt like modifying the passed argument was not the right thing to do.

If you'd rather I just change the incoming event let me know and I'll strip out the cloning logic (and test modifications for it).

### DIFF
--- a/src/event/EventTarget.ts
+++ b/src/event/EventTarget.ts
@@ -42,8 +42,12 @@ export default abstract class EventTarget {
 		let returnValue = true;
 
 		if (this._listeners[event.type]) {
+			const newTarget = Object.create(event)
+			// @ts-ignore internally the target can be set but it cannot be modified in user code
+			newTarget.target = this
+
 			for (const listener of this._listeners[event.type]) {
-				listener(event);
+				listener(newTarget);
 				if (event.cancelable && event.defaultPrevented) {
 					returnValue = false;
 				}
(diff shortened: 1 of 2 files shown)


src/event/EventTarget.ts
```diff
@@ -42,8 +42,12 @@ export default abstract class EventTarget {
 		let returnValue = true;
 
 		if (this._listeners[event.type]) {
+			const newTarget = Object.create(event)
```
capricorn86, line 45: It does not feel good to manipulate the target property in the event object here, but that seem to be what the browser is doing. In order to simulate what the browser is doing we probably need to do the same thing here.

Suggestion:
event.target = this;

capricorn86, line 45: It does not feel good to manipulate the target property in the event object here, but that seem to be what the browser is doing. In order to simulate what the browser is doing we probably need to do the same thing here.

Suggestion:
event.target = this;

capricorn86:
Nice to see a PR from you @allain :+1: 

I think we need to go with your alternative suggestion even if it does not feel right as that seem to be what the browser is doing.

allain:
I've changed it to modify the event directly. Tests have been reverted.

capricorn86:
Thanks @allain for contributing!

Your fix is available in v0.14.1 of happy-dom and v0.0.43 of jest-environment-happy-dom.

PR #40, opened 2020-05-14
Add basic support for dataset

mat3e:
Minimal version of [`dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset) - no support for changing names from camelCase to dash-style and so on.


src/nodes/basic-types/element/Element.ts
```diff
@@ -214,6 +214,23 @@ export default class Element extends Node {
...
 
+	/**
+	 * Read-only property providing access to all the custom data attributes (data-*).
+	 *
+	 * Basic version - no support for converting names (dash-style to camelCase and vice versa).
+	 */
+	public get dataset(): { [key: string]: string } {
+		return (new Proxy(this, {
+			set(target: Element, p: PropertyKey, value: unknown): boolean {
+				target.setAttribute(`data-${p.toString()}`, value.toString());
```
capricorn86, line 225: I believe the key is transformed to kebab case before it is used as an attribute.

**Set dataset value:**
element.dataset.myDataAttribute = 'value';

**Get attribute:**
element.getAttribute('data-my-data-attribute');

**Camel case to kebab case function:**
private camelToKebabCase(text: string): string {
	return text.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
}

Issue #221, opened 2021-02-16
Support for insertAdjecentText


kr8n3r:
Hi, 
looking to replace jsdom so I'm looking to add either just the `insertAdjecentText` method or the entire [insert-adjecent](https://dom.spec.whatwg.org/#insert-adjacent) suite of methods, depending on projects philosophy.

Looking at the code structure I wasn't quite sure how to split the methods. 
Guidance would be appreciated so I can put together a PR.

Thanks.

capricorn86:
Hi @kr8n3r!

Nice to hear that you would like to contribute :slightly_smiling_face: 

Both separating the features as smaller PR:s or doing everything in a bigger one works fine by me.

There is a guide for how to install, compile, test and what formats commits should have etc. here:
https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md

Happy DOM uses interfaces to ensure that the specs are followed. The interfaces are also used to prevent circular dependencies etc. that can occur due to the nature of how the DOM works.

All three methods seem to be part of the Element interface.

I would do the following:

1. Add the methods according to the specs to the IElement interface located at packages/happy-dom/src/nodes/element/IElement.ts. Whenever a Node or Element is used as an attribute or return type, use its interface as type and not the class implementing the interface.
2. Implement the methods in Element.ts that is located in the same directory as IElement. Same here, use interfaces for attributes and return types when it is a Node or Element. Both childNodes and children need to be updated if you add or remove nodes. You also have to update the parent node if nodes are moved. Therefore it is probably best to re-use methods like insertBefore() or appendChild() as they already have this logic.
3. Create unit tests for the implemented methods in packages/happy-dom/test/nodes/element/Element.test.ts

Let me know if something is still unclear or if you have any more questions! :slightly_smiling_face:

capricorn86:
@kr8n3r if you need to parse HTML you can look at ChildNodeUtility.after() as an example where it is using DOMParser.

capricorn86:
@kr8n3r do you still want to make a contribution? Otherwise I can do it as well :slightly_frowning_face:

dkocsis-emarsys:
Hello again!

I wanted to report this similar issue, when i saw that it's already here.
I'm not that experienced with typescript but I've just started to make a contribution on this issue, because it seems pretty easy to implement. And this is where i found another issue in insertBefore's implementation. Should i create another ticket on it?

[...]

PR #228, opened 2021-03-12
Task/221 implement insert adjacent functions

dkocsis-emarsys:
Hi!

I've finished with the changes related to the insertAdjacent functionalities and the missing implementation if insertBefore's null parameter handling.
If there's anything i should change, just tell me.
As for the solution, i basically went with a polyfill like solution, because it looked just fine to add this new functionality without write huge chunks of code. That's why i had to fix insertBefore, it depends on that missing behaviour.


packages/happy-dom/src/nodes/node/INode.ts
```diff
@@ -21,7 +22,10 @@ export default interface INode extends IEventTarget {
 	cloneNode(deep: boolean): INode;
 	appendChild(node: INode): INode;
 	removeChild(node: INode): void;
-	insertBefore(newNode: INode, referenceNode?: INode): INode;
+	insertBefore(newNode: INode, referenceNode?: INode | null): INode;
+	insertAdjacentElement(position: TInsertAdjacentPositions, node: INode): INode | null;
```
capricorn86, line 26: insertAdjacentElement(), insertAdjacentHTML(), insertAdjacentText() is part of the IElement interface and should be implemented in the Element class.

packages/happy-dom/src/nodes/node/Node.ts
```diff
@@ -351,6 +359,60 @@ export default class Node extends EventTarget implements INode {
...
 	}
 
+	/**
+	 * Inserts a node to the given position.
+	 *
+	 * @param position Position to insert element.
+	 * @param element Node to insert.
+	 * @return Inserted node or null if couldn't insert.
+	 */
+	public insertAdjacentElement(position: TInsertAdjacentPositions, element: INode): INode | null {
```
capricorn86, line 369: insertAdjacentElement(), insertAdjacentHTML(), insertAdjacentText() is part of the IElement interface and should be implemented in the Element class.

packages/happy-dom/src/nodes/node/Node.ts
```diff
@@ -351,6 +359,60 @@ export default class Node extends EventTarget implements INode {
...
+	 * Inserts an HTML string to the given position.
+	 *
+	 * @param position Position to insert text.
+	 * @param text HTML string to insert.
+	 * @return Inserted node or null if couldn't insert.
+	 */
+	public insertAdjacentHTML(position: TInsertAdjacentPositions, text: string): void {
+		const templateElement = <IHTMLTemplateElement>this.ownerDocument.createElement('template');
+		templateElement.innerHTML = text;
+		this.insertAdjacentElement(position, templateElement.content);
```
capricorn86, line 401:
```suggestion
		for(const node of XMLParser.parse(text).childNodes.slice()) {
		      this.insertAdjacentElement(position, node);
		}
```
I think we can use the XMLParser directly here :slightly_smiling_face: 

Note that you need to prepend the commit with "#228@minor:" if you apply the suggestion as well.

capricorn86 (changes requested):
Looks really good and thank you for contributing! :+1: 

I just have some minor comments :slightly_smiling_face:

capricorn86:
The tests where really good by the way :+1:

dkocsis-emarsys:
Thanks for the review, I've made the requested changes.
I've used #221 for moving the things to the Element class and #228 for the suggestion.

capricorn86 (approved):
Awesome work! :+1:

Issue #244, opened 2021-04-10
Add support for `Document.adoptNode` method.


YevhenKap:

capricorn86:
Thank you for your contribution @YevhenKap! :slightly_smiling_face: 

I have merged the PR now, so I will close this issue.

capricorn86:
@YevhenKap I found the Typescript errors. They have been fixed now.

The reason for the errors was because Document, Element, Template and DocumentFragment did not return the removed Node, but the interface stated that they should.

YevhenKap:
@capricorn86 I got it. Sorry, my mistake :(

Issue #243, opened 2021-04-10
`removeChild` method should return removed node


YevhenKap:
According to [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Node/removeChild) `removeChild` method [should not return `void`.](https://github.com/capricorn86/happy-dom/blob/7cfdfa08fd738ac367946bcc37a80e5b1a8c0184/packages/happy-dom/src/nodes/node/INode.ts#L23)

capricorn86:
Thank you for your contribution @YevhenKap! slightly_smiling_face

I have merged the PR now, so I will close this issue.

PR #245, opened 2021-04-10
Adding support for `Document.adoptNode` method.

YevhenKap:
Hello. At first, thank you for your great work!
I noticed that the `Document` object does not have an `adoptNode` method, so I tried to add it. Also, there is an inconsistency in the `Node.removeChild` method - it should return the removed node.

- [Issue #244]
- [Issue #243 ]

> When I tried to compile the project I got a lot of TS errors. Thus I had no possibility to test it :( So I do not ask for a merging of the pull request, but a review.


capricorn86 (approved):
Great work! :slightly_smiling_face: 

I have tried to compile and test locally and I can't see any Typescript errors.

PR #322, opened 2021-12-20
#321@patch: `tagName` can be null.

### LINKED ISSUE #321: `tagName` can be null
`tagName` can be `null`, we need to modify the logic on `Element.ts` since it is causing problems to test web components using `lit` on `vitest` repo

userquin:
see https://github.com/vitest-dev/vitest/pull/234

Maybe it is also a problem on `lit` side.

closes #321


davidortnercybercom (approved):
Nice! :star2:

capricorn86 (approved):
Nice! :star2:

Issue #330, opened 2022-01-14
`document.location` is undefined


Ayc0:
According to the spec https://developer.mozilla.org/en-US/docs/Web/API/Location, `location` is supposed to be set to both `document` and `window`.

But in `happy-dom`, it's only set in `window`

Xa3lis:
I've the same problem

capricorn86:
Thank you for your contribution @Ayc0! :slightly_smiling_face: 

It has been merged and released now.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v2.26.0

PR #331, opened 2022-01-18
#330@minor: Add document.location.

Ayc0:
Fixes https://github.com/capricorn86/happy-dom/issues/330

I used a getter instead of a static property for 3 reasons:
- we want to have the exact same object as `window.location` so that changes made in one will also impact the other (so I didn't use `new Location()`)
- the document has a view `this._defaultView` that can be updated, so `document.location` should always be up-to-date with the linked view
- it won't be overridable 

I also added a test that passes `npm test`

image

--- 

Note: this is my first PR to the project, so I'm not sure I followed every guideline


capricorn86 (approved):
Nice! :star2:

Issue #328, opened 2022-01-14
ClassList doesn't include `toggle`


Ayc0:
When using `happy-dom` with `classlist-polyfill`, I have this error:

```
TypeError: testElement.classList.toggle is not a function
```

When looking at https://github.com/capricorn86/happy-dom/blob/ab971406111805e8d940419acc6500c601eb0aa2/packages/happy-dom/src/nodes/element/ClassList.ts, it doesn't seem to have `toggle` defined in it

capricorn86:
Thank you for your contribution @Ayc0! slightly_smiling_face

It has been merged and released now.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v2.27.0

PR #332, opened 2022-01-18
#328@minor: Add toggle() to ClassList.

Ayc0:
Fixes https://github.com/capricorn86/happy-dom/issues/328

Implements `toggle()` on `ClassList` (see https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle)

image


capricorn86 (approved):
Nice! :star2:

PR #333, opened 2022-01-18
fix: HTMLStyleElement should cache the created CSSStyleSheet

sastan:
Without this change every time one calls `styleElement.sheet` a new CSSStyleSheet is returned.


capricorn86 (approved):
Nice! :star2:

capricorn86 (changes requested):
I noticed that the commit message is not following the naming convention:
https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md#commit-convention

It has to follow the naming convention in order for automatic release notes to work.

Can you do "git --amend" and change the message to something like "#333@patch: Fixes issue with HTMLStyleElement.sheet returning a new instance each time."?

capricorn86 (approved):

PR #336, opened 2022-01-20
Task/335 fix xml serializer

### LINKED ISSUE #335: Serializer escapes attributes too much
I tried using happy-dom with Vitest and when switching from jsdom I got a failing test:

image
  
image
 
happy-dom is escaping characters in html attributes that are allowed. These is not need to use `encode` from `he`, `escape` is enough.

Demivan:
Fixes #335


capricorn86 (approved):
Nice! :star2:

PR #347, opened 2022-01-31
@fix: remove one event callback at a time

bhvngt:
I faced this issue while testing one of my functionality. 

Current code removes all the event listers. This becomes an issue if multiple event listeners are registered for the same type.

Keeping the [deleteCount](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) as 1 removes one event listener at a time.


capricorn86 (approved):
Pretty bad bug! :sweat_smile: Nice that you could fix it :star2:

Issue #380, opened 2022-02-24
happy-dom classList has no classes


MichelYpma:
When the first example from the doc is used you'll notice missing classes when the classList is logged:
```
import { Window } from 'happy-dom';

const window = new Window();
const document = window.document;

document.body.innerHTML = '<div class="container"></div>';

const container = document.querySelector('.container');
const button = document.createElement('button');

container.appendChild(button);

// Outputs "<div class="container"><button></button></div>"
console.log( container.classList );
// console.log(document.body.innerHTML);
```

This outputs a ClassList instance without any mention of the container class:
```
ClassList {
    _ownerElement: HTMLElement
}
```

I expect an array of class names or an object representing that array.

Mas0nShi:
it seems not implementation for div tag at this time. lol.

capricorn86:
Hi @MichelYpma and @Mas0nShi! 🙂 
The array like part of the interface has nog been implemented yet. Only methods such as add(), remove() etc.

I will look into adding it as soon as possible.

Mas0nShi:
> Hi @MichelYpma and @Mas0nShi! 🙂 The array like part of the interface has not been implemented yet. Only methods such as add(), remove() etc.
> 
> I will look into adding it as soon as possible.

I've already achieved some of that, maybe.🙂

capricorn86:
Hi @MichelYpma! :slightly_smiling_face: 

Me and @Mas0nShi has collaborated on adding full support for Element.classList now.

You can read more about the releases here:
 - https://github.com/capricorn86/happy-dom/releases/tag/v2.42.0
 - https://github.com/capricorn86/happy-dom/releases/tag/v2.43.0

PR #381, opened 2022-02-25
#380@minor: Adds support for DOMTokenList. Adds support for window.globalThis, document.scripts.

Mas0nShi:
solve #380 (maybe)

not impl section: `classList[0]`、`classList.supports('add-scripts')`

I don't think this good enough, so you can modify or even rejection this pr.

test code section:

```javascript
const { Window } = require("./packages/happy-dom");

const window = new Window();

document.body.innerHTML = '<div class="container b a s d x a z"></div>';

container.appendChild(button);

let classList = container.classList;

console.log(classList.value);
classList.add("b");
console.log(classList.value);

console.log(classList.length);
classList.length = 1;
console.log(classList.length);

console.log(classList.contains('b'));

console.log(classList.value);
classList.remove("b");
console.log(classList.value);

console.log(classList.contains('b'));

console.log(classList.value);
for (const [key, value] of classList.entries()) {
	console.log(key, value);
}


for (const key of classList.keys()) {
	console.log(key);
}

for (const value of classList.values()) {
	console.log(value);
}

console.log(classList.item(0));
console.log(classList.item('0'));
console.log(classList.item('a'));


classList.forEach((currentValue, currentIndex, listObj) => {
	console.log(currentValue, currentIndex, listObj);
}, this);


console.log(classList.value);
console.log(classList.replace("a"));
console.log(classList.value);

console.log(classList.supports("add-scripts"));

```

check pass image: 

image


capricorn86:
@Mas0nShi I have noticed that you have done some really nice contributions. If you are planning to do more contributions I could add you as a collaborator to the repository? In that case it will be easier to do pull-requests and work together on a branch.

Mas0nShi:
> @Mas0nShi I have noticed that you have done some really nice contributions. If you are planning to do more contributions I could add you as a collaborator to the repository? In that case it will be easier to do pull-requests and work together on a branch.


Wow, that's great. thank you for your recognition of me.
I may will pull requests in recently more maybe,  because i learning offence and defence about spider now. 
I've been trying my best to make happy-dom portable, but with more implementation, it's frustrating to inevitably introduce dependencies and more performance overhead. 

The more complete the implementation, the lower the efficiency.

capricorn86:
> 

Yes there are some dependencies to Node.js server side modules, which makes it less portable. My initial goal was to split the functionality that is dependent on server side by having two different Window objects.

Is the goal for you to remove dependencies to server side modules, or something else? :slightly_smiling_face:

Mas0nShi:
> > 
> 
> Yes there are some dependencies to Node.js server side modules, which makes it less portable. My initial goal was to split the functionality that is dependent on server side by having two different Window objects.
> 
> Is the goal for you to remove dependencies to server side modules, or something else? 🙂

No,  My ultimate goal is use it to fight the anti-crawler.😂
although dependency can lead to bloated, it seems that you can only sacrifice some performance if you want more implementations.

Maybe I will consider building a version that suits my personal needs.

Anyway, I will try my best.

PR #389, opened 2022-03-03
#388@minor: Adds partial support for navigator.

Mas0nShi:
**I didn't write the test unit**, because I don't know how to verify...
the navigator based on firefox, so PluginArray / MimeTypeArray length is 0.
and navigator has many function not impl now,  I used temporary empty method instead it.


capricorn86 (approved):
Nice job! :star2: 

I will add a unit test for it after the merge.

PR #472, opened 2022-05-10
Indicate new jest default environment in README

silverwind:
`node` is the new default environment since [Jest 27](https://jestjs.io/blog/2021/05/25/jest-27#flipping-defaults).


capricorn86 (approved):
Nice contribution! :star2:

Issue #478, opened 2022-05-17
"is" defined by createElemtent(NS) options is not being serialized.


Lusito:
First of all, thanks for the project!
I couldn't find any existing issue, but searching for "is" is a bit of a lost cause :confused:  

happy-dom:
```ts
const x = document.createElement("img", { is: "lazy-img" })
x.outerHTML // `<img/>`
```

Real browser (chrome and firefox tested):
```ts
const x = document.createElement("img", { is: "lazy-img" })
x.outerHTML // `<img is="lazy-img">`

// Interestingly though:
x.getAttribute("is") // null
x.getAttributeNames() // []
```

From the above, it seems that `is` doesn't count as attribute when specified via options, but it does get serialized as attribute.

I'd say `XMLSerializer._getAttributes` needs to be adjusted, but `options.is` is not stored on the element.
Looking at the code in `createElementNS`, this may only be wanted if the custom-element has not been defined in the registry.

Does this sound reasonable to you?
- Extend the Element class with an optional `is` attribute
- Set that `is` attribute within `createElementNS` if the registry did not contain said element.
- Check for that attribute in `XMLSerializer._getAttributes` and add it to the result

If so, I can create a PR.

**Update:** SVG Elements seem to work differently:

Real browser:
```ts
const x = document.createElement("svg", { is: "lazy-img" })
x.outerHTML // firefox and chrome: `<svg is="lazy-img"></svg>`

const y = document.createElementNS("http://www.w3.org/2000/svg", "svg", { is: "lazy-img" })
y.outerHTML // chrome: `<svg is="lazy-img"></svg>`, firefox: `<svg></svg>`

const z = document.createElementNS("http://www.w3.org/1999/xhtml", "svg", { is: "lazy-svg"});
z.outerHTML // chrome and firefox: `<svg is="lazy-img"></svg>`
```

Yay for cross-browser consistency :angry:  
So it seems, that firefox makes a difference depending on the namespace.

Anyway. It looks like the xmlns is also not serialized (not in a real browser either). I guess I'll just setAttribute on both of these for now in addition, to make sure they get rendered correctly.

capricorn86:
Hi @Lusito! :slightly_smiling_face: 

It would be great with a contribution.

[...]

PR #489, opened 2022-05-28
#478@major: Fixes issues with HTML serialization of the is value.

Lusito:
This change can potentially break tests as the serialized HTML may differ from before (e.g. Element.innerHTML) and therefore it is released as a major version.


Lusito:
Thanks for your work on this library!

Issue #461, opened 2022-04-29
<select /> throws error if value is empty


catc:
If select value is an empty string (ie: user hasn't selected an option yet), happy dom (via jest env) + rtl throws an error:
```tsx
import { render } from '@testing-library/react';

function Foo() {
  const value = '';
  return (
    <div>
      <select name="" id="" value={value}>
        <option value="foo" />
      </select>
    </div>
  );
}

describe('<Foo />', () => {
  it('renders', () => {
    render(<Foo />);
  });
});
```

Error:
<img width="760" alt="Screen Shot 2022-04-29 at 2 18 01 PM" src="https://user-images.githubusercontent.com/5651132/166000889-cec07cf8-7e36-4412-92d3-5e9e147bfbe6.png">

capricorn86:
Thanks to @fatihcure-jf we now have support for HTMLOptGroupElement, HTMLOptGroupElement and HTMLSelectElement.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v5.1.0

PR #495, opened 2022-06-07
#428@minor: Fixes select element undefined options error.

### LINKED ISSUE #428: Missing "options" property on <select> elements
The `options` property on `<select>` elements is missing leading to problems when using happy-dom in vitest (and probably with Vue in general), see https://github.com/vitest-dev/vitest/issues/1013 .

https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/options

fatihcure-jf:
Hi. I tried to implement HTML `select`, `optgroup`, `option` and `HTMLOptionsCollection` to resolve #428 and #461. There could be "bad implementation"s because of  my misunderstanding. If there are any I would gladly change with your rotation.


capricorn86 (approved):
Nice! Great job! :slightly_smiling_face:

Issue #499, opened 2022-06-09
Missing API atob/btoa


airtoxin:
Hello, I was using happy-dom in my project and I got the following error.

````
ReferenceError: btoa is not defined
```

It seems that neither atob/btoa is implemented.

Mas0nShi:
Hello @airtoxin 😀, thank you for your feedback. I will PR before long!

capricorn86:
Great to hear @Mas0nShi! :smile:

capricorn86:
Thanks to @Mas0nShi we now have support for Window.atob() and Window.btoa() 🙂 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v5.3.0

PR #500, opened 2022-06-09
#499@minor: Adds support for btoa and atob.

Mas0nShi:
solve #499


Mas0nShi:
a speacial in browser(e.g. edge):

```javascript
btoa([1, " ", 2])
// 'MSwgLDI='
```

in fact, it's like this:
 ```javascript
const tmp = ([1, " ", 2]).toString();
// '1, ,2'
btoa(tmp)
// 'MSwgLDI='
```

Issue #504, opened 2022-06-13
XMLParser fails for script tag with JS-Code that looks like a start tag


baranor:
I ran into a problem where adding the following HTML code (simplified):
```html
<html>
<head>
    <title>Title</title>
</head>
<body>
  <script type="text/javascript">
    var vars = []
    for (var i=0;i<vars.length;i++) {}
  </script>
</body>
</html>
```
via `document.write(...)` (with script execution) or `document.documentElement.outerHTML = ...` (without script execution).

After some debugging I found the XMLParser is at fault here, as it tries to parse the `<vars` condition in the for-loop as a new tag.

The problem did have a different impact depending on the actual HTML structure:
- in the original code, it somehow caught itself, but the `textContent` of the script tag was most of the HTML document after that point
- in the above example, it allocated memory until nodejs killed it

If I understand the current parser implementation correctly, there is already handling for the similar case of actual tags within the `script`-tag, but it doesn't catch this one.

I originally stumbled upon this in `happy-dom@4.1.0`, but it still happens in `happy-dom@5.2.0`.

Mas0nShi:
Hi, @baranor 😀, HTML parsers(XML parser) have been buggy for a long time... 
e.g. #376
I will try to fix it in my free time. thank you for your feedback.

Mas0nShi:
Hi, @baranor 😀, HTML parsers(XML parser) have been buggy for a long time... 
e.g. #376
I will try to fix it in my free time. thank you for your feedback.

Mas0nShi:
the reason of this bug:

only one `script` tag is matched.
the regex can't correct match `</script>`

```javascript
html
head
title
title
head
body
script
vars.length
body
html
```

https://github.com/capricorn86/happy-dom/blob/8ece2905eea8e5780e66508230689f0043c1b0e8/packages/happy-dom/src/xml-parser/XMLParser.ts#L85-L93

capricorn86:
Thanks to @Mas0nShi we now have a fix in place 🙂 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v5.3.1

PR #505, opened 2022-06-14
#504@patch: Fixes XmlParser error parse with tags.

Mas0nShi:
now we have perfected tag matching again...

solved #504

### DIFF
--- a/packages/happy-dom/src/xml-parser/XMLParser.ts
+++ b/packages/happy-dom/src/xml-parser/XMLParser.ts
@@ -11,7 +11,7 @@ import INode from '../nodes/node/INode';
 import IElement from '../nodes/element/IElement';
 import HTMLLinkElement from '../nodes/html-link-element/HTMLLinkElement';
 
-const MARKUP_REGEXP = /<(\/?)([a-z][-.0-9_a-z]*)\s*([^>]*?)(\/?)>/gi;
+const MARKUP_REGEXP = /<(\/?)([a-z][-.0-9_a-z]*)\s*([^<>]*?)(\/?)>/gi;
 const COMMENT_REGEXP = /<!--(.*?)-->|<([!?])([^>]*)>/gi;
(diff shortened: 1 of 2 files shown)


capricorn86:
I think we need to find a solution where we turn off the parser when a `<script>` start tag has been found until a `</script>` end tag comes.

Mas0nShi:
> I think we need to find a solution where we turn off the parser when a `<script>` start tag has been found until a `</script>` end tag comes.

I find the **MARKUP_REGEXP** maybe wrong? so i added exclusion `<` beacuse we can't input `<` and `>` when we declare tag..

capricorn86:
Yes, this fixes this specific problem, but we still have problems in these kind of scenarios (where <div> is parsed as a start element):
```javascript
<script>
   for(let i = 0; i < 10;i++) {
	    console.log('<div>');
   }
</script>
```

Mas0nShi:
> 

what? I am good running when i tested this.

image

Mas0nShi:
> Yes, this fixes this specific problem, but we still have problems in these kind of scenarios (where `<div>` is parsed as a start element):
> 
> ```js
> <script>
>    for(let i = 0; i < 10;i++) {
> 	    console.log('<div>');
>    }
> </script>
> ```


Maybe the problem you're talking about is #376 solves?
we had declare  `ChildLessElements` to stop parse the next-level tags.

capricorn86:
> > Yes, this fixes this specific problem, but we still have problems in these kind of scenarios (where `<div>` is parsed as a start element):
> > ```js
> > <script>
> >    for(let i = 0; i < 10;i++) {
> > 	    console.log('<div>');
> >    }
> > </script>
> > ```
> 
> Maybe the problem you're talking about is #376 solves? we had declare `ChildLessElements` to stop parse the next-level tags.

Yes, maybe you are right! I will make a test as well, just to be sure 🙂

PR #529, opened 2022-07-04
#501@patch: allow proxy keys besides strings

### LINKED ISSUE #501: Get on a proxy key that is a Symbol causing an error
MicrosoftTeams-image

From looking at the that stack trace, I look at HTMLElement.js and key is a Symbol in this case so can't call .replace on it. Maybe need to check if key is not a string and key toString on it? Not to sure on the spec if it is expected for a proxy get to be triggered off a symbol access,

kalvenschraut:
fixes #501 

Simplest solution was to add the toString method.

I did try to change types and write a wrapper function the proxy handler uses, but major blocking point of that was a type error I couldn't resolve. Mainly `type symbol can not be used as an index` which since version [4.4](https://github.com/Microsoft/TypeScript/issues/24587#issuecomment-890314169) of typescript is completely valid. Couldn't figure out why it wouldn't allow it in this specific repo, as I wasn't having issues with my other ones.

### DIFF
--- a/packages/happy-dom/src/nodes/html-element/DatasetUtility.ts
+++ b/packages/happy-dom/src/nodes/html-element/DatasetUtility.ts
@@ -23,6 +23,8 @@ export default class DatasetUtility {
 	 * @returns Kebab cased string.
 	 */
 	public static camelCaseToKebab(text: string): string {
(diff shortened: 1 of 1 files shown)


capricorn86 (approved):
Nice! :+1: :slightly_smiling_face:

Issue #484, opened 2022-05-23
ReferenceError: Audio is not defined


Archon-:
Looks like support for audio was never implemented. Reporting issue for it and I'll try to add implementation if there's nothing yet.
```
ReferenceError: Audio is not defined
```

Mas0nShi:
Hello, as you say, Audio is not implemented.

rudywaltz:
Hi @Archon- , did you mean audio HTML element? That is using HTMLMediaElement api like video. I start working on the implementation but be honest, I have very limited time, I can share some draft PR, the API is huge and some of need to implement for the first round. Also have some tricky behaviour at least for me. Related issue here https://github.com/capricorn86/happy-dom/issues/475

Archon-:
Hello @rudywaltz 👋 I'm pretty sure it's about [HTMLAudioElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement/Audio). Not sure if that is part of HTMLMediaElement. If you could share your branch maybe I could continue work.

rudywaltz:
> The HTMLMediaElement interface adds to [HTMLElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement) the properties and methods needed to support basic media-related capabilities that are common to audio and video.

> The [HTMLVideoElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement) and [HTMLAudioElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement) elements both inherit this interface.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement

jsolly:
The lack of support for Audio was why I needed to use jsdom.

Issue #475, opened 2022-05-13
Add support for HTMLMediaElement


rudywaltz:
Dear @capricorn86 ,
I just start to be familiar with happy DOM, and I would like to use it for testing (jest) my hobby project which is a native web components base audio-player. My first experience was good, but now I'm struggling to test the **audio** element because it is not implemented here. I try to create my own mock, but I'm not sure how can I overwrite happy-dom on this level, or maybe contribute here somehow. 
Can you help me with this, with the implementation/how can I extend it locally or point to some good reference in your codebase how can I implement it directly here?

thank you for your help

capricorn86:
Hi @rudywaltz! :slightly_smiling_face: 

Any contribution is very welcome!

First off, please read the guidelines for contributions:
https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md

In order to add a new Node/Element, you need to do the following:

1. Go to ./packages/happy-dom. This is where the code for Happy DOM lives.
2. Copy an existing element implementation. E.g. copy `src/nodes/html-meta-element` to `src/nodes/html-media-element`.
3. Rename files, classes and interfaces to follow the new element name (`HTMLMediaElement` and `IHTMLMediaElement`)
4. Update the interface according to the specs at https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement (it is possible to do a partial implementation if there is too much to implement in the first release).
6. Implement the functionality according to the specs in HTMLMediaElement.
7. Add the HTMLMediaElement class as a property to `IWindow.ts` and `Window.ts`.
8. Remove the element class from `NonImplemenetedElementClasses.ts`.
9. Create tests for the implementation under `test/nodes/html-media-element`

PR #535, opened 2022-07-09
475@minor: Adds support for HTMLMediaElement.

rudywaltz:
Resolves https://github.com/capricorn86/happy-dom/issues/484
Resolves https://github.com/capricorn86/happy-dom/issues/475

It's a very basic implementation mostly covering the API (at least what is common between audio and video and widely supported in browsers) `timeRanges` implementation coming from [jsdom](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLMediaElement-impl.js#L7) as the comment mentioned.
I try to implement some basic logic but mostly is missing, I hope this can be good enough as a first step.

References:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement
https://html.spec.whatwg.org/#htmlmediaelement
https://github.com/jsdom/jsdom


Mas0nShi:
although those tags extend from `HTMLMediaElement`, I think we should implement for `HTMLVideoElement` and `HTMLAudioElement` before removing or adding these tags in `ElementTag.ts`.

Of course this is my personal advice, perhaps repo owner's advice is more valuable.

rudywaltz:
Hi @Mas0nShi ,
thanks for the feedback.
`HTMLAudioElement` has a same `API` as `HTMLMediaElement`, only mozzilla-specific API is a difference if I understand correctly the documentation.
https://html.spec.whatwg.org/multipage/media.html#htmlaudioelement
(maybe better doing this here as well? creating an audio element and that is just extending the Media element? 🤔 )

`HTMLVideoElement` has some widely supported properties/attributes that are not coming from `HTMLMediaElement`, so yes maybe better implement that as well before adding or removing it from the not implemented list, but still after this, most of the API working in the test.
https://html.spec.whatwg.org/multipage/media.html#the-video-element

Mas0nShi:
> (maybe better doing this here as well? creating an audio element and that is just extending the Media element? 🤔 )

yes, it's my advices, we can add HTMLMediaElement and HTMLVideoElement though identical or incomplete.

rudywaltz:
> > (maybe better doing this here as well? creating an audio element and that is just extending the Media element? 🤔 )
> 
> yes, it's my advices, we can add HTMLMediaElement and HTMLVideoElement though identical or incomplete.

done

capricorn86:
Hi @Mas0nShi ! 🙂 

Sorry for not looking into this earlier. It has been a lot going on in my private and work life.

I have managed to get the PR build to work and it seems like this PR is failing to build.

I can merge this when the build is successful.

capricorn86:
Hi @Mas0nShi ! 🙂 

Sorry for not looking into this earlier. It has been a lot going on in my private and work life.

I have managed to get the PR build to work and it seems like this PR is failing to build.

I can merge this when the build is successful.

rudywaltz:
Hi @capricorn86 , thank you for taking care of this repo. Can I ask some hint where is the missing part. I see the error locally as well
` Property 'self' in type 'Window' is not assignable to the same property in base type 'IWindow'.` but I have no idea where is the missing part. I try to follow the DialogElement implementation and I don't find any place where the Audio is missing ):

capricorn86:
> Hi @capricorn86 , thank you for taking care of this repo. Can I ask some hint where is the missing part. I see the error locally as well
> 
> ```
>  src/window/Window.ts(256,18): error TS2416: Property 'self' in type 'Window' is not assignable to the same property in base type 'IWindow'.
>   Type 'this' is not assignable to type 'IWindow'.
>     Property 'HTMLAudioElement' is missing in type 'Window' but required in type 'IWindow'.
> ```
> 
> but I have no idea where is the missing part. I try to follow the DialogElement implementation and I don't find any place where the Audio is missing ):

@rudywaltz I found the error. In Window.ts the property `HTMLAudiolement` should be renamed to `HTMLAudioElement`. :slightly_smiling_face: 

It's failing because the interface is using the name `HTMLAudioElement` and it doesn't follow the interface correctly.

capricorn86:
It seems like we got a small merge conflict as well. When that and the property is fixed I will merge it :slightly_smiling_face:

rudywaltz:
With this last "config resolve commit" the history is not too nice, if you need I can try rebasing and creating a new PR with more flat commit history.

capricorn86 (approved):

capricorn86:
@rudywaltz I made a squash 🙂

PR #541, opened 2022-07-13
add blob arrayBuffer()

a-wing:
https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer


capricorn86 (changes requested):
Nice contribution! :slightly_smiling_face: 

Can you please change the commit message, so that it follows the [commit guidelines](https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md#commit-convention)?

E.g.
`#541@minor: Adds support for Blob.arrayBuffer().`

capricorn86 (approved):

Issue #542, opened 2022-07-15
arguments lost in Response constructor


kombucha:
Hi 👋 

I was trying to upgrade from `happy-dom@5.4` to `happy-dom@6.0.3` but my tests started failing.

It seems as if the `status` code is lost:
```js
console.log(new Response(null, { status: 500 }).status)
// outputs 200!
```

Digging further into my `node_modules`, I noticed this in the compiled version of the library:

##### happy-dom@5.4
```js
// ...
/**
 * Fetch response.
 */
class Response extends NodeFetch.Response {
   // 👋 Notice that there's no constructor defined here 👋 

    /**
     * Returns array buffer.
     *
     * @returns Array buffer.
     */
    arrayBuffer() {
// ...
```

##### happy-dom@6.0.3
```js
// ...
/**
 * Fetch response.
 */
class Response extends NodeFetch.Response {
    /**
     * Constructor.
     */
    constructor() {
        // 👋 super is called without the original arguments! 👋 
        super();
        this._ownerDocument = null;
        this._ownerDocument = this.constructor._ownerDocument;
    }
    /**
     * Returns array buffer.
     *
     * @returns Array buffer.
     */
    arrayBuffer() {
// ...
```

~I'm not sure exactly when the regression was introduced, but I think we have our culprit!~

The regression was introduced in https://github.com/capricorn86/happy-dom/pull/525/files#diff-b57fa67149ab8895fb6b9b5873fa262fa20184a91d9d3a3a65a714713d8f9a22

PR #543, opened 2022-07-17
#542@patch: Fix Response constructor issue.

kombucha:
Fixes https://github.com/capricorn86/happy-dom/issues/542


capricorn86 (approved):
Nice work! 😊

PR #545, opened 2022-07-18
fix: update event.target & event.currentTarget via defineProperty.

### LINKED ISSUE #544: Error when using with @testing-library/user-event / @testing-library/dom
I'm trying to use `happy-dom` together with `@testing-library/user-event` or `@testing-library/dom`. When I try to emulate a click (via `fireEvent` or `userEvent.click`, I'm seeing the following error:

````ts
TypeError: Cannot set property currentTarget of #<MouseEvent> which has only a getter
 ❯ GlobalWindow.dispatchEvent node_modules/.pnpm/happy-dom@6.0.4/node_modules/happy-dom/lib/event/EventTarget.js:44:29
 ❯ HTMLDocument.dispatchEvent node_modules/.pnpm/happy-dom@6.0.4/node_modules/happy-dom/lib/nodes/document/Document.js:698:37
 ❯ HTMLElement.dispatchEvent node_modules/.pnpm/happy-dom@6.0.4/node_modules/happy-dom/lib/nodes/node/Node.js:354:40
 ❯ HTMLElement.dispatchEvent node_modules/.pnpm/happy-dom@6.0.4/node_modules/happy-dom/lib/nodes/node/Node.js:354:40
 ❯ HTMLElement.dispatchEvent node_modules/.pnpm/happy-dom@6.0.4/node_modules/happy-dom/lib/nodes/node/Node.js:354:40
 ❯ HTMLElement.dispatchEvent node_modules/.pnpm/happy-dom@6.0.4/node_modules/happy-dom/lib/nodes/node/Node.js:354:40
 ❯ node_modules/.pnpm/@testing-library+user-event@14.2.5_gwcpuyfvwbszhlmedmugzivgzu/node_modules/@testing-library/user-event/dist/index.mjs:1733:33
    1731|     return !defaultPrevented;
    1732|   }
    1733|   return wrapEvent(() => target.dispatchEvent(event), target);
       |                                 ^
    1734| }
    1735|
 ❯ Object.eventWrapper node_modules/.pnpm/@testing-library+dom@8.16.0/node_modules/@testing-library/dom/dist/config.js:27:23
 ❯ wrapEvent node_modules/.pnpm/@testing-library+user-event@14.2.5_gwcpuyfvwbszhlmedmugzivgzu/node_modules/@testing-library/user-event/dist/index.mjs:1707:23
 ❯ dispatchEvent node_modules/.pnpm/@testing-library+user-event@14.2.5_gwcpuyfvwbszhlmedmugzivgzu/node_modules/@testing-library/user-event/dist/index.mjs:1733:10
 ````
 
The reason is that `event.target` and `event.currentTarget` are readonly. I've provided a pr to fix the problem.

### PR DESCRIPTION
**Pull request by benbender**
`event.currentTarget` & `event.target` are read-only per spec. Setting either via property assignment may result in an error. This pr uses `defineProperty` to correctly set them.

Reference: https://dom.spec.whatwg.org/#interface-event

Closes #544 & #343

### CODE REVIEW

File: packages/happy-dom/src/event/EventTarget.ts [hunk 1]
```diff
@@ -47,10 +47,16 @@ export default abstract class EventTarget implements IEventTarget {
...
 		if (!event.target) {
-			event.target = this;
+			Object.defineProperty(event, "target", {
+				value: this,
+				writable: false
+			});
 		}
 
-		event.currentTarget = this;
+		Object.defineProperty(event, "currentTarget", {
```
capricorn86, line 56: I am a bit worried about the performance impact of calling `defineProperty()` on each event and element where it bubbles. Using `defineProperty()` may change the shape of the object, which could take a way optimizations in the V8 engine.

Maybe we can do a solution where we use getters in the `Event` class which returns `_currentTarget` and `_target` that can be modified here?

We can also do some performance testing and see how it affects it.

capricorn86 (changes requested):

packages/happy-dom/src/event/EventTarget.ts
benbender: Hey 👋 ,

I gave your idea about the getters a quick shot and it did not work. The problem is that `jest-testing-library` uses `lib.dom.d.ts` where `target`/`currentTarget` are readonly. So as soon as we are writing directly to them and types are checked at runtime (`vitest`), this will fail :/

Tbh, I didn't even thought about performance in this context. I would be happy to do some microbenchmarks if there is some infrastructure I could use. Do you have anything at hand which could be repurposed?

capricorn86: It is weird that `jest-testing-library` runs the compiler on the happy-dom internal code :thinking: 

Maybe it is possible to do something like this?

`Event.ts`
```javascript
export default class Event {
        ...
	public _currentTarget: IEventTarget = null;
	public _target: IEventTarget = null;

	/**
	 * Returns target.
	 *
	 * @returns Target.
	 */
	public get target(): string {
	      return this._target;
	}

	/**
	 * Returns current target.
	 *
	 * @returns Current target.
	 */
	public get currentTarget(): string {
	      return this._currentTarget;
	}
	
	...
}
```

`EventTarget.ts`
```javascript
export default class EventTarget {
        ...
	public dispatchEvent(event: Event): boolean {
		if (!event.target) {
			event._target = this;
		}

		event._currentTarget = this;

		....
	}
        ...
}
```

benbender: Sadly this won't help as it's a typing issue. So whenever you try to write to `.target`/`.currentTarget`, it will fail because the types of `lib.dom.d.ts` mark those as readonly. This problem is independent of the implementation and the only fix (at least as far as I can see) is to not write directly to those readonly-props in the interfaces. That's also why the `defineProperty()`-approach works...

capricorn86: @benbender I am not sure if you missed the "_" (underline) before the properties. My idea was that we perhaps can write to other properties instead and then return them from getters that are never changed? Then we will not risk any performance penalty of using `Object.defineProperty()`.

I would also like to understand why `vitest` or `jest-testing-library` are checking the types at runtime in the internal code of `happy-dom`. Do you have some example code I can run?

benbender: Sure. Repro: https://github.com/benbender/happy-dom-repro-544

benbender: Related: https://github.com/vitest-dev/vitest/issues/1494

capricorn86:
Hi @benbender! 🙂 

Sorry for not looking in to this PR again earlier. It's been a lot going on in my private and work life.

I will do some tests and look into fixing this as soon as possible.

benbender:
Hey @capricorn86, any reason that you assigned this one to me?

capricorn86:
@benbender no reason except that I have assigned everybody that has created a PR to the PR to make it easier to look through the list of PRs

benbender:
ah, sure. I was confused because you said you wanted to have look and I'm used to assign pr's to the person who's expected to do smth…

capricorn86:
@benbender I have created another PR with a fix that hopefully solves the problem. This branch contains commits that does not follow the [Happy DOM commit format](https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md), so I could not re-use it.

You will probably run into more problems if you let Vitest transpile external libraries. You need to change the transform config to only transform your app source code to avoid this problem.

```javascript
    transformMode: {
      web: [/src\/\.[tj]sx?$/]
    }
```

Pull request:
https://github.com/capricorn86/happy-dom/pull/611

PR #549, opened 2022-07-24
#548@patch: Fix incorrect logic in selectEl.selectedIndex setter.

### LINKED ISSUE #548: happy-dom errors out "Select elements selected index must be valid" even though `selectEl.selectedIndex` is set to a valid index
A symptom of this bug is that vitest tests fail when run in a happy-dom environment due to Vue 3 (and potentially any other framework) trying to set a valid `selectedIndex` of a select element and happy-dom erroring out.
Note that the tests run fine when the environment is set to jsdom.

Reproduction repo: https://github.com/obadakhalili/happy-dom-set-selectedIndex-error-in-vitest-reproduction.

obadakhalili:
Fixes #548.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2152,9 +2152,9 @@
 			}
 		},
 		"node_modules/@lit/reactive-element": {
-			"version": "1.3.2",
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Nice! :star:

Issue #455, opened 2022-04-14
ReferenceError: CSSMediaRule is not defined


rockwotj:
We have some code that looks like: `if (rule instanceof CSSMediaRule) {` which fails in happy-dom - I'm assuming this global needs to be defined?

capricorn86:
This should be an easy fix as CSSMediaRule exists as a class. We just need to add it to Window.

PR #557, opened 2022-08-03
#455@patch: Fix CSSMediaRule is not defined issue.

zvolcsey:
Hey, thanks for the library!
This is my first contribution to fix CSSMediaRule is not defined issue #455.
I added missing type definitions and a unit test.


capricorn86:
Hi @zvolcsey! 🙂 

Sorry for not looking into this PR earlier and thank you for contributing.

There is already a fix in v.7.0.0 of Happy DOM that should solve this issue.

I will therefore close this PR.

PR #559, opened 2022-08-07
#558@patch: Additional TreeWalker checks for parentNode.

### LINKED ISSUE #558: TreeWalker `nextSibling`/`previousSibling` throw error when `currentNode` is set to another DOM tree
When TreeWalker is walking a DOM tree it was not initialised with it throws an error when `nextSibling`/`previousSibling` are called and it reaches the (other DOM tree's) root.

Although [the spec](https://dom.spec.whatwg.org/#treewalker-current) doesn't explicitly allow or deny it, it's actually possible to initialize a TreeWalker instance and then use the `.currentNode` setter to change to a completely different DOM tree and then walk it.

Given the following reproduction:

```js
const { GlobalWindow } = require("happy-dom");

global.window = new GlobalWindow();
global.document = global.window.document;

const tree1 = document.createElement("div");
const tree2 = document.createElement("div");
const walker = document.createTreeWalker(tree1);

walker.currentNode = tree2;
walker.nextNode();
```

Running the code, this error is throw:

```
> happydom-tree-walker-repro@0.0.0 start
> node index.js

/home/max/Projects/_experiments/happydom-tree-walker-repro/node_modules/happy-dom/lib/tree-walker/TreeWalker.js:130
            const siblings = this.currentNode.parentNode.childNodes;
                                                         ^

TypeError: Cannot read properties of null (reading 'childNodes')
    at TreeWalker.nextSibling (/home/max/Projects/_experiments/happydom-tree-walker-repro/node_modules/happy-dom/lib/tree-walker/TreeWalker.js:130:58)
    at TreeWalker.nextNode (/home/max/Projects/_experiments/happydom-tree-walker-repro/node_modules/happy-dom/lib/tree-walker/TreeWalker.js:41:26)
    at Object.<anonymous> (/home/max/Projects/_experiments/happydom-tree-walker-repro/index.js:11:8)
    at Module._compile (node:internal/modules/cjs/loader:1120:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1174:10)
    at Module.load (node:internal/modules/cjs/loader:998:32)
    at Module._load (node:internal/modules/cjs/loader:839:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47

Node.js v18.7.0
```

I ~abuse~ use this behaviour in [one of my libraries](https://github.com/maxmilton/stage1/blob/master/src/compile.ts#L71) so I can reuse the same TreeWalker instance as a performance optimisation. Works as expected in browsers and `jsdom`, it's just `happy-dom` which throws.

maxmilton:
Fixes #558

Add additional check to TreeWalker `nextSibling` and `previousSibling` to make sure `this.currentNode.parentNode` exists.

Although this fixes my immediate issue, there are plenty of other checks against `this.root` throughout the class so I'm not sure if there are other changes that might be more suitable.

Extra tests should probably also be written for the scenario where the TreeWalker instance is walking a DOM tree it wasn't initialised with... but I didn't want to go too far down the rabbit hole with this PR.


capricorn86 (approved):

capricorn86:
Hi @maxmilton! 🙂 

Sorry for not looking into this PR earlier. It has been a lot going on in my private and work life.

Thank you for contributing!

Issue #534, opened 2022-07-08
TypeError: activeElement.detachEvent is not a function


hornta:
Works in jsdom
https://stackblitz.com/edit/vitest-dev-vitest-ktnq7h?file=test%2Fhappy-dom.test.tsx&initialPath=__vitest__

frankandrobot:
Seeing this as well... somehow related to ant design components. Do you mind placing some more details here @hornta ?

soker90:
Same error with happy-dom + testing-library:

```typescript
import { it } from 'vitest'
import { render } from '@testing-library/react'
import userEvent from '@testing-library/user-event'
import MyComponent from './MyComponent'

const user = userEvent.setup()

it('my test', async () => {
  const { getByPlaceholderText } = render(<MyComponent />)
  const input = getByPlaceholderText('any')

  await user.type(input, 'hello') // Crashs here
})
```

// vite.config.ts
```typescript
/// <reference types="vitest" />
import { defineConfig } from 'vite'
import react from '@vitejs/plugin-react'

export default defineConfig({
  plugins: [react()],
  test: {
    globals: true,
    environment: 'happy-dom',
    coverage: {
      reporter: ['json', 'html'],
      exclude: [
        'node_modules/'
      ]
    }
  }
})
```

hornta:
> Seeing this as well... somehow related to ant design components. Do you mind placing some more details here @hornta ?

I don't know anything about ant design components. I think the bug triggers on different types of code in each code base so it's not trivial figuring out why it occurs but I posted a failing test in the link.

doytch:
Yeah I'm running into this as well in a Remix/Vitest/RTL setup. `blur` events on a textfield seems to be what triggers it for me - whether that's via a simulated `tab`, clicking off, or manually firing a `blur`.

cloud-walker:
Hi guys, any workaround on this?

doytch:
> Hi guys, any workaround on this?

Depends what you consider a workaround… We swapped our Vitest unit tests over to ole `jsdom` with zero friction.

frankandrobot:
LoL. Note that tests don't run faster in vitest. The speedup is because test boot up is wayyyyyy faster, like 3x faster. But tests themselves will run at the same speed

So in our particular case, we're seeing massively slow RTL tests (+30s runs) and happy-dom will let us scale... In addition to vitest

On Jul 28, 2022 11:03 AM, Mark Deutsch ***@***.***> wrote:

[...]

PR #563, opened 2022-08-10
Add detachEvent to prevent react-dom errors

aaronmccall:
This PR addresses the immediate cause of the issues reported in #534.

It does not, however, address the underlying issue that react-dom detects the happy-dom environment as a legacy browser to due to lack of support for on+eventName in document and calling the setAttribute method of a div with an on+eventName attribute name not converting the supplied attribute value to a function with the value as the function body:
```js
// react-dom's expected behavior for modern runtimes (not the exact code but the same result)
var supportsOnInput = 'oninput' in document

// some modern runtimes don't pass the check above, but the following test works:
if (!supportsOnInput) {
  var div = document.createElement('div')
  div.setAttribute('oninput', 'console.log(event)')
 supportsOnInput = typeof div.oninput === 'function'
}
```


ilteoood:
Hi @capricorn86,
is this ok to be merged?

ilteoood:
Ping for this @capricorn86

capricorn86 (approved):

capricorn86:
Hi @aaronmccall! 🙂 

Thank you for contributing!

Sorry for taking such a long time to look into this PR. There has been a lot in my work and private life.

The build is failing because the commit message doesn't follow the format defined in the [contribution documentation](https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md).

If you change the commit message by doing a git amend to something in style with
`#563@minor: Adds support for Event.detachEvent() to prevent "react-dom" errors.`, it will manage to build.

Then I will merge and release this directly.

aaronmccall:
Great! Even better!

PR #579, opened 2022-09-08
Exclude source files from package

kidonng:
Currently the `happy-dom` package is shipping all the source files, which accounts for a quarter of the size:

```shellsession
$ du -sh node_modules/happy-dom/* | sort -h | tail -n2
1.2M    node_modules/happy-dom/src
3.3M    node_modules/happy-dom/lib
```

This pull request excludes them from the package by only publishing the compiled `lib` directory.


capricorn86:
I will close this pull request based on the answer. If you find another way of making source maps work, let me know.

PR #581, opened 2022-09-15
Add support for HTMLButtonElement.

fsoikin:
When writing some tests, I noticed this:

```
const el = document.createElement("button")
el.setAttribute("disabled", "")
console.log(el.disabled)  // <-- Prints "undefined"
```

In a real browser, the last line would print out `true`. Checked in Chrome and FireFox.

After digging a bit, I realized that `happy-dom` uses the generic `HTMLElement` to represent a `<button>`, which doesn't translate between [standard button attributes and properties](https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement).

This PR adds a `HTMLButtonElement` class and makes it translate some (not all) of the properties - specifically, `value`, `disabled`, and `type`.

### DIFF
--- a/packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts
+++ b/packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts
@@ -1,0 +1,86 @@
+import HTMLElement from '../html-element/HTMLElement';
+import IHTMLButtonElement from './IHTMLButtonElement';
+
+const BUTTON_TYPES = ['submit', 'reset', 'button', 'menu'];
+
+/**
+ 
+We can improve performance a bit if we make the types as a constant.
+ * HTML Button Element.
+ *
+ * Reference:
+ * https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement.
+ */
+export default class HTMLButtonElement extends HTMLElement implements IHTMLButtonElement {
+	/**
+	 * Returns value.
+	 *
+	 * @returns Value.
+	 */
+	public get value(): string {
+		return this.getAttributeNS(null, 'value');
+	}
+
+	/**
+	 * Sets value.
+	 *
+	 * @param value Value.
+	 */
+	public set value(value: string) {
+		this.setAttributeNS(null, 'value', value);
+	}
+
+	/**
+	 * Returns disabled.
+	 *
+	 * @returns Disabled.
+	 */
+	public get disabled(): boolean {
+		return this.getAttributeNS(null, 'disabled') !== null;
+	}
+
+	/**
+	 * Sets disabled.
+	 *
+	 * @param disabled Disabled.
+	 */
+	public set disabled(disabled: boolean) {
+		if (!disabled) {
+			this.removeAttributeNS(null, 'disabled');
+		} else {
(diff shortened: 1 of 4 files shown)


packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts
```diff
@@ -0,0 +1,82 @@
...
+	 */
+	public set type(v: string) {
+		this.setAttributeNS(null, 'type', this.sanitizeType(v));
+	}
+
+	/**
+	 *
+	 * @param type
+	 */
+	private sanitizeType(type: string): string {
```
capricorn86, line 73:
```suggestion
	protected _sanitizeType(type: string): string {
```

packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+
+	/**
+	 * Returns type
+	 *
+	 * @returns Type
+	 */
+	public get type(): string {
+		return this.sanitizeType(this.getAttributeNS(null, 'type'));
```
capricorn86, line 57:
```suggestion
		return this._sanitizeType(this.getAttributeNS(null, 'type'));
```

packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts
capricorn86: As the methods will be available on the instances of the button elements, I think it is good if we make it clear that they are not part of the browser standard by adding an underline as prefix.

We could also create a utility or move this logic to `public get type()` as well.

packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts
```diff
@@ -0,0 +1,82 @@
+import HTMLElement from '../html-element/HTMLElement';
+import IHTMLButtonElement from './IHTMLButtonElement';
+
+/**
```
capricorn86, line 4:
```suggestion
const BUTTON_TYPES = ['submit', 'reset', 'button', 'menu'];

/**

We can improve performance a bit if we make the types as a constant.
```

packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	/**
+	 *
+	 * @param type
+	 */
+	private sanitizeType(type: string): string {
+		type = (type && type.toLowerCase()) || 'submit';
+
+		if (!['submit', 'reset', 'button', 'menu'].includes(type)) {
```
capricorn86, line 76:
```suggestion
		if (! BUTTON_TYPES.includes(type)) {
```

capricorn86 (changes requested):
Looks good! ⭐  Just some minor things.

capricorn86:
Thank you for contributing @fsoikin! 🙂 

Sorry for not looking into this PR earlier. There has been a lot in my private and work life.

capricorn86 (approved):
Great work :star:

PR #584, opened 2022-09-26
Pop keys from registered after calling unregister

luca992:
Otherwise this sequence:

```
GlobalRegistrator.register()
GlobalRegistrator.unregister()
GlobalRegistrator.register()
```

 will throw: `Failed to register. Happy DOM has already been globally registered.`


capricorn86 (dismissed):

luca992:
👍 should be all good now.

capricorn86 (approved):

PR #615, opened 2022-10-12
#614@patch: Element.matches not working properly with descendant combinators.

### LINKED ISSUE #614: Element.matches producing false positives when using descendant combinators
**Describe the bug**
When using a [descendant combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator) (" ") with Element.matches, if the rightmost selector matches it always returns true regardless of the other selectors.

I ran into this bug when using vue-test-utils's [Wrapper.findAll](https://v1.test-utils.vuejs.org/api/wrapper/#findall), which under the scenes uses Element.matches.

**To Reproduce**
Given the following HTML:
```html
<div class="red">
   <div>
      <div class="active"></div>
   </div>
</div>
```
The following two statements, running on the innermost div (the one with class="active") both return true even though the second should return false:
```typescript
element.matches('div.red div.active');
element.matches('div.green div.active');
```

**Expected behavior**
Invalid matches should always return false.

**Device:**
 - OS: Windows
 - Browser: happy-dom
 - Version: latest

IGx89:
Fixes #614

### DIFF
--- a/packages/happy-dom/src/query-selector/QuerySelector.ts
+++ b/packages/happy-dom/src/query-selector/QuerySelector.ts
@@ -106,14 +106,14 @@ export default class QuerySelector {
 		const selector = new SelectorItem(selectorParts[0]);
 		const result = selector.match(<Element>currentNode);
 
-		if (targetNode === currentNode && !result.matches) {
+		if ((targetNode === currentNode || !currentNode.parentNode) && !result.matches) {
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Nice fix 🌟

capricorn86:
Thank you for your contribution! 🙂

PR #619, opened 2022-10-12
#618@patch: Fix various issues with <select> and <option>.

### LINKED ISSUE #618: Various <select>/<option> issues
Discovered a number of issues with the behavior of select's and option's:

1. Setting HTMLSelectElement.value doesn't ever work (never updates selected index). This breaks v-model updates in Vue.
2. Setting HTMLSelectElement.selectedIndex to -1 doesn't unselect any/all options
3. Setting HTMLSelectElement.selectedIndex doesn't update `selected` attributes on option nodes
4. Getting HTMLSelectElement.selectedIndex initially is always -1, even if there's an `<option selected>` present in the initial HTML
5. HTMLSelectElement.options doesn't reflect DOM updates when option nodes are added/removed (such as when a `v-for` on an `<option>` in Vue re-runs)

I will be submitting a PR shortly to resolve these issues.

IGx89:
Fixes #618. There are a number of other potential problems/unimplemented behaviors I noticed, but not necessary to fix for my uses (at least yet). A 100% spec compliant implementation would make certain areas a lot more complex. Good enough for now :)

cc @fatihcure-jf for visibility/review since they added the original implementation (nice work!).


packages/happy-dom/src/nodes/html-select-element/HTMLSelectElement.ts
```diff
@@ -28,10 +28,20 @@ export default class HTMLSelectElement extends HTMLElement implements IHTMLSelec
...
 	public _options: IHTMLOptionsCollection = null;
 
+	/**
+	 * Constructor.
+	 */
+	constructor() {
+		super();
+
+		// Clear our options cache when option elements are added/removed from the DOM, keeping it fresh.
+		new MutationObserver(() => {
```
davidortner-ipex, line 40: Option elements must be children of the Select element, so instead of using a MutationObserver we can just override `appendChild()`, `removeChild()` and `insertBefore()` to know which option elements it contains.

When elements are added or removed, then I think we should add and remove them as an index on the Select element as well (e.g. `this[0] = addedOption`);. We need to add the "length" property. After doing that we can easily use the select itself as an array (e.g. `for(let i = 0; i < this.length; i++) {  console.log(this[0]);  }`).

davidortner-ipex (changes requested):
Nice contribution! :star: 

I found a couple of things that I think we need to cover.

packages/happy-dom/src/nodes/html-select-element/HTMLSelectElement.ts
capricorn86, line 40: Option elements must be children of the Select element, so instead of using a MutationObserver we can just override appendChild(), removeChild() and insertBefore() to know which option elements it contains.

When elements are added or removed, then I think we should add and remove them as an index on the Select element as well (e.g. this[0] = addedOption);. We need to add the "length" property. After doing that we can easily use the select itself as an array (e.g. for(let i = 0; i < this.length; i++) { console.log(this[0]); }).

capricorn86 (changes requested):

packages/happy-dom/src/nodes/html-select-element/HTMLSelectElement.ts
IGx89: Sure, I can switch from the listener approach to the override approach -- briefly considered the latter in fact before going the former. Do note though that it won't handle when option groups are added with their own options, though lots of other things are broken with option groups so it probably doesn't matter worrying about that until option groups are properly tackled.

I'm unfortunately not following what you're saying about `this[0] = addedOption` and adding a length property. I don't think it's possible to add an indexer to a class in TypeScript, and that would be a bit odd anyway given that HTMLOptionsCollection already does all that (and is a proper Array). Unless by `this[0]` you're referring to that class instead? If you're just saying that we should simply stop caching the options property and update it dynamically, that makes sense to me.

capricorn86: What I mean is that a common way for accesing options in a Select is to do:

```javascript
const select = <HTMLSelectElement>document.querySelector('select');
for(let i = 0; i < select.length; i++) {
     // Outputs each option:
     console.log(select[i]);
}
```

If we add the selects like that, we could use that list internally as well. Not sure how the types in Typescript will be though :sweat_smile:

I don't think there is any need for a cache if we just update the collection on `appendChild()`, `removeChild()` and `insertBefore()`.

IGx89: Made a few updates, let me know how it looks now!

capricorn86:
@IGx89 I made some adjustment in a commit:

 - `select.options.add()` and `select.options.remove()` has to update the actual DOM with `select.appendChild()`, `select.beforeChild()` or `select.removeChild`
 - `select.options.selectedIndex` was not updated correctly after removing a selected option
 - `HTMLSelectElement` now supports to do get options by index (e.g. `select[1]`)
 - We still have to sanitize the value set in HTMLOptionElement, but we need to stringify it first
 
 Thanks for you contribution! :slightly_smiling_face: 
 
 I will release it now.

IGx89:
> @IGx89 I made some adjustment in a commit:
> 
> * `select.options.add()` and `select.options.remove()` has to update the actual DOM with `select.appendChild()`, `select.beforeChild()` or `select.removeChild()`
> * `select.options.selectedIndex` was not updated correctly after removing a selected option
> * `HTMLSelectElement` now supports to do get options by index (e.g. `select[1]`)
> * We still have to sanitize the value set in `HTMLOptionElement`, but we need to stringify it first
> 
> Thanks for you contribution! 🙂
> 
> I will release it now.

Thanks! Looks like I opened a can of worms when fixing the issue of select.options not reflecting DOM updates :p. Definitely good improvements all around! Only question I have is about still needing to sanitize the value set in HTMLOptionElement. It wasn't in [the spec](https://html.spec.whatwg.org/multipage/form-elements.html#dom-option-value) from what I could see -- where are you seeing it? I see it for HTMLInputElement.value, but not HTMLOptionElement.value.

capricorn86:
@IGx89 you are right! It was added by somebody else as a contribution and I just assumed that there was a reason for why it was added. Good catch!

I have made another release with a fix for removing the sanitize logic:
https://github.com/capricorn86/happy-dom/pull/621

PR #623, opened 2022-10-15
#622@patch: Fix Element.matches failing when using non-matching desce…

### LINKED ISSUE #622: Element.matches fails when using non-matching descendant selector on element attached to document
```ts
document.appendChild(element);
expect(element.matches('.nonexistent-class div')).toBe(false);
```

```
TypeError: Cannot read properties of undefined (reading 'split')

  296 |     ): { priorityWeight: number; matches: boolean } {
  297 |             const regexp = new RegExp(CLASS_REGEXP, 'g');
> 298 |             const classList = element.className.split(' ');
	  |                                                 ^
  299 |             const classSelector = selector.split(':')[0];
  300 |             let priorityWeight = 0;
  301 |             let match: RegExpMatchArray;

  at SelectorItem.matchesClass (src/query-selector/SelectorItem.ts:298:39)
  at SelectorItem.match (src/query-selector/SelectorItem.ts:91:24)
  at Function.matchesSelector (src/query-selector/QuerySelector.ts:107:27)
  at Function.matchesSelector (src/query-selector/QuerySelector.ts:113:15)
  at Function.matchesSelector (src/query-selector/QuerySelector.ts:113:15)
  at Function.matchesSelector (src/query-selector/QuerySelector.ts:113:15)
  at Function.match (src/query-selector/QuerySelector.ts:66:24)
  at HTMLElement.matches (src/nodes/element/Element.ts:734:24)
```

IGx89:
…ndant selector on element attached to document.

Fixes #622. Was calling selector.match on the Document, which isn't an Element.

### DIFF
--- a/packages/happy-dom/src/query-selector/QuerySelector.ts
+++ b/packages/happy-dom/src/query-selector/QuerySelector.ts
@@ -104,15 +104,15 @@ export default class QuerySelector {
 		}
 
 		const selector = new SelectorItem(selectorParts[0]);
-		const result = selector.match(<Element>currentNode);
+		const result = selector.match(<IElement>currentNode);
 
-		if ((targetNode === currentNode || !currentNode.parentNode) && !result.matches) {
+		if ((targetNode === currentNode || !currentNode.parentElement) && !result.matches) {
 			return { priorityWeight: 0, matches: false };
 		}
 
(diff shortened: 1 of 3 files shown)


capricorn86 (approved):
Nice! :star2:

PR #625, opened 2022-10-18
#622@patch: Fix Element.matches failing when using non-matching desce…

IGx89:
…ndant selector on element detached from document.

Fixes second issue reported under #622

### DIFF
--- a/packages/happy-dom/src/query-selector/QuerySelector.ts
+++ b/packages/happy-dom/src/query-selector/QuerySelector.ts
@@ -106,7 +106,14 @@ export default class QuerySelector {
 		const selector = new SelectorItem(selectorParts[0]);
 		const result = selector.match(<IElement>currentNode);
 
-		if ((targetNode === currentNode || !currentNode.parentElement) && !result.matches) {
+		if (result.matches && selectorParts.length === 1) {
+			return {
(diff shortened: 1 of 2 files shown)


packages/happy-dom/src/query-selector/QuerySelector.ts
```diff
@@ -106,7 +106,14 @@ export default class QuerySelector {
...
 
-		if ((targetNode === currentNode || !currentNode.parentElement) && !result.matches) {
+		if (result.matches && selectorParts.length === 1) {
+			return {
+				priorityWeight: priorityWeight + result.priorityWeight,
+				matches: true
+			};
+		}
+
+		if (!currentNode.parentElement || (targetNode === currentNode && !result.matches)) {
```
IGx89, line 116: Refactored slightly to always verify parentElement is not null, making it pass null checks in case you ever turn on strictNullChecks.

capricorn86: Yes that would be good to turn on 🙂

capricorn86 (approved):

RahulNearSt:
I was using TableCell from materialUI-core and experiencing this issue until your fix. Thanks for the quick fix and release!

PR #626, opened 2022-10-18
#308@minor: Implement NamedNodeMap and use it for Element.attributes

### LINKED ISSUE #308: Elment.attributes does not implement the NamedNodeMap interface
I was trying to replace JSDOM with Happy DOM and the app was failing to run something akin to:
```
const someAttribute = document.createAttribute('someattribute');
element.attributes.setNamedItem(someAttribute);
```

I see the current implementation is:
```
readonly attributes: { [k: string]: Attr | number };
```

Is there any chance there will be a switch to [NamedNodeMap ](https://developer.mozilla.org/en-US/docs/Web/API/NamedNodeMap)in the future?

jledentu:
Hi,

This PR implements the [NamedNodeMap](https://developer.mozilla.org/en-US/docs/Web/API/NamedNodeMap) interface and makes it available in `Window`. `Element.attributes` now returns a `NamedNodeMap`.

Closes #308

### DIFF
--- a/packages/happy-dom/src/named-node-map/INamedNodeMap.ts
+++ b/packages/happy-dom/src/named-node-map/INamedNodeMap.ts
@@ -1,4 +1,4 @@
-import IAttr from './IAttr';
+import IAttr from '../nodes/attr/IAttr';
 
 export type INamedNodeMapProps = {
 	readonly length: number;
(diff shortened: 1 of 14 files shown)


capricorn86 (approved):
Thank you for your contribution @jledentu!  It looks really nice :star: 

I have made some adjustment, so that the interface looks more like the build in Typescript DOM interface. Unfortunately Typescirpt dosn't handle both `[k: string]` and other properties and methods at the same time, but it also makes sense in a way and this is how they have done it.

I also added some tests.

PR #629, opened 2022-10-20
#628@patch: Handle style setter in HTMLElement.

### LINKED ISSUE #628: `HTMLElement.style` cannot be set directly
**Describe the bug**

`HTMLElement.style` cannot be set, i.e `element.style = 'display: none'` doesn't work in happy-dom though it works in browsers (even if the `style` property is readonly according to the spec).

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style#setting_styles

> While this property is considered read-only, it is possible to set an inline style by assigning a string directly to the style property. In this case the string is forwarded to [CSSStyleDeclaration.cssText](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/cssText). Using style in this manner will completely overwrite all inline styles on the element.

**To Reproduce**

```ts

const window = new Window();
button.style = 'display: none';

console.log(button.outerHTML); // "<button></button>"
```

**Expected behavior**

```ts
button.style = 'display: none';
console.log(button.outerHTML); // should display "<button style="display: none"></button>"
```

**Device:**
 - Version 7.6.0
 - Node 16.17.1

jledentu:
This PR adds a setter on `HTMLElement.style` the reproduce the behaviour of DOM implementations, as explained in #628. 

This also fixes another difference between happy-dom and browser implementations: setting `null` or `''` as `style` (or in `cssText`)  shouldn't remove the `style` attribute.

```ts
element.style.cssText = null;
console.log(element.getAttribute('style')); // displays '', not null
```

Since TypeScript forces a value returned by a getter to be assignable to a setter (see https://github.com/microsoft/TypeScript/pull/42425), the `style` setter accepts a `CSSStyleDeclaration` but then resets the `cssText`, as implemented in some browsers.

Closes #628

### DIFF
--- a/packages/happy-dom/src/nodes/html-element/HTMLElement.ts
+++ b/packages/happy-dom/src/nodes/html-element/HTMLElement.ts
@@ -207,7 +207,7 @@ export default class HTMLElement extends Element implements IHTMLElement {
 	 * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style#setting_styles
 	 */
 	public set style(cssText: string | CSSStyleDeclaration) {
-		this.setAttribute('style', typeof cssText === 'string' ? <string>cssText : '');
+		this.style.cssText = typeof cssText === 'string' ? <string>cssText : '';
 	}
 
 	/**
--- a/packages/happy-dom/src/nodes/html-element/HTMLElement.ts
+++ b/packages/happy-dom/src/nodes/html-element/HTMLElement.ts
@@ -201,6 +201,16 @@ export default class HTMLElement extends Element implements IHTMLElement {
 	}
 
 	/**
+	 * Sets style.
+	 *
+	 * @param cssText Style as text.
+	 * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style#setting_styles
+	 */
+	public set style(cssText: string | CSSStyleDeclaration) {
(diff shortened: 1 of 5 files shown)


packages/happy-dom/src/nodes/html-element/HTMLElement.ts
```diff
@@ -200,6 +200,16 @@ export default class HTMLElement extends Element implements IHTMLElement {
...
 	}
 
+	/**
+	 * Sets style.
+	 *
+	 * @param cssText Style as text.
+	 * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style#setting_styles
+	 */
+	public set style(cssText: string | CSSStyleDeclaration) {
+		this.setAttribute('style', typeof cssText === 'string' ? <string>cssText : '');
```
capricorn86, line 210:
```suggestion
		this.style.cssText = cssText;
```

We have to parse the CSS string before setting it (e.g. `element.style = 'invalid: invalid'` won't work).

capricorn86 (changes requested):
Nice contribution! :star2: 

I just found a minor thing.

packages/happy-dom/src/nodes/html-element/HTMLElement.ts
jledentu: Thanks, I updated

capricorn86 (approved):

Issue #633, opened 2022-10-23
getElementByTagName doesn't support wildcard searches


IGx89:
getElementByTagName can be passed a special string `"*"` representing all elements ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/getElementsByTagName#parameters)), causing it to return all elements. That currently doesn't work.

jQuery relies on that functionality for non-trivial selector searches, so without this functionality jQuery support is partially broken. For example, `$('[attr]')` doesn't work.

capricorn86:
Thanks for your contribution @IGx89! :slightly_smiling_face: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v7.6.2

PR #634, opened 2022-10-23
#633@patch: Fix getElementByTagName doesn't support wildcard searches (improves jQuery support).

IGx89:


capricorn86 (approved):
Nice contribution @IGx89! :star2:

Issue #635, opened 2022-10-24
Setting an invalid value to HTMLSelectElement.selectedIndex causes a crash


capricorn86:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Device:**

**Additional context**

PR #636, opened 2022-10-24
#635@patch: Multiples fixes regarding how HTMLSelectElement.selectedI…

capricorn86:
…ndex is handled.

### DIFF
--- a/packages/happy-dom/src/nodes/html-option-element/HTMLOptionElement.ts
+++ b/packages/happy-dom/src/nodes/html-option-element/HTMLOptionElement.ts
@@ -1,6 +1,7 @@
 import HTMLElement from '../html-element/HTMLElement';
 import IHTMLElement from '../html-element/IHTMLElement';
 import IHTMLFormElement from '../html-form-element/IHTMLFormElement';
+import IHTMLSelectElement from '../html-select-element/IHTMLSelectElement';
 import IHTMLOptionElement from './IHTMLOptionElement';
 
 /**
@@ -58,7 +59,20 @@ export default class HTMLOptionElement extends HTMLElement implements IHTMLOptio
 	 * @returns Selected.
 	 */
 	public get selected(): boolean {
-		return this.getAttributeNS(null, 'selected') !== null;
+		const parentNode = <IHTMLSelectElement>this.parentNode;
+
+		if (parentNode?.tagName === 'SELECT') {
+			let index = -1;
+			for (let i = 0; i < parentNode.options.length; i++) {
+				if (parentNode.options[i] === this) {
+					index = i;
+					break;
+				}
+			}
+			return index !== -1 && parentNode.options.selectedIndex === index;
+		}
+
+		return false;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


packages/happy-dom/test/nodes/html-option-element/HTMLOptionsCollection.test.ts
```diff
@@ -134,16 +128,21 @@ describe('HTMLOptionsCollection', () => {
...
 			const option2 = document.createElement('option');
+
+			expect(select.options.selectedIndex).toBe(-1);
+
 			select.appendChild(option);
 			select.appendChild(option2);
-			document.body.appendChild(select);
-			expect(select.options.selectedIndex).toBe(-1);
+
+			expect(select.options.selectedIndex).toBe(0);
```
IGx89, line 137: This has me confused: the spec says it should be -1 in this situation ([ref](https://html.spec.whatwg.org/multipage/form-elements.html#dom-select-selectedindex)), but the browser says 0 😖. Did you find a place in the spec indicating it should actually be 0, or are you changing this purely based on observations of how browsers actually act?

packages/happy-dom/src/nodes/html-option-element/HTMLOptionsCollection.ts
```diff
@@ -34,14 +34,7 @@ export default class HTMLOptionsCollection
 	 * @returns SelectedIndex.
 	 */
 	public get selectedIndex(): number {
-		for (let i = 0; i < this.length; i++) {
-			const item = this[i];
-			if (item instanceof HTMLOptionElement && item.selected) {
```
IGx89, line 39: I'm a bit concerned that by removing this behavior you're going to break some things. If the HTML was the following:

```html
<select>
	<option>1</option>
	<option selected>2</option>
</select>
```
The browser says selectedIndex is 1, but with this change it'll wrongly be 0, won't it?

packages/happy-dom/src/nodes/html-option-element/HTMLOptionElement.ts
```diff
@@ -67,10 +81,26 @@ export default class HTMLOptionElement extends HTMLElement implements IHTMLOptio
 	 * @param selected Selected.
 	 */
 	public set selected(selected: boolean) {
-		if (!selected) {
-			this.removeAttributeNS(null, 'selected');
-		} else {
-			this.setAttributeNS(null, 'selected', '');
```
IGx89, line 73: Is this intentional that you're not adding/removing the `selected` DOM attribute anymore? Isn't that necessary? This'll make it harder to implement multi-selects in the future too I'm guessing.

packages/happy-dom/src/nodes/html-option-element/HTMLOptionsCollection.ts
capricorn86: @IGx89 Yes you are right. I will fix it now directly. I wanted to correct the `selectedIndex` behavior as the current behavior does not follow the spec or the browser behavior, but I missed the "selected" attribute.

packages/happy-dom/src/nodes/html-option-element/HTMLOptionElement.ts
capricorn86: Changing the `selectedIndex` property should not modify the attribute value. It should set the `HTMLOptionElement` to be dirty and then it overrides the attribute.

It is not entirely correct now as I missed handling the attribute, sorry about that.

https://html.spec.whatwg.org/multipage/form-elements.html#dom-select-selectedindex
https://html.spec.whatwg.org/multipage/form-elements.html#concept-option-dirtiness

IGx89: Thanks! I assumed setting "selectedness" means setting the selected attribute, which does seem wrong. I can't find anything on those links that specifically details what happens when an option's selectedness is set, which is strange, but what you're saying definitely makes sense

packages/happy-dom/test/nodes/html-option-element/HTMLOptionsCollection.test.ts
capricorn86: I found it by testing the behavior in the browser, but there is a spec for it.

Spec:
https://html.spec.whatwg.org/multipage/form-elements.html#selectedness-setting-algorithm

IGx89: Ahh, thanks a lot! So it looks like -1 is only a valid initial value if every option is disabled.

capricorn86: Yes it seems like that. I will try to fix it so that it works according to spec now. Wish me luck :sweat_smile:

capricorn86:
@IGx89 I have released a fix now. Hopefully everything regarding "selectedness" is fixed now. Thanks for reporting the problem :slightly_smiling_face: 

Here is the pull request:
https://github.com/capricorn86/happy-dom/pull/638

Issue #637, opened 2022-10-25
HTMLSelectElement.selectedIndex no longer cater for selected attribute in options


capricorn86:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Device:**

**Additional context**

PR #638, opened 2022-10-25
#637@patch: Fixes problem with HTMLSelectElement.selectedIndex not re…

capricorn86:
…flecting the select attribute set on options.


IGx89:
Nice work!

capricorn86:
> Nice work!

Thank you! 🙂

PR #641, opened 2022-10-30
#204@minor: Add support for HTMLAnchorElement.

### LINKED ISSUE #204: Add support for HTMLAnchorElement

IGx89:
Resolves #204


capricorn86:
Nice contribution @IGx89! :star2: 

I have done some improvements to performance. It is the same logic, but will improve performance as the URL does not need to be reinitialized unless the "href" is changed.

There was also some problems with the handling of "blob" urls, which is hard to understand in the specs. Chrome doesn't seem to cater for the case, and FF added support blob urls in v49 ([reference](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/origin)).

How I interpret the specs, the idea is that property changes on the URL object should be ignored if it is a "blob" url with an opaque path as they are not valid on the URL. This is also how FF seem to have implemented it.

Please review the changes in the commit. Once you have done that, I am happy to release :slightly_smiling_face:

IGx89:
@capricorn86 thanks! Looks great. Good teamwork :). :shipit:

Issue #645, opened 2022-11-05
Missing `element.click()`


ekwoka:
`IElement` returned by `document.querySelector` has a `click()` method that functions, but TypeScript says it does not exist.

Expected behavior: TypeScript not complain about this existing.

capricorn86:
According to spec `Element.querySelector()` and `Document.querySelector()` should return elements of type `Element` which does not have the `click()` method. However, elements of type `HTMLElement` should have the `click()` method.

See HTMLElement interface:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click

See `Element.querySelector()`:
https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector

In order to solve this you just need to cast:
```javascript
const element = <HTMLElement>document.querySelector('div');
```

PR #646, opened 2022-11-05
#645@patch: Declares type for element.click()

ekwoka:
https://github.com/capricorn86/happy-dom/issues/645

This declares `click` on the `IElement` interfact


capricorn86 (changes requested):
Thank you for your contribution :slightly_smiling_face: 

The `click()` method should be part of the `HTMLElement` interface and not the `Element` interface.

This has already been implemented:
https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-element/IHTMLElement.ts#L56

See:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click

capricorn86:
According to spec `Element.querySelector()` and `Document.querySelector()` should return elements of type `Element` which does not have the `click()` method.

In order to solve this you just need to cast:
```javascript
const element = <HTMLElement>document.querySelector('div');
```

Issue #647, opened 2022-11-07
document.querySelector('#--test--') returns null.


coji:
**Describe the bug**

document.querySelector('#--test--') returns null.

**Expected behavior**

The id specification in document.querySelector must support all characters except non-whitespace characters. 

It's [defined in the WHATWG standard.](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute)
>When specified on HTML elements, the id attribute value must be unique amongst all the IDs in the element's tree and must contain at least one character. The value must not contain any ASCII whitespace.

**To Reproduce**

1. Open [stackblitz](https://stackblitz.com/github/coji/happy-dom-id-bug-reproduction?file=query-selector-id-start-with-hyphen.js)
2. In terminal, run `node query-selector-id-start-with-hyphen.js`

reproduce repo is here. 
https://github.com/coji/happy-dom-id-bug-reproduction

**Device:**
 - OS: MacOS 13.0 / node 16.17.1
 - happy-dom 7.6.6

karpiuMG:
@capricorn86 can be closed

follow repro steps but update happy-dom before running a script

PR #651, opened 2022-11-08
#647@patch: The id specification in document.querySelector must suppo…

coji:
…rt all characters except non-whitespace character.


capricorn86 (changes requested):
Thank you for your contribution! :slightly_smiling_face: 

We do not fully solve the problems with ID selectors here. The problem is that ID selectors can be combined with other selectors.

The current solution doesn't fully cater for that either, but I'm afraid that we may introduce a new bug if we release this fix :sweat_smile: 

Example:
```javascript
var div1 = document.createElement('div');
div1.id = 'id';
div1.setAttribute('attr', 'a');

var div2 = document.createElement('div');
div2.id = 'id';
div2.setAttribute('attr', 'b');

document.body.appendChild(div1);
document.body.appendChild(div2);

// Outputs "true"
console.log(document.querySelector('#id[attr="a"]') === div1);

// Outputs "true"
console.log(document.querySelector('#id[attr="b"]') === div2);
```

packages/happy-dom/src/query-selector/SelectorItem.ts
```diff
@@ -8,7 +8,7 @@ const ATTRIBUTE_NAME_REGEXP = /[^a-zA-Z0-9-_$]/;
 const PSUEDO_REGEXP = /:([a-zA-Z-]+)\(([0-9n+-]+|odd|even)\)|:not\(([^)]+)\)|:([a-zA-Z-]+)/g;
 const CLASS_REGEXP = /\.([a-zA-Z0-9-_$]+)/g;
 const TAG_NAME_REGEXP = /^[a-zA-Z0-9-]+/;
-const ID_REGEXP = /#[A-Za-z][-A-Za-z0-9_]*/g;
+const ID_REGEXP = /#[^\s]+/g;
```
capricorn86, line 11:
```suggestion
const ID_REGEXP = /#[A-Za-z][A-Za-z0-9_-]*/g;
```

Based on what is written on [this Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id), it seems like browsers will throw an error for any other value than the above RegExp.

Example:
```javascript
// Throws an error
document.querySelector('#1myid');

// Throws an error
document.querySelector('#myid[');
```

capricorn86 (changes requested):

coji:
Thanks for the review. I understand that the bug may occur in a case I did not anticipate.

I will close this issue for now. Thank you very much.

capricorn86:
Thank you for taking time to create this PR. Hope the problem got solved :slightly_smiling_face:

Issue #661, opened 2022-11-20
Does not apply styles from tailwind-like classes like `md:flex`


petejodo:
**Describe the bug**
`happy-dom` is not applying the styles of classes defined in a tailwind-like way These are classes that
have colon (`:`) separators like `md:flex`. I suspect the fundamental issue has to do with escaping because
the CSS for these classes look like:

```css
.md\:flex {
  display: flex;
}
```

I created a repro repo [here](https://github.com/petejodo/happy-dom-repro/tree/tailwind-css-classes).
The test looks like:

```tsx
import { render, screen } from "@testing-library/react";
import { Example } from "./example";

const css = `
.hidden { display: none; }
@media (min-width: 768px) {
    .md-flex {
        display: flex;
    }
    .md\:flex {
        display: flex;
    }
}
`;

test("should be visible", () => {
    const style = document.createElement("style");
    style.innerHTML = css;
    document.head.appendChild(style);

    // @ts-ignore
    window.happyDOM.setInnerWidth(768);

    render(<Example />)
    expect(screen.getByText("example with dash")).toBeVisible()
    expect(screen.getByText("example with tailwind-like class")).toBeVisible()
});
```

The `Example` component looks like:

```tsx
export function Example() {
    return (
        <div>
            <div className="hidden md-flex">example with dash</div>
            <div className="hidden md:flex">example with tailwind-like class</div>
        </div>
    );
}
```

I expect the test to pass but it isn't and is failing on this line:

```ts
expect(screen.getByText("example with tailwind-like class")).toBeVisible()
```

**To Reproduce**
1. Clone repro repo [here](https://github.com/petejodo/happy-dom-repro/tree/tailwind-css-classes)
2. Run `npm test`
3. See test fail

**Expected behavior**
The test should pass with the element in question on the failed line should be visible.

**Device:**
 - OS: MacOS Monterey
 - Browser: N/A, ran in terminal

petejodo:
Found at least one part of the bug. [This line](https://github.com/capricorn86/happy-dom/blob/bd300f17993aa32cbec875fda0c061a904772b84/packages/happy-dom/src/query-selector/SelectorItem.ts#L33) in `SelectItem` is altering the selector from `.md\\:flex` (the double `\` is the string form and escaping the `\`) to `.md\\`. This means that `PSUEDO_REGEXP` is not quite correct

petejodo:
[...]

PR #662, opened 2022-11-21
#661@patch: Take class selector escaping into account.

petejodo:
This is definitely not the best way of doing this but it does seem to fix the issue #661. This is more of a proof of concept or pathfinder for how to fix the issue properly. See the issue for more details.

### DIFF
--- a/packages/happy-dom/src/query-selector/SelectorItem.ts
+++ b/packages/happy-dom/src/query-selector/SelectorItem.ts
@@ -7,9 +7,9 @@ const ATTRIBUTE_REGEXP =
 const ATTRIBUTE_NAME_REGEXP = /[^a-zA-Z0-9-_$]/;
 const PSUEDO_REGEXP =
 	/(?<!\\):([a-zA-Z-]+)\(([0-9n+-]+|odd|even)\)|(?<!\\):not\(([^)]+)\)|(?<!\\):([a-zA-Z-]+)/g;
-const CLASS_REGEXP = /\.(([a-zA-Z0-9-_$]|\\:)+)/g;
+const CLASS_REGEXP = /\.(([a-zA-Z0-9-_$]|\\.)+)/g;
 const TAG_NAME_REGEXP = /^[a-zA-Z0-9-]+/;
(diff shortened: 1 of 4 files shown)


packages/happy-dom/src/query-selector/SelectorItem.ts
```diff
@@ -5,10 +5,13 @@ import Element from '../nodes/element/Element';
...
 	/\[([a-zA-Z0-9-_]+)\]|\[([a-zA-Z0-9-_]+)([~|^$*]{0,1})[ ]*=[ ]*["']{0,1}([^"']+)["']{0,1}\]/g;
 const ATTRIBUTE_NAME_REGEXP = /[^a-zA-Z0-9-_$]/;
-const PSUEDO_REGEXP = /:([a-zA-Z-]+)\(([0-9n+-]+|odd|even)\)|:not\(([^)]+)\)|:([a-zA-Z-]+)/g;
-const CLASS_REGEXP = /\.([a-zA-Z0-9-_$]+)/g;
+const PSUEDO_REGEXP =
+	/(?<!\\):([a-zA-Z-]+)\(([0-9n+-]+|odd|even)\)|(?<!\\):not\(([^)]+)\)|(?<!\\):([a-zA-Z-]+)/g;
+const CLASS_REGEXP = /\.(([a-zA-Z0-9-_$]|\\:)+)/g;
 const TAG_NAME_REGEXP = /^[a-zA-Z0-9-]+/;
 const ID_REGEXP = /#[A-Za-z][-A-Za-z0-9_]*/g;
+const CSS_ESCAPE_REGEXP = /(?<!\\):/g;
```
petejodo, line 13: this escape should be more than just for `:` but don't know if I can just have it be any arbitrary character or a set of characters that can be escaped like `[:#&...]`

capricorn86:
Thank you for your contribution @petejodo! :slightly_smiling_face: 

I have made a commit to include any escaped character. I also added some more unit tests.

I feel pretty confident in that this should work as there is good test coverage, so I will therefore release it.

Issue #285, opened 2021-07-06
querySelectorAll shoud respect document order


olace:
When using multiple selectors seprated by coma with querySelectorAll, the returned elements should be in document order as per the [specifications](https://www.w3.org/TR/selectors-api/#queryselectorall).

For example, with the following HTML:
```
<h1>1.0</h1>
    <h2>1.1</h2>
    <h2>1.2</h2>
<h1>2.0</h1>
    <h2>2.1</h2>
    <h2>2.2</h2>
```
If I call document.querySelectorAll("h1,h2") happy-dom returns [1.0, 2.0, 1.1, 1.2, 2.1, 2.2] instead of the expected [1.0, 1.1, 1.2, 2.0, 2.1, 2.2]

capricorn86:
Hi @olace! :slightly_smiling_face: 

Sorry for taking such long time answering this. I will look into fixing this as soon as possible.

kevinbarabash:
I'm trying to use `happy-dom` with `@testing-library/user-event`, but ran into an issue with simulated `tab` key presses not focusing the next tabbable element.  I thought I could work around it by querying for all of the tabbable elements and using that result to figure out which element to focus next, but then I ran into this issue with the results of `querySelectorAll` being out of order.

capricorn86:
Hi @olace and @kevinbarabash! :slightly_smiling_face: 

This has finally been fixed.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v9.10.0

ariofrio:
This still seems to be occurring using happy-dom 9.19.2, for example in Substack posts. See

```
const { Window } = require("happy-dom");

async function main() {
  const url =
    "https://magneticgrowth.substack.com/p/api-renaissance-the-dawn-of-api-unicorns";
  const response = await fetch(url);
  const html = await response.text();

  const window = new Window({
    url,
    settings: {
      disableCSSFileLoading: true,
      disableIframePageLoading: true,
      disableJavaScriptEvaluation: true,
      disableJavaScriptFileLoading: true,
    },
  });
  window.document.write(html);
  const lines = window.document
    .querySelector("article")
    .querySelectorAll("h1, h2, h3, h4, h5, h6, p, blockquote, li")
    .map((element) => element.textContent);
  console.log(lines.map((line) => line.trim().substring(0, 50)).join("\n"));
}

main();
```

Generates:

```
API Renaissance: The Dawn of API Unicorns Supporti
[...]

PR #683, opened 2022-12-20
#285@patch: Results of querySelector & querySelectorAll are now in or…

Schleuse:
…der of occurrence in the document.


capricorn86:
Thanks for your contribution @Schleuse! :slightly_smiling_face: 

I will have to look into the code a bit more. I would like to see if there is a way we can avoid more for loops as these are quite performance sensitive methods.

capricorn86:
Again, thank you for your contribution @Schleuse! :star2: 

This has been fixed in another PR as I wanted to do some refactoring of the entire query selection logic in order to not hurt performance when implementing document order and adjacent sibling selectors.

Pull request:
https://github.com/capricorn86/happy-dom/pull/878

Release:
https://github.com/capricorn86/happy-dom/releases/tag/v9.10.0

PR #677, opened 2022-12-07
#531@patch: Adds support for setting the checked property on checkbox…

capricorn86:
-   Implements `Window.fetch()`, `Window.Request`, `Window.Response` and `Window.Headers` in Happy DOM, instead of using the "node-fetch" library. The code is based on "node-fetch". This improves the integration for functionality such as `FormData`, `Blob`, `AbortController` and `AbortSignal` with `Window.fetch()`. It will also make it possible to add support for **ESM** while still keeping support for CommonJS.
-   Adds support for `FormData`
-   Adds support for `AbortController` and `AbortSignal`
-   Adds full support for `ValidityState`
-   Changes internal logic to favor `Element.setAttribute()` and `Element.removeAttribute()` instead of `Element.setAttributeNS()` and `Element.removeAttributeNS()` to improve performance
-   Adds support for element name and id as property to `HTMLCollection`
-   Adds support for `HTMLCollection.namedItem()`
-   Adds support for `HTMLButtonElement.name`
-   Adds support for `HTMLButtonElement.form`
-   Adds support for `HTMLButtonElement.checkValidity()`
-   Adds support for `HTMLFormControlsCollection`
-   Improves support for `HTMLFormElement.elements`
-   Adds support for form element name and id as property to `HTMLFormElement`
-   Adds support for `HTMLFormElement.checkValidity()`. It previously just returned "true".
-   Adds support for `HTMLFormElement.reportValidity()`.
-   Adds support for `HTMLFormElement.submit()`.
-   Adds support for `HTMLFormElement.reset()`.
-   Adds support for `RadioNodeList`.
-   Improves support for `HTMLInputElement.checked`
-   Adds support for `HTMLInputElement.labels`
-   Improves performance of `HTMLInputElement.form`
-   Adds support for `HTMLInputElement.checkValidity()`. It previously just returned "true".
-   Adds support for `HTMLInputElement.reportValidity()`
-   Adds support triggering a "click" event on `HTMLInputElement` to trigger functionality on checkboxes, radio buttons, submit buttons and reset buttons
-   Improves performance of `HTMLLabelElement.form`
-   Improves support for `HTMLSelectElement.options`
-   Adds support for `HTMLSelectElement.labels`
-   Improves support option index properties in `HTMLSelectElement`
-   Adds support for `HTMLSelectElement.checkValidity()`.
-   Adds support for `HTMLSelectElement.reportValidity()`
-   Improves performance of `HTMLSelectElement.form`
[...]

PR #685, opened 2022-12-24
#678@patch: Assign of window.location.href.

### LINKED ISSUE #678: regression: assign of window.location.href
**Describe the bug**

Starting with version 7.8.0, setting the `href` property like this:
```
window.location.href = '/foo'
```

Will throw an exception: `ERR_INVALID_URL`.

**To Reproduce**
1. Create a test with happy-dom
2. In the test, insert `window.location.href = '/foo'`
3. See error `ERR_INVALID_URL`

**Expected behavior**
I can set the `href` property using a string which doest not represent a full url, in the same exact way in which both chrome and firefox works.

**Screenshots**
image

**Device:**
 - OS: OS X
 - Browser: Vitest
 - Version: v0.25.6

**Additional context**
This happens using happy-dom during tests.

Mas0nShi:
Fixes #678.


packages/happy-dom/src/location/Location.ts
```diff
@@ -11,6 +11,13 @@ export default class Location extends URL {
 		super('about:blank');
 	}
 
+	/**
+	 * Override set href.
+	 */
+	set href(value: string) {
+		super.href = new URL(value, this).href;
```
ilteoood, line 18: I don't think that this fixes the issue. If you set /Foo as value, the new url will throw an error

packages/happy-dom/test/location/Location.test.ts
```diff
@@ -30,4 +30,12 @@ describe('Location', () => {
 			expect(location.href).toBe(HREF);
 		});
 	});
+
+	describe('href', () => {
+		it('Set href.', () => {
+			location.replace(HREF);
```
ilteoood, line 36: You should test against a plain string like /Foo instead of using a valid url

Mas0nShi: did you means `location.href = '/foo'` directly, like this:

```javascript
const {Window} = require('happy-dom');
const window = new Window();

window.location.href = '/foo'
```

ilteoood: Yes, exactly. I'm watching this from my phone but it seems that HREF is a complete and valid url.

Mas0nShi: > Yes, exactly. I'm watching this from my phone but it seems that HREF is a complete and valid url.


we need have a valid url before we `location.href = '/foo'`, it's same in browser. `/foo` only a path, **it's not valid url**. you can set href `/foo` because it's replace pathname only.
you can test in your browser devtools:

```javascript
location.href = 'about:blank' // the blank page in browser.

location.href = '/foo'

// Uncaught DOMException: Failed to set the 'href' property on 'Location': '/foo' is not a valid URL.
//    at <anonymous>:1:15

```

hugo-vrijswijk:
I hate to be that guy, but any update on this? It's currently blocking me from using happy-dom instead of jsdom

capricorn86:
> I hate to be that guy, but any update on this? It's currently blocking me from using happy-dom instead of jsdom

@hugo-vrijswijk the build for this PR is failing and I would also like to look into the solution a bit more. I am currently working on #677, which adds support for FormData and implements "fetch" into Happy DOM. It is a bit time consuming. I will look into this as soon as I can 😅

packages/happy-dom/src/location/Location.ts
capricorn86: I this is the only solution. We need to use @ts-ignore here as the types are wrong for the URL library.

capricorn86 (dismissed):

packages/happy-dom/test/location/Location.test.ts
capricorn86: @ilteoood as @Mas0nShi said, the URL needs to be relative to something and by default the URL is set to "about:blank".

Only setting `location.href = '/foo'` will fail with an error as it is not possible to construct a relative URL from "about:blank". Happy DOM could use a different default value for the URL, but what would that be?

After this fix, you will still have to use a setup script for your tests and set a default URL for the URL:s to be relative from.

capricorn86 (approved):

packages/happy-dom/test/location/Location.test.ts
capricorn86: I have added unit tests that demonstrates this.

PR #692, opened 2023-01-19
#692@patch: Fixed  CSSContainerRule, CSSMediaRule conditionText typo

dr2009:
CSSContainerRule, CSSMediaRule has conditionText prop, not conditionalText

https://developer.mozilla.org/en-US/docs/Web/API/CSSConditionRule


capricorn86 (approved):
Nice! ⭐

PR #696, opened 2023-01-23
#656@minor: Rename `nodeEquals` to `isEqualNode`.

### LINKED ISSUE #656: Node.isEqualNode()
I wanted to compare two happy-dom nodes in my tests, however the library does not implement the standard
[Node.isEqualNode](https://developer.mozilla.org/en-US/docs/Web/API/Node/isEqualNode) browser method.

Instead I noticed there is a static function (`nodeEquals`) that implements this functionality in `NodeUtility`.

The issue is I did not realise this functionality was available  in the library until I dug through the source 
code, and I had to do this import in order to use it:
```js
import NodeUtility from "happy-dom/lib/nodes/node/NodeUtility"
```

I was wondering if there was some design choice behind this that I haven't considered, or whether it
would be possible to add the method to the `INode` type, the same way it works in browsers?

**Additional context**
I am actually using happy-dom to test a web framework I am developing, and it involves comparing
generated HTML trees, hence the above.

The official implementation makes this tricky sometimes because of [whitespace between nodes](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace). I have
actually implemented a wrapper function around `nodeEquals` which first strips the two node trees of
any text nodes only containing whitespace. This makes comparison between two node trees more
meaningful in a test environment for generated HTML.

I'm not sure if the above function would have any use in `NodeUtilities` as well?

---

Happy to put in a pull request if any of the above seems desirable.

gangsthub:
Fixes #656.

Not sure if this is something you want to include. I'm just trying to help with this issue.

Here's the link to the spec: https://dom.spec.whatwg.org/#concept-node-equals

If you think I should change anything in the PR, let me know! I assigned the `minor` label, but I'll leave that to you to decide if it's correct. Cheers 🍻 !

### DIFF
--- a/packages/happy-dom/src/nodes/node/Node.ts
+++ b/packages/happy-dom/src/nodes/node/Node.ts
@@ -623,7 +623,7 @@ export default class Node extends EventTarget implements INode {
 					/**
 					 * 5.2.1.1. If attr equals attr1, then return the result of adding DOCUMENT_POSITION_IMPLEMENTATION_SPECIFIC and DOCUMENT_POSITION_PRECEDING.
 					 */
-					if (NodeUtility.nodeEquals(<IAttr>attr, attr1)) {
+					if (NodeUtility.isEqualNode(<IAttr>attr, attr1)) {
 						return (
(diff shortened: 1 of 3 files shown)


capricorn86 (approved):
Nice! :star2:

PR #698, opened 2023-01-25
#697@patch: Upgrade jest dependencies to 29.4.0.

### LINKED ISSUE #697: Jest 29 not supported by jest-environment
**Describe the bug**

@happy-dom/jest-environment uses v27 of various jest packages.

When using in an app that has jest v29, tests fail with:

```
TypeError: Cannot read properties of undefined (reading 'bind')

      at Runtime._createJestObjectFor (node_modules/jest-runtime/build/index.js:2099:64)
```

The [referenced line in jest-runtime](https://github.com/facebook/jest/blob/9fa2a463a0d5b1a7763ef4a28004db082b8bcf6e/packages/jest-runtime/src/index.ts#L2322) accesses `moduleMocker.replaceProperty`, which doesn't exist on @happy-dom/jest-environment's jest-mock v27.

Updating jest-mock (and the other jest deps) in @happy-dom/jest-environment to v29 fixes this issue for me, and I'll submit a PR making that update.

**To Reproduce**
1. Create or clone an app using Jest v27 and @happy-dom/jest-environment (such as https://github.com/chanzuckerberg/edu-stack).
2. Run tests, which are green.
3. Upgrade Jest to v29.
4. Run tests, which are red.

**Expected behavior**
@happy-dom/jest-environment works in Jest v29.

ahuth:
Fixes #697 by upgrading all jest dependencies (especially the ones in @happy-dom/jest-environment) to 29.4.0.

I also verified that @happy-dom/jest-environment works when being used in Jests v27 and v29.

### DIFF
--- a/packages/jest-environment/package.json
+++ b/packages/jest-environment/package.json
@@ -49,11 +49,11 @@
 		"testEnvironment": "./lib"
 	},
 	"dependencies": {
-		"@jest/environment": "^27.5.1",
-		"@jest/fake-timers": "^27.5.1",
-		"@jest/types": "^27.5.1",
-		"jest-mock": "^27.5.1",
-		"jest-util": "^27.5.1",
+		"@jest/environment": "^29.4.0",
+		"@jest/fake-timers": "^29.4.0",
+		"@jest/types": "^29.4.0",
+		"jest-mock": "^29.4.0",
+		"jest-util": "^29.4.0",
 		"happy-dom": "^0.0.0"
 	},
 	"devDependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,9 +11,9 @@
 				"@angular/core": "^10.0.7",
 				"@angular/platform-browser": "^10.0.7",
 				"@angular/platform-browser-dynamic": "^10.0.7",
-				"@jest/environment": "^27.5.1",
-				"@jest/fake-timers": "^27.5.1",
-				"@jest/types": "^27.5.1",
+				"@jest/environment": "^29.4.0",
+				"@jest/fake-timers": "^29.4.0",
+				"@jest/types": "^29.4.0",
 				"@lit/reactive-element": "^1.3.1",
 				"@testing-library/react": "^12.1.2",
 				"@testing-library/user-event": "^14.4.3",
@@ -24,8 +24,8 @@
 				"css.escape": "^1.5.1",
 				"glob": "^7.2.0",
 				"he": "^1.2.0",
-				"jest-mock": "^27.5.1",
-				"jest-util": "^27.5.1",
+				"jest-mock": "^29.4.0",
+				"jest-util": "^29.4.0",
 				"jquery": "^3.6.0",
 				"lit": "^2.2.1",
 				"lit-element": "^3.2.0",
@@ -41,7 +41,7 @@
 			},
 			"devDependencies": {
 				"@types/he": "^1.1.2",
-				"@types/jest": "^27.4.1",
+				"@types/jest": "^29.4.0",
 				"@types/node": "^15.6.0",
 				"@typescript-eslint/eslint-plugin": "^5.16.0",
 				"@typescript-eslint/parser": "^5.16.0",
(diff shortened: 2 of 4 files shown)


jereklas:
In my opinion this should have been solved in a different way. I would have preferred this just become a series of peerDependencies, instead of dependencies. You would then update the installation steps to be the following. 

```yarn add --dev @happy-dom/jest-environment @types/jest @jest/environment @jest/fake-timers jest-mock jest-util```

There's nothing in the class that really cares about what versions of jest is being used. So why force users to a particular jest version just to use happy dom? @ahuth

**Comment by jereklas** [comment]
Also, `@types/jest` should not be in the `dependencies` list, it's just type information and thus should be a `devDependency` so that it's not installed for end users.

File: packages/jest-environment/package.json [hunk 1]
```diff
@@ -49,11 +49,11 @@
...
 	},
 	"dependencies": {
-		"@jest/environment": "^27.5.1",
-		"@jest/fake-timers": "^27.5.1",
-		"@jest/types": "^27.5.1",
-		"jest-mock": "^27.5.1",
-		"jest-util": "^27.5.1",
+		"@jest/environment": "^29.4.0",
+		"@jest/fake-timers": "^29.4.0",
+		"@jest/types": "^29.4.0",
```
jereklas, line 54: this shouldn't be a dependency, it needs to be a devDependency so that it's not installed for end users of the package.

jereklas: same goes for `@jest/environment`?

ahuth:
> There's nothing in the [class](https://github.com/capricorn86/happy-dom/blob/master/packages/jest-environment/src/index.ts) that really cares about what versions of jest is being used. So why force users to a particular jest version just to use happy dom?

I see where you're coming from, @jereklas. But I don't think that's the solution here.

To use Jest I'd expect to have to install
1. Jest
2. maybe @types/jest
3. a jest environment (node, jsdom, or happy-dom).

Not the fake timers, mocking, utils, other jest sub-packages, etc.

In fact, both `jest-environment-node` and `jest-environment-jsdom` have dependencies on the same libraries. So I think it makes sense that @happy-path/jest-environment does as well.

Maybe it would be clearer if @happy-dom/jest-environment had the same major version number as the version of Jest it worked with.

> Note: This still working with jest 27 is maybe just a happy coincidence

For sure. Definitely a possibility that a future version of Jest breaks it.

jereklas:
I guess chalk it up to a difference of opinion of how dependencies are managed. At the end of the day, I have a workaround irrespective of how this gets resolved that can allow me to have all versions of the packages align to exactly the versions I want (using pnpm to override what the package.json file says to install). 

**Below are a continuation of my opinion, feel free to skip over if you'd like :)**

I would much rather be told to install 2-3 more things upon installation so that I have control over keeping everything used by my tooling version aligned, then hope for everything to "magically" work today while all versions are aligned, and then fail tomorrow when they're no longer aligned. 

To continue my concerns, updating to ^29.4.0 and working with 27.x.x is "working" only because the v28 and v29 versions support enough of the `ModuleMocker`, `jest-util`, `LegacyFakeTimers`, and `ModernFakeTimers` APIs. Without reading and understanding everything that's changed in Jest between 27 and 29, I cannot say without a doubt that having version 27 of Jest installed in an environment and this package using version 29 of the auxiliary packages means that everything a user could use in Jest is actually supported with the happy-dom jest environment.

This just speculation, but I wouldn't be surprised at all if 90% of Jest usages just work with Jest 27.x.x being used with the 29.x.x support packages, but then there could be ~10% of Jest that flat out fail, and those users will be stuck wondering why this package doesn't work. 

All of the above happens because people don't actually know what their dependencies are, and having someone install more of them while slightly a pain, at least makes everything less "magic".

capricorn86:
@jereklas @ahuth it has happens that Jest changes how the environment is implemented. It has happened a couple of times in the past.

One solution would be to split out `@happy-dom/jest-environment` to its own repo and use the same versions as Jest for the package. Then `happy-dom` can be a peer dependency instead as there are less dependencies towards how Happy DOM is implemented compared to Jest here.

Maybe this should be handled in a separate ticket in that case? I have created #703 for it.

packages/jest-environment/package.json
capricorn86: @jereklas I think `@jest/environment` has to be a dependency here. `@jest/types` can be a dev dependency.

capricorn86 (approved):

ahuth:
> One solution would be to split out @happy-dom/jest-environment to its own repo and use the same versions as Jest for the package. Then happy-dom can be a peer dependency instead as there are less dependencies towards how Happy DOM is implemented compared to Jest here.

👍 Makes sense to me

PR #702, opened 2023-01-26
#701@patch: fix handling of falsy CustomEvent detail value

### LINKED ISSUE #701: CustomEvent does not take falsy detail value
**Describe the bug**

**To Reproduce**
1. Create a CustomEvent `new CustomEvent(eventName, { detail: undefined })`
2. Wondering why `detail` is null
image

**Expected behavior**
The `detail` value should return the given input `detail` value.

**Device:**
 - Version happy-dom@8.1.5

**Additional context**
Wrong handling of `detail` with falsy check:
https://github.com/capricorn86/happy-dom/blob/4c7ed339323f993d8e1c01c204d524594da15c1c/packages/happy-dom/src/event/events/CustomEvent.ts#L19-L22

CSchulz:
fixes #701

### DIFF
--- a/packages/happy-dom/src/event/events/CustomEvent.ts
+++ b/packages/happy-dom/src/event/events/CustomEvent.ts
@@ -17,7 +17,7 @@ export default class CustomEvent extends Event {
 		super(type, eventInit);
 
 		if (eventInit) {
-			this.detail = eventInit.detail || null;
+			this.detail = eventInit.detail !== undefined ? eventInit.detail : null;
 		}
 	}
 


CSchulz:
I have already checked for some existing tests about the CustomEvent, but didn't find anything.
If you tell me the location to put it, I will create one.

packages/happy-dom/src/event/events/CustomEvent.ts
```diff
@@ -17,7 +17,7 @@ export default class CustomEvent extends Event {
 		super(type, eventInit);
 
 		if (eventInit) {
-			this.detail = eventInit.detail || null;
+			this.detail = eventInit.detail != null ? eventInit.detail : null;
```
capricorn86, line 20:
```suggestion
			this.detail = eventInit.detail != null && eventInit.detail != undefined ? eventInit.detail : null;
```

capricorn86 (changes requested):

packages/happy-dom/src/event/events/CustomEvent.ts
capricorn86: `detail` should not be able to be "undefined". In this scenario it should become `null`.

capricorn86:
Thanks for your contribution @CSchulz! :slightly_smiling_face: 

The build is failing as you have to fix the commit message to follow [this convention](https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md#commit-convention). I believe you can do a `git commit --amend` to change it.

packages/happy-dom/src/event/events/CustomEvent.ts
CSchulz: The equality check covers null and undefined cases. There is no further check needed.

Is there any documentation stating we have to use null in CustomEvent.detail? I am just wondering because in our code base we are also using undefined. Would be good to know if it is wrong.

capricorn86: According to [spec](https://dom.spec.whatwg.org/#interface-customevent), the default value is "null". Passing "undefined" is considered as not passing any value, and it should then use the default value.

I have also looked how the browser behaves and it will set detail to "null" if we pass "undefined" as value. 

```javascript
var customEvent = new CustomEvent('myCustomEvent', { detail: undefined });

// Will output "null"
console.log(customEvent.detail);
```

CSchulz: Thanks for the explanation.

I think we do not need to change the condition, but I would change if you want to.

> "null == undefined" true
> "null != null" false
> "null != undefined" false

capricorn86: I got confused by the double (!=) not triple condition (!==). I have changed the my suggestion above. I think using triple condition is less error prone :sweat_smile:

capricorn86 (approved):

packages/happy-dom/src/event/events/CustomEvent.ts
CSchulz: Yeah I was thinking about that.

Closed by accident?

PR #710, opened 2023-02-01
#709@patch: Properly detect `ShadowRoot` boundary when constructing `composedPath`

### LINKED ISSUE #709: `composedPath` is incorrect for composed bubbling events originating from `HTMLAnchorElement`
**Describe the bug**
There's an issue where `Event.composedPath` will be constructed incorrectly if `event.target` is `HTMLAnchorElement` with `href` attribute being set.

**To Reproduce**
1. Create an <a> element with `href` attribute set to a valid domain, e.g. `https://example.com` and it attach to `body`.
2. Add event listener for created element, e.g. `click`.
3. Trigger `click` event on this element. Event should have both `bubbles` and `composed` set to `true`.
4. Inspect value of `event.composedPath()` inside `click` event handler.

**Expected behavior**
Composed path should be the following array: `[HTMLAnchorElement, document.body, document.documentElement, document, window]`.

**Additional context**
Looks like `composedPath` checks `eventTarget.host` to detect `IShadowRoot` boundary. Unfortunately `HTMLAnchorElement` also has [`host` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/host) which breaks `while` loop through parent nodes of the event target.

https://github.com/capricorn86/happy-dom/blob/62085eba63a46c5e7716e52d50af82b6ee499fee/packages/happy-dom/src/event/Event.ts#L71-L78

romansp:
Fixes #709.

Not exactly sure about this `instanceof ShadowRoot` check with an extra import as I couldn't find other similar usages the codebase. Will be happy to update this PR If there's a more desirable pattern.

### DIFF
--- a/packages/happy-dom/src/css/declaration/utilities/CSSStyleDeclarationElementStyle.ts
+++ b/packages/happy-dom/src/css/declaration/utilities/CSSStyleDeclarationElementStyle.ts
@@ -129,7 +129,10 @@ export default class CSSStyleDeclarationElementStyle {
 				}
 
 				styleAndElement = { element: null, cssTexts: [] };
-			} else if ((<IShadowRoot>styleAndElement.element).host) {
+			} else if (
+				styleAndElement.element.nodeType === NodeTypeEnum.documentFragmentNode &&
+				(<IShadowRoot>styleAndElement.element).host
+			) {
 				const styleSheets = <INodeList<IHTMLStyleElement>>(
 					(<IShadowRoot>styleAndElement.element).querySelectorAll('style,link[rel="stylesheet"]')
 				);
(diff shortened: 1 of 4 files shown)


capricorn86 (approved):

capricorn86:
Thank you for your contribution @romansp! :star: 

Good that you found this bug :sweat_smile: 

The biggest reason for avoiding checking with `instanceof` is to avoid the import as it very easy becomes circular dependencies because of the nature of the DOM (Document -> Node -> Document -> Node).

I have made a commit which checks with `nodeType` and I also found a couple of more places with similar problems. It should solve the problem even if it would be even better to check specifically for "ShadowRoot".

PR #717, opened 2023-02-05
Migrate lerna to turborepo

nmengual:
Hi, we have been reviewing your project and we find it very interesting. Seeing that you used lerna to manage the monorepo, we thought that by migrating it to turborepo you could gain a bit of agility both locally and running on github actions. If you start using turborepo, you can cache both local and ci-cd executions and and run the scripts in parallel. There are some more advantages like using npm to manage the workspace and not that lerna has to do it.

We leave you some benchmarks for you to compare and see if you find it interesting.

Executions on github actions:

First Execution: 1m 55s (average)
Second execution with turborepo cache: 29s (average)

If you want to investigate a little more, here you have the documentation and any questions we are here to help you.
https://turbo.build/repo/docs/core-concepts/caching


capricorn86 (approved):
Nice, this was planned, but has been down prioritized. Great contribution :star:

Issue #512, opened 2022-06-20
Selector with dot fails


vrslev:
Selector by attribute (`[foo="bar.baz"]`) works in browser or using JSDom, but fails with happy-dom.

JSDom outputs `HTMLDivElement {}`:

```js
const { JSDOM } = require("jsdom");

const {
  window: { document },
} = new JSDOM();

document.body.innerHTML = `
    <div name="foo.bar"></div>
`;

const result = document.querySelector('[name="foo.bar"]');
console.log(result);
```

happy-dom returns `null`, when `HTMLElement` is expected:

```js
const { Window } = require("happy-dom");

const { document } = new Window();

document.body.innerHTML = `
    <div name="foo.bar"></div>
`;

const result = document.querySelector('[name="foo.bar"]');
console.log(result);
```

capricorn86:
Relates to #715

capricorn86:
Thanks for reporting @vrslev! :slightly_smiling_face: 

This has been fixed now.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v8.2.6

PR #719, opened 2023-02-05
#512@trivial: Selector with dot fails 

EstDavid:
We have done the following:
1) Fixed identify the css classes in SelectorItem. 
2) added a test to check the BUG correction.


capricorn86:
Thanks for your contribution @EstDavid! :star2: 

This was fixed by @btea in #727.

Sorry that you put time on this when it was fixed in another PR.

Issue #421, opened 2022-03-17
incorrect insertBefore work


seliverstov-logema:
Hey! 

There is a incorrect work of insertBefore

```
const d = document.createElement('div');
d.innerHTML = '<span id="a"></span><span id="b"></span><span id="c"></span><span id="d"></span>';

const a = d.querySelector('#a');
const b = d.querySelector('#b');

d.insertBefore(a, b);

console.log(d.innerHTML)
```

in it has to show not changed order of elements

but it shows `<span id="b"></span><span id="a"></span><span id="c"></span><span id="d"></span>`

capricorn86:
Thanks for reporting @seliverstov-logema and @EstDavid! :slightly_smiling_face: 

The issue has been fixed.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v9.7.1

PR #720, opened 2023-02-05
#421@trivial: Added 'Example correct use of insertBefore'.

EstDavid:


capricorn86:
@EstDavid I am not sure about this one. It seems like the test is passing? Is it working as intended then?

capricorn86:
@EstDavid this issue has been fixed in another PR:
https://github.com/capricorn86/happy-dom/pull/864

PR #752, opened 2023-02-14
#724@minor: Add toggleAttribute.

### LINKED ISSUE #724: toggleAttribute is not a function
**Describe the bug**
When attempting to access function `toggleAttribute` after `.querySelector`, an error is displayed:

```
container.toggleAttribute('disabled');
          ^

TypeError: container.toggleAttribute is not a function
    at file:///D:/Projects/happ-dom-repro1/index.js:9:11
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:530:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)

Node.js v18.12.0
```

According to MDN, this function should be supported:
https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute

**To Reproduce**
Run this code with `node index.js` after basing it off the basic example:

```js

const window = new Window();

document.body.innerHTML = '<div class="container"></div>';

container.toggleAttribute('disabled');
```

**Expected behavior**
Attributes should be toggle on/off as expected of the spec from MDN.

**Device:**
 - OS: Windows 11

btea:
close #724

### DIFF
--- a/packages/happy-dom/src/nodes/element/Element.ts
+++ b/packages/happy-dom/src/nodes/element/Element.ts
@@ -592,6 +592,30 @@ export default class Element extends Node implements IElement {
 	}
 
 	/**
+	 * Toggle an attribute.
+	 * Returns `true` if attribute name is eventually present, and `false` otherwise.
+	 *
+	 * @param name A DOMString specifying the name of the attribute to be toggled.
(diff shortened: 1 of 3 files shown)


capricorn86 (approved):

PR #754, opened 2023-02-15
#0@trivial: Update ts-jest config remove warn.

btea:
rel https://github.com/kulshekhar/ts-jest/pull/3780


capricorn86 (approved):
Nice! :star2:

PR #756, opened 2023-02-15
#755@patch: Replace properties by getter in Navigator.

### LINKED ISSUE #755: Error: Property onLine does not have access type get
**Describe the bug**
If you are switching from jsdom to happy-dom you will find differences in the implementation of how the navigator is used.

With jsdom you are able to use some spy on the navigator, happy-dom does not support it:

`jest.spyOn(window.navigator, 'onLine', 'get').mockReturnValue(true);`

**To Reproduce**
Add the code line to a spec and run it.

**Expected behavior**
A switch between jsdom and happy-dom should be possible without any changes in the source code.

**Additional context**
Compare with jsdom
https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/navigator/Navigator-impl.js
https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/navigator/NavigatorOnLine-impl.js

CSchulz:
Fixes #755


capricorn86 (approved):

Issue #759, opened 2023-02-16
Spurious HTML comments in embedded scripts


fcapolini:
**Describe the bug**

The content of `<script>` and `<style>` tags is parsed with the same logic used for all other tags, which considers `<!--`, `<!` and `<?` as HTML comment markers.

Because of that this fragment from [highlight.js](https://github.com/highlightjs):

```javascript
{className:"meta",begin:/<![a-z]/,end:/>/,contains:[t,i,l,c]}
```

is erroneously parsed and is then serialized as:

```javascript
{className:"meta",begin:/<!--[a-z]/,end:/-->/,contains:[t,i,l,c]}
```

which breaks the script.

**To Reproduce**

```typescript
	const window = new Window({
		settings: {
			disableJavaScriptFileLoading: true,
			disableJavaScriptEvaluation: true,
			disableCSSFileLoading: true,
			enableFileSystemHttpRequests: false
		}
	});
	const domParser = new window.DOMParser();

	const newDocument = domParser.parseFromString(
		`<html>
			<body>
				<script>
					var test = {className:"meta",begin:/<![a-z]/,end:/>/,contains:[t,i,l,c]};
				</script>
			</body>
		</html>`,
		'text/html'
	);

	const output = new XMLSerializer().serializeToString(newDocument);
	// Spurious comment `<!--[a-z]/,end:/-->` appears in output markup
```

**Expected behavior**

In the test code above, the output should preserve the original JS code.

capricorn86:
Thanks again for your contribution @fcapolini! :star:

fcapolini:
Perfect, thanks to you! Super quick 💯

PR #760, opened 2023-02-16
#759@patch: Spurious HTML comments in embedded scripts.

fcapolini:
Solves #759, added test case in `DOMParser.test.ts`.

### DIFF
--- a/packages/happy-dom/src/xml-parser/XMLParser.ts
+++ b/packages/happy-dom/src/xml-parser/XMLParser.ts
@@ -112,7 +112,9 @@ export default class XMLParser {
 				} else {
 					stack.pop();
 					parent = stack[stack.length - 1] || root;
-					parentTagName = (<IElement>parent).tagName ? (<IElement>parent).tagName.toLowerCase() : null;
+					parentTagName = (<IElement>parent).tagName
+						? (<IElement>parent).tagName.toLowerCase()
+						: null;
 					parentUnnestableTagName = this.getUnnestableTagName(parent);
 
 					lastTextIndex = markupRegexp.lastIndex;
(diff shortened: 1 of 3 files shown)


packages/happy-dom/src/xml-parser/XMLParser.ts
```diff
@@ -51,7 +52,14 @@ export default class XMLParser {
 
 				if (parent && match.index !== lastTextIndex) {
 					const text = data.substring(lastTextIndex, match.index);
-					this.appendTextAndCommentNodes(document, parent, text);
+					if (
+						parent.nodeType === Node.ELEMENT_NODE &&
+						PlainTextElements.includes((<IElement>parent).tagName.toLowerCase())
```
capricorn86, line 57: I think we could store the parent tag name to improve performance a bit as we only need to convert to lower case once.

We also need to check that the parent in fact has a tag name as it can also be a Document or DocumentFragment.

Example:
```javascript
const parentTagName = <IElement>parent).tagName ? <IElement>parent).tagName.toLowerCase() : null;
```

We can also then add a check for if the `parentTagName` is set to the if statement at row 57.

capricorn86 (approved):
Great contribution! :star2:  I just have a small comment.

capricorn86 (changes requested):
Sorry selected approved, but meant to select request changes

packages/happy-dom/src/xml-parser/XMLParser.ts
fcapolini: Sure, excellent suggestions, and thank you for your quick response 👍

capricorn86 (approved):

PR #762, opened 2023-02-18
#761@minor: Window add SVGGraphicsElement.

### LINKED ISSUE #761: `instanceof SVGGraphicsElement` causes typeError
**Describe the bug**
I am using happy-dom with vitest and I noticed after upgrading to antd@5.1.2 A bunch of my unit tests are failing with:

this error:
```
TypeError: Right-hand side of 'instanceof' is not an object
 ❯ _default node_modules/.pnpm/rc-util@5.27.1_biqbaboplfbrettd7655fr4n2y/node_modules/rc-util/lib/Dom/isVisible.js:14:15
```

At first I thought this was a bug in `rc-util` a library used by `antd`. But upon further inspection, it seems like it's a missing feature/case in `happy-dom`

This is the file that causes the error:

```js
var _default = function _default(element) {
  if (!element) {
    return false;
  }
  if (element instanceof HTMLElement && element.offsetParent) {
    return true;
  }
  if (element instanceof SVGGraphicsElement && element.getBBox) {
    var _element$getBBox = element.getBBox(),
      width = _element$getBBox.width,
      height = _element$getBBox.height;
    if (width || height) {
      return true;
    }
  }
  if (element instanceof HTMLElement && element.getBoundingClientRect) {
    var _element$getBoundingC = element.getBoundingClientRect(),
      _width = _element$getBoundingC.width,
      _height = _element$getBoundingC.height;
    if (_width || _height) {
      return true;
    }
  }
  return false;
};
```

The part that is causing the error using happy-dom is `element instanceof SVGGraphicsElement`. If I comment out this if statement all of my unit tests will pass without issues.

**To Reproduce**
1. Create a file that uses `instanceof SVGGraphicsElement`
2. Run it in a test environment using `happy-dom@8.4.3`

**Expected behavior**
parsing `instanceof SVGGraphicsElement` should not cause an error.

**Device:**
 - OS: MacOS
 - Browser: node
 - Version: "happy-dom": "8.4.3",
 
 **Temporary solution**
 For now I have switched back to use `JSdom` where this problem does not occur. Will set up happy-dom again once this issue is resolved.

btea:
close #761

### DIFF
--- a/packages/happy-dom/src/window/Window.ts
+++ b/packages/happy-dom/src/window/Window.ts
@@ -28,6 +28,7 @@ import HTMLBaseElement from '../nodes/html-base-element/HTMLBaseElement';
 import HTMLDialogElement from '../nodes/html-dialog-element/HTMLDialogElement';
 import SVGSVGElement from '../nodes/svg-element/SVGSVGElement';
 import SVGElement from '../nodes/svg-element/SVGElement';
+import SVGGraphicsElement from '../nodes/svg-element/SVGGraphicsElement';
 import HTMLScriptElement from '../nodes/html-script-element/HTMLScriptElement';
 import HTMLImageElement from '../nodes/html-image-element/HTMLImageElement';
 import { default as ImageImplementation } from '../nodes/html-image-element/Image';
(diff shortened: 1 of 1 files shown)


capricorn86 (approved):
Nice! :star:

Issue #700, opened 2023-01-26
Support AddEventListenerOptions


mrcaidev:
**Is your feature request related to a problem? Please describe.**

[`EventTarget.addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) can accept an optional third argument `options`, where developers can tweak the listener's behaviors.

In my codebase, I have to test these options can work as expected. After I switched from `jsdom` to `happy-dom` today, I found my test suites broken, because `options` are currently not supported in `happy-dom`.

**Describe the solution you'd like**

It would be great if we can support `options` in `EventTarget.addEventListeners`.

And I guess this can be achieved by modifying the inner data structure of [`EventTarget`](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/event/EventTarget.ts).

Current:

```ts
public readonly _listeners: {
  [k: string]: (((event: Event) => void) | IEventListener)[];
} = {}
```

Ideal:

```ts
public readonly _listeners: {
  [k: string]: {
    fn: ((event: Event) => void) | IEventListener;
    options: AddEventListenerOptions
  }[];
} = {}
```

And update other methods accordingly.

**Describe alternatives you've considered**

None.

**Additional context**

None.

capricorn86:
Thanks for reporting @mrcaidev! :slightly_smiling_face: 

@btea has made a fix which adds support for the "once" option :star2: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v8.6.0

capricorn86:
I will leave this open for the other options.

karpiuMG:
@capricorn86 can it be closed? all options seem to be implemented now.

<img width="608" alt="Image" src="https://github.com/user-attachments/assets/c159d93c-b3ef-42dd-8994-da466c3b0438" />

PR #763, opened 2023-02-18
#700@minor: Add event listener options.

btea:
rel #700

### DIFF
--- a/packages/happy-dom/src/event/EventTarget.ts
+++ b/packages/happy-dom/src/event/EventTarget.ts
@@ -41,8 +41,14 @@ export default abstract class EventTarget implements IEventTarget {
 		listener: ((event: Event) => void) | IEventListener
 	): void {
 		if (this._listeners[type]) {
-			const _listeners = this._listeners[type].map((l) => l.fn);
-			const index = _listeners.indexOf(listener);
+			const _listeners = [];
+			let index = -1;
+			for (let i = 0; i < this._listeners[type].length; i++) {
+				_listeners.push(this._listeners[type][i].fn);
+				if (listener === this._listeners[type][i].fn) {
+					index = i;
+				}
+			}
 			if (index !== -1) {
 				this._listeners[type].splice(index, 1);
 			}
--- a/packages/happy-dom/src/event/EventTarget.ts
+++ b/packages/happy-dom/src/event/EventTarget.ts
@@ -1,16 +1,17 @@
 import IEventListener from './IEventListener';
 import Event from './Event';
 import IEventTarget from './IEventTarget';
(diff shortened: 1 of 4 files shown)


packages/happy-dom/src/nodes/document/Document.ts
```diff
@@ -723,7 +723,7 @@ export default class Document extends Node implements IDocument {
 		this._isFirstWriteAfterOpen = true;
 
 		for (const eventType of Object.keys(this._listeners)) {
-			const listeners = this._listeners[eventType];
+			const listeners = this._listeners[eventType].map((listener) => listener.fn);
```
capricorn86, line 726: I think we can speed this up a bit by checking directly in the for loop instead as we can avoid the extra loop with the `map()`.

packages/happy-dom/src/event/EventTarget.ts
```diff
@@ -32,7 +41,8 @@ export default abstract class EventTarget implements IEventTarget {
 		listener: ((event: Event) => void) | IEventListener
 	): void {
 		if (this._listeners[type]) {
-			const index = this._listeners[type].indexOf(listener);
+			const _listeners = this._listeners[type].map((l) => l.fn);
```
capricorn86, line 44: For loops are faster than map(), so I think we are better of with a for loop here, but we could improve performance even further if we store options and listeners in separate objects.

Example:
```typescript
public readonly _options: {
	[k: string]: { once: boolean }[];
} = {};
public readonly _listeners: {
	[k: string]: (((event: Event) => void) | IEventListener)[];
} = {};

public addEventListener(
	type: string,
	listener: ((event: Event) => void) | IEventListener,
	options?: { once: boolean }
): void {
	this._listeners[type] = this._listeners[type] || [];
	this._options[type] = this._options[type] || [];
	this._listeners[type].push(listener);
	this._options[type].push(options || null);
}
```

capricorn86 (changes requested):
Great contribution! :slightly_smiling_face: Just found some minor things.

packages/happy-dom/src/nodes/document/Document.ts
btea: Yes, you are right. 👍

packages/happy-dom/src/event/EventTarget.ts
btea: Great.

capricorn86 (approved):

capricorn86:
@btea I made a commit with the proposed adjustments for improved performance. Please have a look if you want :slightly_smiling_face:

btea:
It looks very good. 👍

Issue #695, opened 2023-01-19
checked pseudo-class not working with querySelector


jgregor5:
Using happy-dom@8.1.4.
I understand happy-dom implements attribute selectors and pseudo-classes, doesn't it?

The following script does not print anything in the console.
This HTML and DOM manipulation works in any browser and also with jsdom.

```
import { Window } from 'happy-dom';

const window = new Window();
const document = window.document;

document.body.innerHTML = `
<form>
  <input type="radio" id="id1" name="op" value="one" checked="true"/>
  <input type="radio" id="id2" name="op" value="two"/>
  <input type="submit" id="submitbutton" value="Submit"/>
</form>
`;

// DOM manipulation

const twoEl = document.querySelector("input[value='two']");
twoEl.checked = true;

const elems = document.querySelectorAll("input[name='op']:checked");
elems.forEach(item => {
  console.log(`checked item is ${item.value}`);
});
```

capricorn86:
This has finally been fixed.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/edit/v9.0.0

PR #764, opened 2023-02-20
#695@patch: Pseudo checked match element.

btea:
rel #695

### DIFF
--- a/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
+++ b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
@@ -558,6 +558,24 @@ export default class HTMLInputElement extends HTMLElement implements IHTMLInputE
 		if (!checked) {
 			this.removeAttributeNS(null, 'checked');
 		} else {
+			// Need to remove all other radio buttons checked with the same name
+			const name = this.getAttributeNS(null, 'name');
+			if (name) {
+				const form = this.form;
+				if (form) {
+					const elements = form.elements;
+					for (let i = 0; i < elements.length; i++) {
+						const element = elements[i];
+						if (
+							element instanceof HTMLInputElement &&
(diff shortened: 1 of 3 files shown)


packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
```diff
@@ -558,6 +558,24 @@ export default class HTMLInputElement extends HTMLElement implements IHTMLInputE
 		if (!checked) {
 			this.removeAttributeNS(null, 'checked');
 		} else {
+			// Need to remove all other radio buttons checked with the same name
+			const name = this.getAttributeNS(null, 'name');
```
capricorn86, line 562: I'm working on a PR with a lot of Form/Input element related things, including fixing the "checked" property:
https://github.com/capricorn86/happy-dom/pull/677/files#diff-bb649479cc3389b9fff9883a1cc18a2b9edc08aac12cb0689ecabf13f21be16bL557

I have been working on it for quite some time, but my goal is to release it this week.

capricorn86 (changes requested):
Looks good! :star2:  I already have a fix for the "checked" property in a separate PR (see comment). I think we should wait for that fix instead of implementing a new :slightly_smiling_face:

packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
capricorn86: The PR contains so many fixes that it might break the UI of Github. In that case you can press "." on the keyboard to open the online editor to view it.

btea:
Oh, I didn't notice it. It's a huge change，awesome! 👍 
I'll close the pr.

capricorn86:
@btea I copied the logic you made regarding the query selector now to the #677 pull request.

PR #772, opened 2023-02-21
#771@patch: Don't bubble `focus` and `blur` events through DOM.

### LINKED ISSUE #771: `blur` and `focus` events should not bubble through DOM
**Describe the bug**
When triggering `focus/blur` events on the element its ancestor nodes also receive `focus/blur` when they should not.

Per MDN docs `focus` and `blur` events don't bubble in DOM. But related `focusin` and `focusout` do bubble.
https://developer.mozilla.org/en-US/docs/Web/API/Element/focus_event
https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event

**Expected behavior**
`focus` or `blur` events should not bubble in DOM.

**Device:**
 - Version 8.7.1

Will open a PR soon with updated test cases.

romansp:
Fixes #771.

Only updated existing `HTMLElement` tests that expect `bubbles` value of `blur/focus` events to be `false` from now.

Actual event bubbling is still covered by `Node.test.ts`.

### DIFF
--- a/packages/happy-dom/src/nodes/html-element/HTMLElement.ts
+++ b/packages/happy-dom/src/nodes/html-element/HTMLElement.ts
@@ -383,15 +383,18 @@ export default class HTMLElement extends Element implements IHTMLElement {
 
 		this.ownerDocument['_activeElement'] = null;
 
-		for (const eventType of ['blur', 'focusout']) {
-			const event = new FocusEvent(eventType, {
+		this.dispatchEvent(
+			new FocusEvent('blur', {
+				bubbles: false,
+				composed: true
+			})
+		);
+		this.dispatchEvent(
+			new FocusEvent('focusout', {
 				bubbles: true,
 				composed: true
-			});
-			event._target = this;
-			event._currentTarget = this;
-			this.dispatchEvent(event);
-		}
+			})
+		);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


packages/happy-dom/src/nodes/html-element/HTMLElement.ts
```diff
@@ -383,9 +383,13 @@ export default class HTMLElement extends Element implements IHTMLElement {
 
 		this.ownerDocument['_activeElement'] = null;
 
-		for (const eventType of ['blur', 'focusout']) {
+		const events = <const>[
+			['blur', false],
+			['focusout', true]
+		];
+		for (const [eventType, bubbles] of events) {
```
capricorn86, line 390: I think we can replace the for loop logic with just (the same for the other for loop):
```javascript
this.dispatchEvent(new FocusEvent('blur', {
	bubbles: false,
	composed: true
}));
this.dispatchEvent(new FocusEvent('focusout', {
	bubbles: true,
	composed: true
}));
```

Settings `event._target` and `event._currentTarget` doesn't seem to do anything as it will be set by `dispatchEvent()` anyway.

capricorn86 (changes requested):
Thank you for your contribution! :star2: I just have a small comment.

packages/happy-dom/src/nodes/html-element/HTMLElement.ts
romansp: Sounds good. Updated

capricorn86 (approved):
Nice! 🌟

PR #777, opened 2023-02-22
#776@patch: Add self equality check in node.contains method.

### LINKED ISSUE #776: Element.contains does not check itself
**Describe the bug**
The Element contains method does not check if the given element is the element itself resulting in wrong results.

**To Reproduce**
Pseudo code:
```
const element = queryElement();
expect(element.contains(element)).toBeTruthy();
```

**Expected behavior**
contains should return true if the given element is the same element running the method on.

**Device:**
 - Version 8.7.1

CSchulz:
fixes #776.


btea (approved):

capricorn86 (approved):
Great contribution! 🌟

PR #781, opened 2023-02-23
#780@minor: Add previousElementSibling and nextElementSibling.

### LINKED ISSUE #780: Add `previousElementSibling` and `nextElementSibling`

Add `nextElementSibling` and `previousElementSibling` for happy-dom.

btea:
fix #780


capricorn86 (changes requested):
Thank you for the contribution! :slightly_smiling_face: 

`previousElementSibling` and `nextElementSibling` has already been implemented on `Element`, which is according to spec.

See:
https://developer.mozilla.org/en-US/docs/Web/API/Element/previousElementSibling

btea:
oh, sorry i didn't notice this.

PR #797, opened 2023-02-26
#642@patch: Handle selector special value.

### LINKED ISSUE #642: TypeError: Cannot read properties of undefined (reading 'includes')
This may be similar to: https://github.com/capricorn86/happy-dom/issues/540

I'm getting this error from `node.querySelector(selector)` if `selector` is `undefined`. For context if it helps, I'm using Quasar and opening a QDialog modal in my tests:

```
TypeError: Cannot read properties of undefined (reading 'includes')
 ❯ Function.getSelectorParts node_modules/happy-dom/lib/query-selector/QuerySelector.js:180:44
 ❯ Function.querySelector node_modules/happy-dom/lib/query-selector/QuerySelector.js:42:34
 ❯ HTMLElement.querySelector node_modules/happy-dom/lib/nodes/element/Element.js:690:40
 ❯ Array.<anonymous> node_modules/quasar/src/components/dialog/QDialog.js:241:40
    239|           return
    240|         }
    241|         node = (selector !== '' ? node.querySelector(selector) : null)
       |                                        ^
    242|           || node.querySelector('[autofocus][tabindex], [data-autofocus][tabindex]')
    243|           || node.querySelector('[autofocus] [tabindex], [data-autofocus] [tabindex]')
 ❯ Module.removeFocusWaitFlag node_modules/quasar/src/utils/private/focus-manager.js:18:30
 ❯ showPortal node_modules/quasar/src/composables/private/use-portal.js:53:6
 ❯ node_modules/quasar/src/components/dialog/QDialog.js:208:9
```

Seems to be coming from here: https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/query-selector/QuerySelector.ts#L233

The code in Quasar seems to be relying on a `null` value being returned from a bad or non-existent selector and then checking for other selectors from there.

If I try the same thing in a browser, I get:

```javascript
> div = document.createElement('div')
<div>​</div>​

> div.querySelector(undefined)
null

> div.querySelector(null)
null
```

FWIW, it _does_ seem odd to me to pass in an `undefined` or `null` to `.querySelector()`, but on the other hand, the browser appears to handle it just fine and not throw an error unless the selector is an empty string.

btea:
fix #642


jdk2pq:
Looks great to me, and this would solve the issue I was having in #642. Thanks for putting this together @btea!

PR #806, opened 2023-03-05
#805@patch: Improve event listener.

### LINKED ISSUE #805: Event Dispatch - handleEvent undefined
**Describe the bug**
When adding multiple event listeners to an element with option `once: true`, then dispatching that same event will cause the dispatch loop to access the listener array out of bounds because the `max` has been pinned to the previous value.

```
TypeError: Cannot read properties of undefined (reading 'handleEvent')
    at HTMLElement.dispatchEvent (D:\Projects\happ-dom-repro1\node_modules\happy-dom\lib\event\EventTarget.js:59:30)
    at HTMLElement.dispatchEvent (D:\Projects\happ-dom-repro1\node_modules\happy-dom\lib\nodes\node\Node.js:383:35)
    at file:///D:/Projects/happ-dom-repro1/index2.js:14:6
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:530:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)
```

**To Reproduce**

Reproduction Code:
```js

const window = new Window();

document.body.innerHTML = '<div class="container"><input type="text" name="expiry"></div>';

const eventName = 'myEvent';
const text = document.querySelector('div');
const action1 = () => console.log('1');
const action2 = () => console.log('2');
text.addEventListener(eventName, action1, { once: true });
text.addEventListener(eventName, action2, { once: true });
text.dispatchEvent(new Event(eventName));
```

**Expected behavior**
Dispatching the event does not cause an unhandled exception.

**Device:**
 - Node.js v18.12.0

**Additional context**
Link: https://github.com/capricorn86/happy-dom/blob/13bcfe77ae9a202a93ce8fddcf4a7c1b43ecde16/packages/happy-dom/src/event/EventTarget.ts#L76

btea:
fix #805


capricorn86 (approved):

capricorn86:
It seems like the unit test is failing. Not sure why :man_shrugging:

btea:
I have fixed it 😀

capricorn86 (approved):

capricorn86:
> @capricorn86 I've fixed the test issue and it looks ok so far, can you review it again?

Looks good! :+1:

Issue #807, opened 2023-03-05
Template Content is not handled properly


ekwoka:
**Describe the bug**
Template contents are treated as nodes in Chrome/browsers. Happy-dom does not treat them as such, so when templates are cloned to use elsewhere, the contents is treated as a text node

**To Reproduce**
```js
import { Window } from 'happy-dom';

const window = new Window();
const { document, happyDOM } = window;

document.write('<div><template><span>test</span></template></div>');

console.log(document.querySelector('template').content.children[0].outerHTML);
```

This will error.

**Expected behavior**

Chrome logs:

```
<span>test</span>
```

It looks like `template` is marked as childless, which is not the case. Removing that does fix the above behavior. will pr later if I can ensure it solves the other issues with cloning template content I'm having.

for context, doing tests with happydom for AlpineJS using the `x-if` on template.

and alpine does this:

```js
const clone = templateEl.content.cloneNode(true).firstElementChild
```
which returns `null` in happydom.

capricorn86:
This has been fixed :slightly_smiling_face: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v9.1.3

PR #808, opened 2023-03-05
#807@patch: Fixes Template content not having nodes.

ekwoka:
solves #807 

Does not treat templates as childless (they aren't in actual dom).

The tests never really checked for this before, since they mostly verified with innerhtml only, and not checking the nodes themselves, as well as using `template.innerHTML =` would parse the html into nodes and append them, which had different behavior than parsing a template element from HTML directly, so even the tests claiming to check for `nodes` were not actually checking for nodes at all.

I'm not sure if there could be knock on effects, I don't know why the template was being treated as childless before.

### DIFF
--- a/packages/happy-dom/src/config/ChildLessElements.ts
+++ b/packages/happy-dom/src/config/ChildLessElements.ts
@@ -1,1 +1,1 @@
-export default ['style', 'script', 'template'];
+export default ['style', 'script'];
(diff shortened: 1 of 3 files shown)


capricorn86:

I don't know the reasoning for why it was added as childless. Must have been a misunderstanding of the spec by the person contributing somehow and I missed it in the review. Good find!

Issue #809, opened 2023-03-05
Template element has no siblings


ekwoka:
**Describe the bug**
Template elements always return null for `nextSibling` and `lastSibling` which is against the spec. It should behave as normal and return the next and last siblings.

**To Reproduce**
```js
import { Window } from 'happy-dom';

const window = new Window();
const { document, happyDOM } = window;

document.write('<template></template><span>test</span>');

console.log(document.querySelector('template').nextSibling);
```

**Expected behavior**
It should return the `span` node like it does in browsers

capricorn86:
This has been fixed :slightly_smiling_face: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v9.1.3

PR #810, opened 2023-03-05
#809@patch: Fixes Template siblings.

ekwoka:
Prevents overriding of sibling methods, enabled `after` and `before` to actually work on template tags (and not just append to the parent)


capricorn86:
I think we both started on this at the same time. That is perhaps why I didn't see your PRs :sweat_smile: Sorry about that.

Issue #811, opened 2023-03-07
SVGElement.ownerSVGElement cannot exit if not in svg


linghaoSu:
**Describe the bug**
When I was using SVG nodes and tried to use the ownerSVGElement property, I ended up in an infinite loop. Upon checking the code, I found the following issue.

![image](https://user-images.githubusercontent.com/56526981/223290250-e91aadf7-aad5-4462-9dc2-b28a5c0ae6ca.png)

**To Reproduce**
1. create a svg element, such as 'g',
2. gInstance.ownerSVGElement;
3. inifinite loop now....

**Expected behavior**
Return a correct SVG node or null value instead of an infinite loop.

capricorn86:
Thanks for your contribution @linghaoSu! :slightly_smiling_face:

PR #812, opened 2023-03-07
#811@patch: Fix ownerSVGElement insteadof inifite loop.

linghaoSu:
After the conditional statement, assign parent.parentNode to parent to facilitate the next step of evaluation.

fix https://github.com/capricorn86/happy-dom/issues/811


capricorn86 (approved):
Thank you for your contribution! :slightly_smiling_face:

PR #814, opened 2023-03-08
#813@patch: Unset previous cookie.

### LINKED ISSUE #813: Unable to unset a previous cookie without a value.
**Describe the bug**
Hi all, great lib btw.
I've noticed that if I have a cookie set and I attempt to unset it using the following code snippet it fails to remove the cookie and returns the previous value. 

```js
const { Window } = require("happy-dom");

const window = new Window();

// set cookie
document.cookie = "name=Dave; expires=Thu, 01 Jan 2025 00:00:00 GMT;";

// unset cookie
document.cookie = "name=; expires=Thu, 01 Jan 1970 00:00:00 GMT;";

// output name=Dave
console.log(document.cookie);
```

**Expected behavior**
I would expect that not setting a value is valid behaviour and the cookie should be removed.

I did find a workaround by using a value
`document.cookie = "name=unset; expires=Thu, 01 Jan 1970 00:00:00 GMT;`

I believe this is what may be causing the issue.
https://github.com/capricorn86/happy-dom/blob/13bcfe77ae9a202a93ce8fddcf4a7c1b43ecde16/packages/happy-dom/src/cookie/Cookie.ts#L38-L41

btea:
fix #813


capricorn86 (approved):

PR #821, opened 2023-03-11
#0@patch: Fixes missing property key.

ekwoka:
The CSS property manager had a duplicate key (and a missing one).

this changes the duplicate to the missing.


btea (approved):

capricorn86:

PR #822, opened 2023-03-11
#0@patch: Handles supports for Solidjs.

ekwoka:
This fix is necessary to handle rendering of solidjs apps.

Throwing an error breaks things, just returning false works fine. If happy-dom doesn't support anything, then just tell the stuff trying to call it that it's not supported.


capricorn86 (approved):

PR #823, opened 2023-03-11
code over

Titiksha2608:


capricorn86 (changes requested):
Please explain why we should remove the commit validation? It helps people checking the commits before pushing in the code. Otherwise the build will fail instead.

capricorn86:
If you have a problem contributing, you can check out this guide:
https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md

Or just reach out and I can explain how it works.

ekwoka:
Also, if really needed, for some reason, the commit can be done with `--no-validation` to bypass the commit hooks

capricorn86:
I will close this PR based on the answer in the comments.

PR #824, opened 2023-03-12
#818@patch: Convert special characters to HTML entities.

### LINKED ISSUE #818: `happy-dom` doesn't convert `&gt;` to `>` whereas `jsdom` does
**Describe the bug**

When using `happy-dom` to set and then view the contents of a DOM node via `node.textContent`, if the DOM node has a `&gt;` character, `happy-dom` will return its `textContent` as `&gt;` whereas `jsdom` and web browsers will return the `>` character. Looks like `happy-dom` isn't converting those values.

**To Reproduce**

The following code can be used to reproduce the problem:

```js
document.documentElement.innerHTML = '<div id="testnode">&gt;howdy</div>';
expect(document.getElementById('testnode').textContent).toBe('>howdy');
```

Here is a StackBlitz which runs that code in both `jsdom` and `happy-dom` -- note the difference in output: https://stackblitz.com/edit/vitest-dev-vitest-fujcfs?file=test%2Fhappydom.test.ts%3AL6,test%2Fjsdom.test.ts 

**Expected behavior**

I expect `happy-dom` to have the same behaviour as `jsdom` when setting and reading `&gt;` and `&lt;` characters that are injected into a DOM node's body.

**Screenshots**

Screenshot 2023-03-08 at 4 51 52 PM

**Device:**

N/A

btea:
fix #818


capricorn86 (approved):

I made some fixes to the code, so that it follows the same style of coding as other logic in Happy DOM. I also made some improvements to performance.

btea:
Looks great 👍

capricorn86:
> Looks great 👍
> 
> `HTML_ENTITIES` currently only lists a part, do we need to add other characters? reference [documents](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references)

That would be good 👍😊 But I guess the most prioritized list is in?

btea:
If adding all of them seems like too much, it may be enough for now to add characters that are common to the keyboard, such as `@`, `#`, `$`. . .

What do you think?

PR #826, opened 2023-03-12
Handles Passive listeners

ekwoka:
This PR:
- completes the behavior of passive event listeners.
- removes unneeded overriding of methods in UI Event class.


capricorn86:

capricorn86:
I think we both started on this at the same time. That is perhaps why I didn't see your PRs :sweat_smile: Sorry about that.

Issue #842, opened 2023-04-06
Declaration maps are missing.


ekwoka:
**Describe the bug**
When working with happydom and right clicking and selectiong "go to definition" in vscode, it leads to the .d.ts files instead of the actual source.

**To Reproduce**
Have any project using happydom, right click on something from happydom, and hit "go to definition"

**Expected behavior**
should go to the src ts file

capricorn86:
Thanks for your contribution! :slightly_smiling_face:

PR #843, opened 2023-04-06
#842@trivial: Adds declaration maps.

ekwoka:
this will build and include declaration maps alongside the declarations, allowing "go to definition" to lead to the source instead of a definition file. To see just the types "go to type definition" will still work.


capricorn86 (approved):
Nice! Thank you for your contribution :slightly_smiling_face:

PR #848, opened 2023-04-09
#846@patch: After add element correctly.

### LINKED ISSUE #846: getElementsByClassName return wrong results when innerHTML ends with /n
**To Reproduce**
```ts

const window = new Window();
document.body.innerHTML = `<div class="container"></div>/n`;
// document.body.innerHTML = `<div class="container"></div>
// `;
// document.body.innerHTML = `<div class="container"></div>`;

const div1 = document.createElement('div');
div1.classList.add('someClassName');
container.after(div1);

const div2 = document.createElement('div');
div2.classList.add('someClassName');
div1.after(div2);

const l1 = Array.from(document.getElementsByClassName('someClassName'));
console.log(l1.length); // output 3, wrong

const l2 = Array.from(document.querySelectorAll('.someClassName'));
console.log(l2.length); // output 2, OK

```

happy-dom: v9.1.9
node.js: v18.12.1

btea:
fix #846


capricorn86 (approved):

capricorn86:

There was a problem with the logic in `ElementUtility.insertBefore()` and apparently there was no unit test covering it.

I have made two commits for fixing the problem with the logic and also added a unit test that covers the test case.

capricorn86:
The commits are done with the wrong user as I forgot to change user from the computer generated one.

Issue #853, opened 2023-04-13
Method `Range.cloneContents`: endless loop


danielrentz:
**Describe the bug**
Method `Range.cloneContents` crashes with an endless loop in special situations.

**To Reproduce**
Consider HTML markup:
```
<div>
  <span>text</span>
  <span>text</span>
</div>
```
A range selecting from first text node to second text node will loop endlessly in the method `cloneContents`. Root cause seems to be a wrong initialization of a subrange cloning the initial part of the selection (`firstPartialContainedChild`).

**Expected behavior**
No crash.

**Additional context**
I will prepare a PR with unit test.

PR #854, opened 2023-04-13
#853@minor: Fix method Range.cloneContents

danielrentz:
Seems the method initializes the initial subrange (the first element containing the start of the selection) incorrectly (Range.ts, line 282):
- Currently, the subrange will be created from **end** of current range to end of `firstPartialContainedChild`.
- Obviously, the subrange should clone from **start** of original range to end of `firstPartialContainedChild`.

Minor additional fix: I removed the else-if branches in the `startOffset` and `endOffset` getters. Code would only run into the else-if branch if the `offset` property is already 0 (otherwise the initial if-branch would trigger and reset the `offset` property to 0). Therefore the else-if branch does not have any effect.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10119,6 +10119,7 @@
 			"dependencies": {
 				"css.escape": "^1.5.1",
 				"he": "^1.2.0",
+				"iconv-lite": "^0.6.3",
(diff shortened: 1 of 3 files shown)


capricorn86 (approved):
NIce find! 🌟

PR #855, opened 2023-04-13
#820@minor: Add SubmitEvent.

### LINKED ISSUE #820: Missing window.SubmitEvent
Presently, there is no `window.SubmitEvent` available.

Add one, as has been done for other common event types. Seems like the only attribute you'd need to add is a `SubmitEvent.submitter` property as per [these docs](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent).

**Describe alternatives you've considered**
The alternative is to construct a base `Event` and add a `submitter` to it manually.

danielrentz:
fix #820

- add class `SubmitEvent`
- `HTMLFormElement.requestSubmit` triggers a `SubmitEvent`
- adapted unit tests


capricorn86 (approved):
Nice contribution! 🌟

PR #858, opened 2023-04-13
#857@patch: Throw DOMException from Text.splitText.

### LINKED ISSUE #857: Text.splitText does not throw on invalid index
**Describe the bug**
Calling `Text.spliText` does not throw on invalid index as specified.

**To Reproduce**
`document.createTextNode("abc").splitText(-1)`

**Expected behavior**
Throws a `DOMexception`.

**Additional context**
I will prepare a PR.

danielrentz:
fix #857 and unit tests for `Text.splittext`:
- the DOMException needs to be actually thrown
- missing check for negative index
- added unit test

### DIFF
--- a/packages/happy-dom/test/nodes/text/Text.test.ts
+++ b/packages/happy-dom/test/nodes/text/Text.test.ts
@@ -1,5 +1,7 @@
 import Window from '../../../src/window/Window';
 import Document from '../../../src/nodes/document/Document';
+import DOMException from '../../../src/exception/DOMException';
+import Text from '../../../src/nodes/text/Text';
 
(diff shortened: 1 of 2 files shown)


packages/happy-dom/test/nodes/text/Text.test.ts
```diff
@@ -35,4 +38,22 @@ describe('Text', () => {
 			expect(clone.data).toBe(node.data);
 		});
 	});
+
+	describe('splitText()', () => {
+		it('Splits the text node.', () => {
+			const node = document.createTextNode('test');
```
capricorn86, line 44: :star:

PR #871, opened 2023-04-21
#870@patch: Convert HTML entities should include  semicolon.

### LINKED ISSUE #870: DOMParser ignoring semicolons in hexadecimal html entities
**Describe the bug**
Using DOMParser for a string with characters such as `&lt;` or `&gt;` does not remove the semicolon as does the browser version

**To Reproduce**
1. Create a test for this function:
```ts
function htmlDecode(input: string): string {
  const doc = new DOMParser().parseFromString(input, "text/html");
  return doc.documentElement.textContent;
}
```
Like so:
```ts
it("should decode HTML entities", () => {
  expect(htmlDecode('&#x3C;div&#x3E;Hello, world!&#x3C;/div&#x3E;')).toBe(
    '<div>Hello, world!</div>',
  );
});
```

image

**Expected behavior**
The DOMParser should remove the semicolons, as does the browser:
image

**Screenshots**
Added inline

**Device:**
 - OS: Windows 10
 - Browser Chrome/Vitest
 - Version Vitest 0.30.1, happydom 9.8.0

**Additional context**

btea:
fix #870

### DIFF
--- a/packages/happy-dom/src/nodes/character-data/CharacterDataUtility.ts
+++ b/packages/happy-dom/src/nodes/character-data/CharacterDataUtility.ts
@@ -2,23 +2,23 @@ import ICharacterData from './ICharacterData';
 
 const HTML_ENTITIES = [
 	{ regex: /&quot;/g, value: '"' },
-	{ regex: /&#34/g, value: '"' },
-	{ regex: /&#x22/g, value: '"' },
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Nice! :star2:

Issue #887, opened 2023-05-03
CSS comment with asterisk gets parsed as part of selectorText


htuomola:
**Describe the bug**
If CSS comment contains `*` character in the middle, the comment is not omitted and ends up being parsed as part of selectorText.
This, of course, makes selectorText invalid for later use.

**To Reproduce**
1. Add following CSS:
```css
/*
* Some comment here 
*/
.foo { color: red; }
```
2. Run happy-dom (well, running vitest with happy-dom)
3. Run fails with error similar to
```
DOMException: Invalid selector: "/*
* Some comment here 
*/
.foo"
 ❯ Function.getSelectorGroups node_modules/happy-dom/src/query-selector/SelectorParser.ts:143:10
 ❯ Function.match node_modules/happy-dom/src/query-selector/QuerySelector.ts:100:38
 ❯ CSSStyleDeclarationElementStyle.parseCSSRules node_modules/happy-dom/src/css/declaration/utilities/CSSStyleDeclarationElementStyle.ts:292:42
 ❯ CSSStyleDeclarationElementStyle.getComputedElementStyle node_modules/happy-dom/src/css/declaration/utilities/CSSStyleDeclarationElementStyle.ts:124:12
 ❯ CSSStyleDeclarationElementStyle.getElementStyle node_modules/happy-dom/src/css/declaration/utilities/CSSStyleDeclarationElementStyle.ts:59:16
 ❯ CSSStyleDeclaration.getPropertyValue node_modules/happy-dom/src/css/declaration/AbstractCSSStyleDeclaration.ts:193:37
```

**Expected behavior**
CSS is parsed correctly and commit is omitted.

** Analysis **
It's just a problem with CSSParser regex used for comments.
I have a test and a fix for CssParser ready and fill submit PR soon.

capricorn86:
@htuomola your fix has been released 🙂 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v9.10.7

PR #888, opened 2023-05-03
#887@patch: Fix parsing CSS comments with asterisks.

htuomola:
Fix CSS regex so comment content is a non-greedy "any character or whitespace" instead of the previous "anything but *".


capricorn86 (approved):
Thank you for your contribution! 🌟

Issue #890, opened 2023-05-03
`HTMLAnchorElement.toString()` should return `href` instead of `outerHTML`


Schleuse:
**Describe the bug**
`HTMLAnchorElement` has a specific implementation of `toString()` that differs from `Element`.
`Element.toString()` returns `outerHTML` where `HTMLAnchorElement.toString()` should return `href`.

[HTMLAnchorElement.toString()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/toString)

**To Reproduce**
```js
import { Window } from 'happy-dom';

const window = new Window();
const document = window.document;
const anchor = document.createElement('a');
anchor.setAttribute('href', 'https://github.com/capricorn86/happy-dom');
anchor.innerHTML = 'Happy DOM';

console.log({
  toString: anchor.toString(),
  href: anchor.href
});

/* prints:
{
  toString: '<a href="https://github.com/capricorn86/happy-dom">Happy DOM</a>',
  href: 'https://github.com/capricorn86/happy-dom'
}
*/

/* should be :
{
  toString: 'https://github.com/capricorn86/happy-dom',
  href: 'https://github.com/capricorn86/happy-dom'
}
*/


```

**Expected behavior**
`toString()` should return the same value as the `href` property

capricorn86:
@Schleuse your fix has been released 🙂 

Read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v9.10.8

PR #891, opened 2023-05-03
#890@patch: Fixes a issue where "HTMLAnchorElement.toString" returns outerHTML instead of the url

Schleuse:
Fixes `toString()` of `HTMLAnchorElement` as described in #890


capricorn86 (approved):
Thank you for your contribution @Schleuse! 🌟

PR #893, opened 2023-05-03
#892@patch: Implement `Request.formData()`.

### LINKED ISSUE #892: TypeError: request.clone(...).formData is not a function
**Describe the bug**
The method `formData()` is missing at Request interface and its implementation.

https://developer.mozilla.org/en-US/docs/Web/API/Request/formData

**To Reproduce**
```typescript
it('should work with form data', async () => {
  const formData = new FormData();
  formData.append('some', 'test');
  const request = new Request('http://foo.bar/baz', { method: 'post', body: formData });
  const formDataFromRequest = await request.formData();
  expect(formDataFromRequest).toBe(formDataFromRequest);
});
```

**Actual Behavior**
```
TypeError: request.formData is not a function
```

**Expected behavior**
test should succeed

**Additional context**
We start to migrate a project to remix and to implement the action/loader unit tests where we need to construct `Request`s manually and pass them the form data.

rolandjohann:
fixes #892


rolandjohann:
@capricorn86 thanks for the quick response and release of of the fix 🎉

PR #902, opened 2023-05-10
#0@patch: `&nbsp;` should convert to non-breaking space.

btea:


capricorn86:

I have looked into encoding HTML entities. It seems like it has to happen during the parsing of the HTML. When the HTML is serialized to a string it should be escaped.

I have implemented a solution for the problem in this PR:
https://github.com/capricorn86/happy-dom/pull/912

The PR I've implemented will solve the problem with HTML spaces, so I will close this PR.

btea:
Ok, I see. Thanks for the hard work!! 👍

adaboese:
How to opt-out from decoding of HTML entities?

Issue #908, opened 2023-05-11
DOMParser().parseFromString(...) doesn't decode html entities


mreinstein:
setup code for happy-dom:

```javascript
const window = new Window();
const doc = new window.DOMParser().parseFromString('<p>here is some</p> html el&#225;stica ', 'text/html');
console.log('text:', doc.body.textContent);
```

output:
```bash
text: here is some html el&#225;stica
```

similar code, in a browser:
```javascript
const doc = new window.DOMParser().parseFromString('<p>here is some</p> html el&#225;stica ', 'text/html');
console.log('text:', doc.body.textContent);
```

output:
```bash
text: here is some html elástica 
```

le bug? tested with happy-dom@9.12.0 and chrome@113

capricorn86:
Thank you for reporting @mreinstein! :slightly_smiling_face: 

I have implemented a fix for the issue now.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v9.16.0

mreinstein:
wow, the speed!

adaboese:
This change seems to cause a few issues https://github.com/capricorn86/happy-dom/issues/944#issuecomment-1869080464

PR #912, opened 2023-05-11
#908@minor: Adds support for endoing HTML entities to the XML parser.…

capricorn86:
… Adds support for escaping text to XMLSerializer.serializeToString().

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3520,6 +3520,17 @@
 				"once": "^1.4.0"
 			}
 		},
+		"node_modules/entities": {
+			"version": "4.5.0",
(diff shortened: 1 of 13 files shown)


adaboese:
How to opt-out from decoding of HTML entities?

btea:
@adaboese You can only use lower versions that have not yet been implemented or modify the source code. 
I'd like to ask, why do you want to do this? 🤔

adaboese:
The current implementation breaks HTML.

```html
<li>%3Chtml%3E</li>
```

becomes

```html
<li><html></li>
```

btea:
It seems to be a problem with `entities`.

Issue #916, opened 2023-05-13
markupRegexp doesn't parse comments properly if they contain a dash in the text


gtm-nayan:
**Describe the bug**
Given a comment like `<!-- svelte-xyz -->`, the parser will incorrectly include a trailing `--` in the textContent of the comment.

https://github.com/capricorn86/happy-dom/blob/07448050a759d55f220abbfe85951e9da7c1157c/packages/happy-dom/src/xml-parser/XMLParser.ts#LL14C1-L27C104

**To Reproduce**
Run `MARKUP_REGEXP.exec("<!-- a-b -->")`, third group will have `a-b --`

**Expected behavior**
The third group contains `a-b `

capricorn86:
Thank you for your contribution @gtm-nayan! :star2: 

Your fix has been released.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v9.18.2

PR #917, opened 2023-05-13
#916@patch: Handle comments with dash in text.

gtm-nayan:


packages/happy-dom/src/xml-parser/XMLParser.ts
```diff
@@ -24,7 +24,7 @@ import * as Entities from 'entities';
  * Group 8: End of start tag (e.g. ">" in "<div>").
  */
 const MARKUP_REGEXP =
-	/<([a-zA-Z0-9-]+)|<\/([a-zA-Z0-9-]+)>|<!--([^-]+)-->|<!--([^>]+)>|<!([^>]+)>|<\?([^>]+)>|(\/>)|(>)/gm;
+	/<([a-zA-Z0-9-]+)|<\/([a-zA-Z0-9-]+)>|<!--([^-]+(?:-[^-]+)*)-->|<!--([^>]+)>|<!([^>]+)>|<\?([^>]+)>|(\/>)|(>)/gm;
```
capricorn86, line 27:
```suggestion
	/<([a-zA-Z0-9-]+)|<\/([a-zA-Z0-9-]+)>|<!--([\s\S]+)-->|<!--([^>]+)>|<!([^>]+)>|<\?([^>]+)>|(\/>)|(>)/gm;
```

capricorn86: My suggestion is that we just capture all characters until "-->" instead.

capricorn86 (changes requested):

packages/happy-dom/src/xml-parser/XMLParser.ts
capricorn86: I have commited my suggestion now as I would like to release a fix as soon as possible :sweat_smile:

capricorn86:
Thank you for your contribution @gtm-nayan! :star2:

packages/happy-dom/src/xml-parser/XMLParser.ts
capricorn86: That suggestion did not work. The problem is that the RegExp you suggested doesn't cover when there multiple dashes (e.g. "").

I'm looking into another solution.

gtm-nayan: I think making it non-greedy should work so 

`/<([a-zA-Z0-9-]+)|<\/([a-zA-Z0-9-]+)>|<!--([\s\S]+?)-->|<!--([^>]+)>|<!([^>]+)>|<\?([^>]+)>|(\/>)|(>)/gm`

capricorn86:
@gtm-nayan I have changed the logic so that it will work with comments containing multiple dashes (e.g. `<!-- comment -a-b-c -->`).

capricorn86 (approved):

gtm-nayan:
Thanks for the fix.

PR #924, opened 2023-05-17
#0@patch: MIME type add apng.

btea:


capricorn86:
Thank you @btea for your contribution! :star2:

Issue #945, opened 2023-06-01
happy-dom doesn't support setter for HTMLInput.valueAsNumber property


malko:
**Describe the bug**
when testing with happy-dom you can get value using input.valueAsNumber but not set value with it as only getter are implemented

**To Reproduce**
1. declare a test with an input[type=number] element
2. try to set value using input.valueAsNumber
3. See error: TypeError: Cannot set property valueAsNumber of [object Object] which has only a getter

**Expected behavior**
expect the input value to be set properly

malko:
Preparing a PR for adding both support for valueAsNumber and valueAsDate setter

malko:
Will add valueAsDate in a second PR if valueAsNumber is approved.

capricorn86:
Your fix has been released :slightly_smiling_face: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v10.1.0

capricorn86:
Should we close this ticket or keep it open for valueAsDate?

malko:
I think we can create another issue for valueasdate and close this one if this is ok for you. Will work on it as soon as i can.

PR #946, opened 2023-06-04
#945@major Add support for HTMLInputElement set valueAsNumber

malko:
- Enhance HTMLInputElementValueSanitizer to be more compliant with specs at https://html.spec.whatwg.org/multipage/input.html (breaking change)
- Change HTMLInputElement get valueAsNumber to be more compliant with specs at https://html.spec.whatwg.org/multipage/input.html (breaking change)
- Add support for HTMLInputElement set valueAsNumber based on specs at https://html.spec.whatwg.org/multipage/input.html

### DIFF
--- a/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
+++ b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
@@ -797,7 +797,48 @@ export default class HTMLInputElement extends HTMLElement implements IHTMLInputE
 	 * @returns Number.
 	 */
 	public get valueAsNumber(): number {
-		return this.value ? parseFloat(this.value) : NaN;
+		const value = this.value;
+		if (!this.type.match(/^(range|number|date|datetime-local|month|time|week)$/) || !value) {
+			return NaN;
+		}
+		switch (this.type) {
+			case 'number':
+				return parseFloat(value);
+			case 'range': {
+				const number = parseFloat(value);
+				const min = parseFloat(this.min) || 0;
+				const max = parseFloat(this.max) || 100;
+				if (isNaN(number)) {
+					return max < min ? min : (min + max) / 2;
+				} else if (number < min) {
+					return min;
+				} else if (number > max) {
+					return max;
+				}
+				return number;
+			}
+			case 'date':
+				return new Date(value).getTime();
+			case 'datetime-local':
+				return new Date(value).getTime() - new Date(value).getTimezoneOffset() * 60000;
+			case 'month':
+				return (new Date(value).getFullYear() - 1970) * 12 + new Date(value).getMonth();
+			case 'time':
+				return (
+					new Date('1970-01-01T' + value).getTime() - new Date('1970-01-01T00:00:00').getTime()
+				);
+			case 'week': {
+				// https://html.spec.whatwg.org/multipage/input.html#week-state-(type=week)
+				const match = value.match(/^(\d{4})-W(\d{2})$/);
+				if (!match) {
+					return NaN;
+				}
+				const d = new Date(Date.UTC(parseInt(match[1], 10), 0));
+				const day = d.getUTCDay();
+				const diff = ((day === 0 ? -6 : 1) - day) * 86400000 + parseInt(match[2], 10) * 604800000;
+				return d.getTime() + diff;
+			}
+		}
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


malko:
This solves issue #945
If this PR is accepted I'll propose another one to enhance get valueAsDate and add support for set valueAsDate.

capricorn86:
Thank you for your contribution @malko! :star2: 

There are some merge conflicts and commits have "@major", but should be "@minor" as it is not considered a breaking change.

packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
```diff
@@ -23,6 +23,28 @@ import IShadowRoot from '../shadow-root/IShadowRoot';
 import NodeList from '../node/NodeList';
 import EventPhaseEnum from '../../event/EventPhaseEnum';
 
+/**
+ * Return iso-week number from given date
+ *
+ * @param date Date|number.
+ * @returns Iso-week string.
+ */
+export const dateIsoWeek = (date: Date | number): string => {
```
capricorn86, line 32: I think it is better if we create a separate utility. Perhaps it can be called HTMLInputDateUtility or something.

capricorn86 (changes requested):

packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
malko: Yes, it seems like a good idea, where would you like to put this utility in the file tree ?

capricorn86: I think in `packages/happy-dom/src/nodes/html-input-element/` to start with. If we start to use it some place else we can think about moving it.

capricorn86:
> 

We can do a squash merge and change the commit type there to solve the "@major" release version that should be "@minor" :slightly_smiling_face:

major:
Y'all keep up the good work. 👏

malko:
> > 
> 
> We can do a squash merge and change the commit type there to solve the "@major" release version that should be "@minor" slightly_smiling_face

Will do with other suggested changes during the week

malko:
@capricorn86 I have some trouble with commit messages starting with # while rebasing, what's the correct git config to use in order to let git accept those # at the beginning of line without complaints ? Any help appreciated on this.

malko:
> @capricorn86 I have some trouble with commit messages starting with # while rebasing, what's the correct git config to use in order to let git accept those # at the beginning of line without complaints ? Any help appreciated on this.

Ok setting core.commentChar to auto does the trick, it would be nice to add a note in the contribution guide for new comers to the project.

capricorn86:
> > @capricorn86 I have some trouble with commit messages starting with # while rebasing, what's the correct git config to use in order to let git accept those # at the beginning of line without complaints ? Any help appreciated on this.
> 
> Ok setting core.commentChar to auto does the trick, it would be nice to add a note in the contribution guide for new comers to the project.

Yes, the commit format causes some problems. I'm thinking about moving over to conventional commits as format instead as it is a more common standard.

To solve similar issues I usually create a new branch and manually merge everything into it. I'm glad that you found a solution :slightly_smiling_face:

capricorn86:
I have updated the [Contribute Guideline](https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md) now.

malko:
Made requested changed and moved it to PR #974 as it was simpler to create a new one to change from major to minor.

malko:
I let you close this one when it's ok for you.

minor:
Just for the future, would you be able to change tags to maybe #minor or #major instead? Tagging @minor tags me into each thread and sends me email notifications for issues I'm not a part of (as I'm sure it does for [at]major as well). I'd appreciate it if you can—if not, it's probably not too big a deal, I can figure out a way to sort emails. Thanks!

malko:
> Yes, the commit format causes some problems. I'm thinking about moving over to conventional commits as format instead as it is a more common standard.
I think this would be a nice improvement and will also solve problem reported by github user minor https://github.com/capricorn86/happy-dom/pull/946#issuecomment-1629436938

capricorn86:
> 

@minor I had no idea that it was considered tagging, but I guess it makes sense. Sorry for that.

I have created a task for moving over to conventional commits (#975).

PR #947, opened 2023-06-07
#948@patch: Use macro task timing for postmessage api.

### LINKED ISSUE #948: postMessage api does not use macroTask timing.
**Describe the bug**
The `window.postMessage` api uses "macro task" timing which some tools/libraries rely on to provide a shim of the `setImmediate` api in node js.

Currently the `postMessage` api implemented by HappyDOM is sync which can cause unexpected behavior.

**To Reproduce**
Call the `postMessage` api and the `message` event is fired synchronously.

**Expected behavior**
Call the `postMessage` api and the `message` event is fired asynchronously.

DylanPiercey:
The `window.postMessage` api uses "macro task" timing which some tools/libraries rely on to provide a shim of the `setImmediate` api in node js.

Currently the `postMessage` api implemented by HappyDOM is sync which can cause unexpected behavior.

This PR updates the `postMessage` api to use the global `setImmediate` if available (and falls back to `setTimeout`).

Resolves #948


capricorn86 (approved):

capricorn86 (approved):

capricorn86:
@DylanPiercey I have switched to `setTimeout()` instead of `setImmediate()` as a ran into some problems with `setImmediate()` not being part of the Vitest environment. It feels safer to go with `setTimeout()` for now as it is part of the Javascript standard and `postMessage` probably doesn't have to be high priority and go before other tasks in the event queue.

I also fixed a bug in HTMLIframeElement, where the iframe page wasn't loaded after the element has been connected to DOM and `src` was set.

Issue #950, opened 2023-06-22
Calling getRootNode on document should return document but returns null 


Schleuse:
**Describe the bug**
Calling `getRootNode()` on `document` should return the document but returns `null`

**To Reproduce**
Call `getRootNode()` on `document`

**Expected behavior**
It should return `document` but return's `null`

See Note above `getRootNode` in the [Spec](https://dom.spec.whatwg.org/#dom-node-getrootnode)

capricorn86:
Your fix has been released :slightly_smiling_face: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v10.1.1

PR #951, opened 2023-06-22
#950@patch: Calling getRootNode on document should return the document.

Schleuse:
As described in #950 `document.getRootNode` should return `document`.

This PR makes sure that `Document` fills the `_rootNode` property of the `Node`-Class with itself. `_rootNode` get's used in the `getRootNode` implementation of `Node`.

Fixes: #950


capricorn86:
Thank you for your contribution @Schleuse! :star2:

capricorn86 (approved):

PR #952, opened 2023-06-23
#953@minor: Add missing the movement values into MouseEvent.

mertcan:
`movementX` and `movementY` values provide mouse movements basically. I was trying to test a panning method and noticed that this is not working.


capricorn86:
Thank you for your contribution @mertcan! :slightly_smiling_face: 

The build is failing as the commit format is not following the [Contributing Guide](https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md).

It might not be easy to fix it, so you should consider creating a new branch and a new PR.

We are looking into improving this in #975.

Issue #960, opened 2023-07-02
Implement `childNodes` and `children` as getters


edoardocavazza:
**Is your feature request related to a problem? Please describe.**
I am working to a DOM library that is using the `childNodes` prototype descriptor, but in the happy-dom environment it is missing since it is implemented as class field.

**Describe the solution you'd like**
Espose `childNodes` and `children` as getters, transforming read/write class fields to `_childNodes` and `_children`.

**Describe alternatives you've considered**
I don't know.

**Additional context**
I am going to open a PR with an implementation proposal.

capricorn86:
Your fix has been released @edoardocavazza :slightly_smiling_face: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v10.11.0

PR #961, opened 2023-07-02
#960@minor: Implement childNodes and children as getters.

edoardocavazza:
* Renamed `children` and `childNodes` class fields to `_children` and `_childNodes`
* Expose `children` and `childNodes` as getters


capricorn86:
@edoardocavazza I need to look into this as using getters for children and childNodes will decrease performance as the function for the properties will be executed each time we try to access them.

capricorn86:
I guess we can use the"_children" and "_childNodes" internally. In that case the internal code will at least not be affected by the decrease in performance.

rkurbatov:
Any progress on this? I wanted to try replacement of jsdom with happy-dom on our codebase with pretty big suite of tests and it didn't work because of `children` being written. Seems like this PR should help.

capricorn86:
I made some performance fixes, so that the internal logic will not be affected by the properties being getter functions.

@rkurbatov @edoardocavazza the fix has now been released :slightly_smiling_face:

PR #974, opened 2023-07-10
#945@minor Add support for HTMLInputElement set valueAsNumber

malko:
- Enhance HTMLInputElementValueSanitizer to be more compliant with specs at https://html.spec.whatwg.org/multipage/input.html (breaking change)
- Change HTMLInputElement get valueAsNumber to be more compliant with specs at https://html.spec.whatwg.org/multipage/input.html (breaking change)
- Add support for HTMLInputElement set valueAsNumber based on specs at https://html.spec.whatwg.org/multipage/input.html

This is come as a replacement of PR#946


capricorn86 (approved):
Nice! :star2:

Issue #977, opened 2023-07-11
Implement HTMIInputElement setter for valueAsDate 


malko:
**Is your feature request related to a problem? Please describe.**
For now HTMLInputElement doesn't implement a setter for valueAsDate for input type that should support it

**Describe the solution you'd like**
Implement setter for valueAsDate

this is a follow up of work made about valueAsNumber #945 #946 #974

malko:
@capricorn86 you can assign me this one if you want to

malko:
PR #980 proposed

capricorn86:
@malko your fix has been released :slightly_smiling_face: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v10.3.0

PR #980, opened 2023-07-12
#977@minor Add support for HTMLInputElement set valueAsDate

malko:
As announced here's the PR for valueAsDate here's a quick summary of the changes:
* Make get valueAsDate compliant to the spec -> this can break user's code using valueAsDate with input other than "date", "month", "time", "week"
* Add set valueAsDate 
* Add tests for HTMLInputDateUtility, HTMLInputElement valueAsDate getter and setter


capricorn86 (approved):
Nice! :star2:

PR #985, opened 2023-07-13
#976@minor Fix incorrect comma split on some css properties

### LINKED ISSUE #976: Does not work with urls encoded in base64 or svg
**Describe the bug**
happy-dom does not work with urls encoded in base64 or svg((

**To Reproduce**
1. Create `test.js` file
2. Copy this code 
```js 
const window = new Window();
const _tmp = document.createElement('div');
_tmp.style.setProperty('--test_1', "url(data:image/svg+xml;base64,TEST)");
_tmp.style.backgroundImage = "url(data:image/svg+xml;base64,TEST)"
console.log(_tmp.outerHTML);
```
4. Run `node test.js`
5. See console `<div style="--test_1: url(data;"></div>`

**Expected behavior**
`<div style="--test_1: url(data:image/svg+xml;base64,TEST); background-image: url(data:image/svg+xml;base64,TEST)"></div>`

**Additional context**
The problem [here](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/css/declaration/css-parser/CSSStyleDeclarationCSSParser.ts#L15-L19) is because of the primitive cssText parsing.

malko:
This will fix #976 and probably other bugs related to css values split on comma. 
The fix will avoid splitting in comma if they are inside parentheses, which should make things a little bit better but not perfect.


packages/happy-dom/src/css/declaration/css-parser/CSSStyleDeclarationCSSParser.ts
```diff
@@ -12,24 +12,19 @@ export default class CSSStyleDeclarationCSSParser {
...
+		const rules = [
+			...cssText.matchAll(
+				// PropName   => \s*([^:;]+?)\s*:
+				// PropValue  => \s*((?:[^(;]*?(?:\([^)]*\))?)*?) <- will match any non ';' char except inside (), nested parentheses are not supported
+				// !important => \s*(!important)?
+				// EndOfRule  => \s*(?:$|;)
+				/\s*([^:;]+?)\s*:\s*((?:[^(;]*?(?:\([^)]*\))?)*?)\s*(!important)?\s*(?:$|;)/g
+			)
+		];
+		rules.forEach(([, key, value, important]) => {
```
capricorn86, line 24: For loops are much faster than `forEach()`.

https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/

packages/happy-dom/src/css/declaration/css-parser/CSSStyleDeclarationCSSParser.ts
```diff
@@ -12,24 +12,19 @@ export default class CSSStyleDeclarationCSSParser {
...
-						const important = trimmedValue.endsWith(' !important');
-						const valueWithoutImportant = trimmedValue.replace(' !important', '');
-
-						if (valueWithoutImportant) {
-							callback(trimmedName, valueWithoutImportant, important);
-						}
-					}
-				}
+		const rules = [
+			...cssText.matchAll(
```
capricorn86, line 16: I believe that `Array.from()` is faster than "...".

packages/happy-dom/src/css/declaration/css-parser/CSSStyleDeclarationCSSParser.ts
```diff
@@ -12,24 +12,19 @@ export default class CSSStyleDeclarationCSSParser {
...
+				// PropName   => \s*([^:;]+?)\s*:
+				// PropValue  => \s*((?:[^(;]*?(?:\([^)]*\))?)*?) <- will match any non ';' char except inside (), nested parentheses are not supported
+				// !important => \s*(!important)?
+				// EndOfRule  => \s*(?:$|;)
+				/\s*([^:;]+?)\s*:\s*((?:[^(;]*?(?:\([^)]*\))?)*?)\s*(!important)?\s*(?:$|;)/g
```
capricorn86, line 21: We could store this Regexp as a global constant in the top of the file.

capricorn86 (changes requested):

Found some things that I think we can optimize.

packages/happy-dom/src/css/declaration/css-parser/CSSStyleDeclarationCSSParser.ts
malko: can't find benchmaks that prove that, only some references that in some engines spread operator can have limits regarding the maximum number of element to spread. I will change that.

packages/happy-dom/src/css/declaration/css-parser/CSSStyleDeclarationCSSParser.ts
malko: done

packages/happy-dom/src/css/declaration/css-parser/CSSStyleDeclarationCSSParser.ts
malko: done

capricorn86 (approved):
Nice! :star2:

PR #987, opened 2023-07-13
#530@minor: Broader support for HTMLLabelElement `control`.

### LINKED ISSUE #530: span-wrapped checkboxes are not selectable (ant design)
So we currently use antd which produces the following DOM. Notice the DOM structure `label > span > input`.  It appears that this particular hierarchy prevents the checkbox from being selectable in `happy-dom`.  However, it works as is in `jsdom`.  But If we drop the span, it then works with `happy-dom`:

```jsx
// Does not work
      <label>
        <span className="ant-checkbox">
          <input type="checkbox" value="" defaultValue="" />
        </span>
        Default group
      </label>

// works
      <label>
        <input type="checkbox" value="" defaultValue="" />
        Default group
      </label>
```

Repo example: <https://gitlab.com/robotandkid/the-sandbox/-/tree/feat/upgrade-happy#9174d1c0208cb45596ee7569e67ff371667941e9>

environment:

* happy-dom: 6.0.2
* "@testing-library/dom": "^8.11.3",
* "@testing-library/jest-dom": "^5.14.1",
*  "@testing-library/react": "^12.1.0",
* "@testing-library/user-event": "^14.0.0-beta.12",

Notes:

* again, we're using antd, so we most likely do not have control over the generated DOM. 

Output:

```
    TestingLibraryElementError: Unable to find an accessible element with the role "checkbox" and name `/Default/i`

    Here are the accessible roles:

      textbox:

      Name "name":
      <input
        aria-label="name"
        defaultvalue="hello"
        id="name"
        style="max-width: 280px;"
        type="text"
      />

      --------------------------------------------------
      checkbox:

      Name "checkbox":
      <input
        defaultvalue="checked"
        id="checkbox"
        name="checkbox"
        type="checkbox"
        value="checked"
      />

      Name "":
      <input
        defaultvalue=""
        type="checkbox"
        value=""
      />

      --------------------------------------------------
```

joshkel:
This supports nested / descendent input elements for HTMLLabelElement's `control` property and also supports other labelable elements, as described in the spec:

https://html.spec.whatwg.org/multipage/forms.html#category-label

To be fully spec-compliant, the `control` getter would ignore `for` element(s) that aren't labelable elements, and labelable elements would also include form-associated custom elements.

I'm using a CSS selector for this, because that seemed like the most convenient approach, but if another approach would be better (e.g., for performance?), please let me know.

Fixes #530.

### DIFF
--- a/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElement.ts
+++ b/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElement.ts
@@ -42,12 +42,9 @@ export default class HTMLLabelElement extends HTMLElement implements IHTMLLabelE
 		if (htmlFor) {
 			return <IHTMLElement>this.ownerDocument.getElementById(htmlFor);
 		}
-		for (const child of this.children) {
-			if (child.tagName === 'INPUT') {
-				return <IHTMLElement>child;
-			}
-		}
-		return null;
+		return <IHTMLElement>(
+			this.querySelector('button,input:not([type="hidden"]),meter,output,progress,select,textarea')
+		);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Thank you for your contribution @joshkel! :star2:

Issue #1008, opened 2023-08-04
TypeError: Cannot read properties of null (reading 'name') on Button element


tinaboyce:
**Describe the bug**
Receiving the following error while trying to run Vitest with any happy-dom version after 10.5.2
```
TypeError: Cannot read properties of null (reading 'name')
 ❯ HTMLButtonElementNamedNodeMap.removeNamedItem node_modules/happy-dom/src/nodes/html-button-element/HTMLButtonElementNamedNodeMap.ts:45:17
 ❯ HTMLButtonElement.removeAttribute node_modules/happy-dom/src/nodes/element/Element.ts:650:19
 ❯ setValueForProperty node_modules/react-dom/cjs/react-dom.development.js:1326:10
 ❯ setInitialDOMProperties node_modules/react-dom/cjs/react-dom.development.js:5831:7
 ❯ setInitialProperties node_modules/react-dom/cjs/react-dom.development.js:6026:3
 ❯ finalizeInitialChildren node_modules/react-dom/cjs/react-dom.development.js:7499:3
 ❯ completeWork node_modules/react-dom/cjs/react-dom.development.js:18978:17
 ❯ completeUnitOfWork node_modules/react-dom/cjs/react-dom.development.js:22189:16
 ❯ performUnitOfWork node_modules/react-dom/cjs/react-dom.development.js:22165:12
 ❯ workLoopSync node_modules/react-dom/cjs/react-dom.development.js:22130:22
```

**To Reproduce**
Update happy-dom from 10.5.2 to any version after 10.5.2 and run a test with a button element. Test is using testing-library and Vitest.

**Expected behavior**
Test to not fail like when using happy-dom 10.5.2

**Screenshots**
![Screenshot 2023-08-04 at 9 48 52 am](https://github.com/capricorn86/happy-dom/assets/12929562/4f62475d-2caf-4ecc-8ba1-1e119eba94e7)
![Screenshot 2023-08-04 at 9 44 07 am](https://github.com/capricorn86/happy-dom/assets/12929562/a2bbdd8b-0dce-4956-adf9-5bfb086faa82)

**Device:**
 - OS: Mac OSX Ventura 13.4

capricorn86:
Thank you for reporting @tinaboyce! :slightly_smiling_face: 

This has been fixed now.

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v10.8.0

tinaboyce:
[...]

PR #1009, opened 2023-08-04
#1008@patch: Fixed an error in Button Element when removeItem is null.

massy7:
#1008


capricorn86:
Thank you @massy7 for your contribution! This has been fixed in https://github.com/capricorn86/happy-dom/pull/1011.

Issue #1016, opened 2023-08-09
`Content-Type` header is not set while constructing Request object


orionmiz:
**Describe the bug**
`Content-Type` header is not set while constructing Request object.

**To Reproduce**
```js
const request = new Request(TEST_URL, { method: "POST", body: "Hello World" })
console.log(request.headers.get("Content-Type")) // null
```

**Expected behavior**
Above `request.headers.get("Content-Type")` should return `"text/plain;charset=UTF-8"`

**Additional context**
I'll fix it.

capricorn86:
Your fix has been released and is available here:
https://github.com/capricorn86/happy-dom/releases/tag/v10.10.0

PR #1017, opened 2023-08-09
#1016@minor: Add Content-Type header for Request.

orionmiz:
- Add Content-Type header while constructing Request
- Add tests


capricorn86 (approved):

Issue #1026, opened 2023-08-25
dateIsoWeek depends on the timezone of the machine running the tests


joeframbach:
**Describe the bug**
At https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-input-element/HTMLInputElementDateUtility.ts#L17-L19, `date.getDay()` is dependent on the timezone of the machine. The results are inconsistent.

**To Reproduce**

```
$ TZ="UTC" npm run test
happy-dom:test:  Test Files  89 passed (89)
happy-dom:test:       Tests  2344 passed (2344)

$ TZ="US/Pacific" npm run test

happy-dom:test:  FAIL  test/nodes/html-input-element/HTMLInputDateUtility.test.ts > HTMLInputElementDateUtility > dateToIsoWeek() > Returns the ISO week number
happy-dom:test: AssertionError: expected '2021-W02' to be '2021-W01' // Object.is equality

happy-dom:test:  FAIL  test/nodes/html-input-element/HTMLInputDateUtility.test.ts > HTMLInputElementDateUtility > IsoWeekToDate() > Returns the ISO week number
happy-dom:test: AssertionError: expected 2020-12-29T00:00:00.000Z to deeply equal 2020-12-28T00:00:00.000Z

happy-dom:test:  FAIL  test/nodes/html-input-element/HTMLInputElement.test.ts > HTMLInputElement > get valueAsDate() > Should return valid date for type 'week' with valid value
happy-dom:test: AssertionError: expected 2023-05-30T00:00:00.000Z to deeply equal 2023-05-29T00:00:00.000Z
```

To reproduce the root cause:
```
$ TZ="US/Pacific" node -e "console.log(new Date('2021-01-03').getDay())"
6

$ TZ="UTC" node -e "console.log(new Date('2021-01-03').getDay())"
0
```

**Expected behavior**
Tests should pass anywhere in the world.

**Screenshots**

**Device:**

Latest version

**Additional context**

capricorn86:
It should be possible to set timezone for the entire NodeJS process by just adding the environment variable `TZ=utc`.

Example 1:
```json
     {
           "scripts": {
                 "test": "TZ=utc jest"
            }
     }
```

Example 2:
```javascript
process.env.TZ = 'Europe/Amsterdam' 
```

PR #1027, opened 2023-08-25
#1026@patch: ISO formatting of date inputs are no longer system timez…

joeframbach:
…one dependent.

### DIFF
--- a/packages/happy-dom/src/nodes/html-input-element/HTMLInputElementDateUtility.ts
+++ b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElementDateUtility.ts
@@ -14,8 +14,8 @@ export default class HTMLInputElementDateUtility {
 		date.setUTCDate(date.getUTCDate() - day + 3);
 		const firstThursday = date.getTime();
 		date.setUTCMonth(0, 1);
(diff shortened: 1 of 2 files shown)


joeframbach:
Doesn't look like github linked the issue although I named it in the title. -> #1026

capricorn86:
Thank you for reporting @joeframbach! :slightly_smiling_face: 

I'm a bit unsure about changing the time to UTC. This might not be expected when testing or performing server side rendering.

capricorn86:
It should be possible to set timezone for the entire NodeJS process by just adding the environment variable `TZ=utc`.

Example 1:
```json
     {
           "scripts": {
                 "test": "TZ=utc jest"
            }
     }
```

Example 2:
```javascript
process.env.TZ = 'Europe/Amsterdam' 
```

capricorn86:
Hi again @joeframbach! :slightly_smiling_face: 

I am sorry, but I think I misunderstood your fix. After looking into it more, I have realized that it should be fixed.

It has been fixed by another user in this PR:
https://github.com/capricorn86/happy-dom/pull/1145

As it has been fixed now, I will close this one.

Thank you again!

Issue #1025, opened 2023-08-24
"eval" in a Window is always scoped to the Window


joeframbach:
**Describe the bug**
`eval` has two modes -- a "direct" eval which has access to local scope, and an "indirect" eval which has access only to global scope. The `eval` which executes in a HappyDom document is indirect-only. Behavior in Happy-Dom is different from a real browser.

**To Reproduce**

Execute the script in node:

```
var HappyDom = require('happy-dom');
new HappyDom.Window().document.write(`<!doctype html>
  <script type="text/javascript">
    var globalVariable = 'globalVariableValue';
    (function () {
      var localVariable = 'localVariableValue';
      console.log(eval('globalVariable'));
      console.log(eval('localVariable'));
    })();
  </script>
`);
```

See output

```
globalVariableValue
evalmachine.<anonymous>:1
localVariable
^

ReferenceError: localVariable is not defined
```

Then execute this script in a browser:

```
    var globalVariable = 'globalVariableValue';
    (function () {
      var localVariable = 'localVariableValue';
      console.log(eval('globalVariable'));
      console.log(eval('localVariable'));
    })();
```

See output

```
globalVariableValue
localVariableValue
```

**Expected behavior**

Expected to see "globalVariableValue" and "localVariableValue" in both node and browser.

**Screenshots**
N/A

**Device:**

happy-dom 10.11.0

**Additional context**

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#direct_and_indirect_eval

joeframbach:
I found that the `eval` is a "direct" eval, but the issue is that it is always scoped from the Window, not from the scope where it is called in client code.

https://github.com/capricorn86/happy-dom/blob/1f598077b56905c47e4c81b8599bcecb83d5a82b/packages/happy-dom/src/window/Window.ts#L632

Test case:

```
<!doctype html>
<html>
  <script type="text/javascript">
    (function () {
      console.log(eval('this.location'));
    }).call({});
  </script>
</html>
```

In browser:

```
undefined
```

In happy-dom:

```
Location {
  href: 'https://example.com',
  ...
```

capricorn86:
Thanks again for reporting and contributing @joeframbach! :slightly_smiling_face: 

There is a fix released now.

You can read more about it here:
[1](https://github.com/capricorn86/happy-dom/releases/tag/v11.0.0)

PR #1028, opened 2023-08-25
#1025@major: Replaced window.eval with window.happyDOM.evaluate.

joeframbach:
The existing `window.eval` is overloaded - it handles calls from inside and from outside the VM. This breaks the "direct" eval call:

```
it('Respects direct eval.', () => {
	const result = <string>window.eval(`(function test () {
		window.variable = 'globally defined';
		var variable = 'locally defined';
		return eval('variable');
	})()`);
	expect(result).toBe('locally defined');
});
```

This test returns `expected 'globally defined' to be 'locally defined'`.

In this PR, `window.eval` is left untouched, to only handle calls from inside the VM. For calls originating outside the VM, there is now `window.happyDOM.evaluate`. The naming is better to show that it is not the native `eval`.

This is a breaking change, of course.

### DIFF
--- a/packages/happy-dom/src/nodes/element/Element.ts
+++ b/packages/happy-dom/src/nodes/element/Element.ts
@@ -927,7 +927,7 @@ export default class Element extends Node implements IElement {
 			const attribute = this.getAttribute('on' + event.type);
 
 			if (attribute && !event._immediatePropagationStopped) {
-				this.ownerDocument.defaultView.eval(attribute);
+				this.ownerDocument.defaultView.happyDOM.evaluate(attribute);
 			}
 		}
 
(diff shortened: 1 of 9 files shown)


joeframbach:
Again the issue did not link to the PR. Link it -> #1025

capricorn86:
Thank you for contributing @joeframbach! :star2: 

This PR was quite related to another PR I was working on which was also a potentially breaking change. I adapted your fix, but did not use the `window.happyDOM.evaluate()` logic as I realized that it was not needed anymore (it was before). Instead we are using the built in eval() everywhere now. I have also adapted your unit tests.

I'm sorry that I did not merge in your commits, but I felt that it was too much work. I hope it's ok.

It was fixed in #1044.

PR #1038, opened 2023-08-30
#652@patch: Allow deletion of nonexistent keys from dataset

### LINKED ISSUE #652: delete element.dataset.keyname throws TypeError: 'deleteProperty' on proxy
**Describe the bug**

To remove an attribute from an element's dataset, you should be able to [use the delete operator](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#setting_values): `delete element.dataset.keyname` whether `.keyname` exists or not.

If `.keyname` does not already exist on the dataset, the statement throws with a
`TypeError: 'deleteProperty' on proxy: trap returned falsish for property 'keyname'`.

Native DOM allows to do so and even returns `true`.

**To Reproduce**
inside a Browser with native DOM
```js
const elt = document.body;
delete elt.dataset.fubar;  // true
```

inside a unit test with Happy DOM
```js
const elt = document.body;
delete elt.dataset.fubar;     // throws
delete elt.dataset['fubar'];  // throws too
```
same but keys are initialised first with "anything"
```js
const elt = document.body;
elt.dataset.foo = null;
delete elt.dataset.foo;  // works

elt.dataset.bar = undefined;
delete elt.dataset.bar;  // works too
```

**Expected behavior**
To not throw a TypeError when attempting to `delete` a non-existing dataset key, but silently ignore that the key does not exist.

**Additional context**
 - Node 16.18.0
 - Happy DOM 7.6.6
 - Vite 3.2.3 
 - Vitest vitest/0.24.5

I'm writing a utility to manage multiple values inside data attributes (data-fubar="foo bar bat baz") similar to `classList` . Between tests I wanted to clean `dataset` and ran into this issue.
Not a showstopper 'cos I can just used a different key for each suite.

RussianCow:
Fixes #652. This matches the behavior in the browser.

### DIFF
--- a/packages/happy-dom/src/nodes/element/Dataset.ts
+++ b/packages/happy-dom/src/nodes/element/Dataset.ts
@@ -46,10 +46,10 @@ export default class Dataset {
 				return true;
 			},
 			deleteProperty(dataset: DatasetRecord, key: string): boolean {
-				return (
-					!!element.attributes.removeNamedItem('data-' + Dataset.camelCaseToKebab(key)) &&
-					delete dataset[key]
-				);
+				if (element.attributes.removeNamedItem('data-' + Dataset.camelCaseToKebab(key))) {
+					return delete dataset[key];
+				}
+				return true;
 			},
 			ownKeys(dataset: DatasetRecord): string[] {
 				// According to Mozilla we have to update the dataset object (target) to contain the same keys as what we return:
--- a/packages/happy-dom/test/nodes/html-element/HTMLElement.test.ts
+++ b/packages/happy-dom/test/nodes/html-element/HTMLElement.test.ts
@@ -300,6 +300,8 @@ describe('HTMLElement', () => {
 			expect(element.getAttribute('data-test-delta')).toBe(null);
 			expect(Object.keys(dataset)).toEqual(['testAlpha', 'testBeta', 'testGamma']);
 			expect(Object.values(dataset)).toEqual(['value2', 'value4', 'value5']);
+
+			delete dataset.nonExistentKey;
 		});
 
 		// https://github.com/capricorn86/happy-dom/issues/493


packages/happy-dom/src/nodes/element/Dataset.ts
```diff
@@ -46,10 +46,10 @@ export default class Dataset {
 				return true;
 			},
 			deleteProperty(dataset: DatasetRecord, key: string): boolean {
-				return (
-					!!element.attributes.removeNamedItem('data-' + Dataset.camelCaseToKebab(key)) &&
-					delete dataset[key]
-				);
+				if (element.attributes.removeNamedItem('data-' + Dataset.camelCaseToKebab(key))) {
```
btea, line 49: `removeNamedItem` might throw an [NotFoundError](https://developer.mozilla.org/en-US/docs/Web/API/NamedNodeMap/removeNamedItem#exceptions), maybe it should be wrapped with a `try catch`.

RussianCow: @btea Just to be clear, you're talking about making this change for future compatibility with the spec, correct? Because currently, the Happy DOM implementation of `removeNamedItem` does not appear to ever throw that error. Please correct me if I'm wrong!

btea: Yes, as you said, the current implementation of Happy DOM will return `null` if the attribute does not exist.

So your current modification should be fine, maybe you can create a PR to modify the implementation compatibility specification of `removeNamedItem` in the future.

According to the standard, if the attribute does not exist, an error will be thrown directly. There should be no need to wrap it with `try catch` here. I thought wrong before.

RussianCow: I would be happy to make that change in this PR if you'd like! I can make `removeNamedItem` throw an error if the item does not exist, and then make `deleteProperty` catch the error. Does that sound correct?

btea: I think this is correct, but `deleteProperty` should not need to catch errors, just throw them globally.

RussianCow: My understanding is that deleting an item from a dataset should never throw an error, even if an item with that key does not exist.

```javascript
const el = document.createElement("div")
delete el.dataset.blah // returns true; no error thrown
```

So I think you were right originally and we should catch the error thrown by `removeNamedItem` in the `deleteProperty` method of the dataset proxy.

btea: Sorry, I confused `removeNamedItem` and `deleteProperty`, you are right. 👍

RussianCow: I'll work on this tomorrow!

RussianCow: @btea I just pushed those changes as discussed. Do you think this should still be considered a patch change? It technically might break someone's code if they were relying on `removeNamedItem` not ever throwing, but that behavior was also technically incorrect, so I'm not sure whether to consider it a breaking change. My own feeling is it should still be a patch, but I wanted to check with you.

btea: I think this should still be a patch change.

capricorn86:
Thank you for contribution @RussianCow! :star: 

I would like to look into a solution where we don't have a use try/catch at that many places as it may have a significant performance impact. Hopefully I have some time over tomorrow to look into this.

RussianCow:
Thanks @capricorn86! That seems reasonable. We could have a separate `_removeNamedItemWithoutThrowing` method that simply returns a bool indicating whether the item existed, and have `Dataset` and `Element` use that directly instead? Then `removeNamedItem` can just check the bool and throw if it's false. What do you think? I can throw this together later today.

capricorn86:
> Thanks @capricorn86! That seems reasonable. We could have a separate `_removeNamedItemWithoutThrowing` method that simply returns a bool indicating whether the item existed, and have `Dataset` and `Element` use that directly instead? Then `removeNamedItem` can just check the bool and throw if it's false. What do you think? I can throw this together later today.
> 
> Edit: Actually, it looks like `_removeNamedItemWithoutConsequences` already exists. We could just call that directly from the other classes?
> 
> Edit 2: Ahh, I see that `_removeNamedItemWithoutConsequences` is not exposed in `INamedNodeMap`, so that likely won't work. I'm curious to see how you'd like to handle this. 😃

`_removeNamedItemWithoutConsequences()` is not exactly the same, as it exists to be able to remove an item without consequences such as removing the "style" attribute should empty the "style" property CSS etc. We still want those consequences to happen. I guess we could have a new method called `_removeNamedItem()` that just returns a boolean. Then all sub classes needs to override that method instead of `removeNamedItem()`.

When we want to call it, we will need to cast `INamedNodeMap` to `NamedNodeMap`. E.g. `(<HTMLElementNamedNodeMap>this.attributes)._removeNamedItem`.

It is not pretty, but I guess that would work.

RussianCow:
@capricorn86 Sorry for the delay, but I've made the changes discussed above. Let me know what you think!

capricorn86 (approved):
Nice! :star2:

Issue #1040, opened 2023-09-06
Parsing SVGs adds additional tags around the parsed data. 


drtimwright:
**Describe the bug**
Parsing SVGs using:
``` parser.parseFromString(svg, "image/svg+xml"); ```
adds HTML and BODY tags around the SVG. 

Wheras doing this using JSDOM or in the browser console does not add the additional tags. This means happy-dom is not a drop-replacement for jsdom and does not emulate browser behaviour.

**To Reproduce**

See git repo here with jsdom and happydom examples:
https://github.com/drtimwright/happy-dom-tests/tree/main/src

I'd expect this command to return "svg" not "html":

```new DOMParser().parseFromString("<svg><g></g></svg>", "image/svg+xml").getRootNode().childNodes[0].nodeName;```

**Expected behavior**
* parsing XML that is not HTML should not add HTML tags.

**Device:**
 - OS: Windows (but probably all)
 - Browser N/A
 - Version: 10.11.2

**Additional context**
I'll try to make a PR for this, but no promises!

drtimwright:
![image](https://github.com/capricorn86/happy-dom/assets/7540034/a54c9cd0-8be0-4049-a539-8233dbeb4fbe)

capricorn86:
@drtimwright your fix has been released :slightly_smiling_face: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v11.1.1

drtimwright:
Awesome thanks!

PR #1041, opened 2023-09-06
#1040@minor: Don't add HTML tags when parsing SVGs.

drtimwright:
Match SVG parsing with browser and jsdom.

This PR only does the match for SVGs and leaves all other parsing the same. However, potentially the "default" case should be to not add html tags.

For issue: https://github.com/capricorn86/happy-dom/issues/1040


capricorn86 (approved):
Thank you for your contribution @drtimwright! :star:

Issue #1050, opened 2023-09-11
Support `crypto`


Ayc0:
In the DOM, there is `window.crypto`:

<img width="273" alt="image" src="https://github.com/capricorn86/happy-dom/assets/22725671/ff68f177-dc51-4620-b716-63b45ddfdd99">

Should it be built within HappyDOM?
Note: it can be built using the native node's cypto's webcrypto:

```js
import { webcrypto } from 'node:crypto';
```

capricorn86:
Yes it should be part of Window, but it has not been added yet. Using 'node:crypto' sounds like a good idea.

Ayc0:
I opened https://github.com/capricorn86/happy-dom/pull/1061 to fix this issue

PR #1061, opened 2023-09-13
Add window.crypto

Ayc0:
Fixes https://github.com/capricorn86/happy-dom/issues/1050

- Update `@types/node`
  - The v15 was listed but the v16 is used in the CI (by the way, the v16 is OEL, see [Node versions support](https://github.com/nodejs/release#release-schedule), so it should be updated to v18 or to the v20, and same for the types)
  - Node v15 supports webcrypto but the types for Node v15 don't include it
  - Update import for `ConsoleConstructor` (not available anymore on the global `NodeJS` object, but has to be imported from `node:console`)
- Define as `window.crypto = require('crypto').webcrypto`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
 			],
 			"devDependencies": {
 				"@types/he": "^1.1.2",
-				"@types/node": "^15.6.0",
+				"@types/node": "^16.11.7",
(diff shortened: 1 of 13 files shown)


capricorn86 (approved):
Thank you for your contribution @Ayc0! :star:

Issue #1064, opened 2023-09-13
HTMLDialogElement.close() emits a close Event even when the dialog is already closed


Schleuse:
**Describe the bug**
`HTMLDialogElement.close()` emits an close Event even when the dialog is already closed.
This behaviour differs from the implementation in Chrome or Firefox where calls to the `close` method on already closed dialogs don't trigger 'close' Events.

The [Spec](https://html.spec.whatwg.org/multipage/interactive-elements.html#dom-dialog-close) says:

> The close(returnValue) method steps are:
> 
> If returnValue is not given, then set it to null.
> 
> [Close the dialog](https://html.spec.whatwg.org/multipage/interactive-elements.html#close-the-dialog) [this](https://webidl.spec.whatwg.org/#this) with returnValue.
> 
> When a [dialog](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element) element subject is to be closed, with null or a string result, run these steps:
> 
> If subject does not have an [open](https://html.spec.whatwg.org/multipage/interactive-elements.html#attr-dialog-open) attribute, then return.
> 
> Remove subject's [open](https://html.spec.whatwg.org/multipage/interactive-elements.html#attr-dialog-open) attribute.
> 
> If the [is modal](https://html.spec.whatwg.org/multipage/interactive-elements.html#is-modal) flag of subject is true, then [request an element to be removed from the top layer](https://drafts.csswg.org/css-position-4/#request-an-element-to-be-removed-from-the-top-layer) given subject.
> 
> Let wasModal be the value of subject's [is modal](https://html.spec.whatwg.org/multipage/interactive-elements.html#is-modal) flag.
> 
> Set the [is modal](https://html.spec.whatwg.org/multipage/interactive-elements.html#is-modal) flag of subject to false.
> 
> If result is not null, then set the [returnValue](https://html.spec.whatwg.org/multipage/interactive-elements.html#dom-dialog-returnvalue) attribute to result.
> 
> If subject's [previously focused element](https://html.spec.whatwg.org/multipage/interactive-elements.html#previously-focused-element) is not null, then:
> 
> Let element be subject's [previously focused element](https://html.spec.whatwg.org/multipage/interactive-elements.html#previously-focused-element).
> 
> Set subject's [previously focused element](https://html.spec.whatwg.org/multipage/interactive-elements.html#previously-focused-element) to null.
> 
[...]

PR #1065, opened 2023-09-13
#1064@patch: Emit close event only when dialog was open.

Schleuse:
See issue: https://github.com/capricorn86/happy-dom/issues/1064


capricorn86 (approved):
Thank you for your contribution @Schleuse! :star:

Issue #1067, opened 2023-09-14
Fix problems with the userAgent implementation


capricorn86:
There are a bunch of problems with the user agent that we need to resolve.

**1. Be able to detect Happy DOM as a user agent**
Add "HappyDOM/v{version}" to the default user agent, where {version} is the current version package.json.

**2. Make it possible to pass user agent as an option**
Make it possible to pass "userAgent" as an option to the Window constructor.

**3. Resolve issues with React console warnings**
This is related to https://github.com/facebook/react/issues/24283

We want to disable the "Download the React DevTools" log for a better development experience" log entries somehow in Jest and Vitest.

We should consider changing the user agent to be more similar to JSDOM ("Mozilla/5.0 (linux) AppleWebKit/537.36 (KHTML, like Gecko).

Another option is to default to a different user agent in Vitest and Jest.

PR #1068, opened 2023-09-15
#1067@major: Changes the default user agent.

capricorn86:


mash-graz:
The use of `process.platform`and `process.arch` in the `UserAgent` string construction will crash the Window() initialization when used in `vite` environments, because these entries are not available.

There is already a fix for this issue included in PR #1158

Issue #1069, opened 2023-09-15
Upgrading from 11.0.2 -> 12.0.0: ReferenceError: __dirname is not defined


dflock:
**Describe the bug**
My vitest tests run fine using happy-dom 11.0.2. After upgrading to 12.0.0, I get this for every test suite:

```
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
ReferenceError: __dirname is not defined
 ❯ Function.getHappyDOMVersion node_modules/.pnpm/happy-dom@12.0.0/node_modules/happy-dom/src/navigator/NavigatorUtility.ts:14:45
 ❯ new Window node_modules/.pnpm/happy-dom@12.0.0/node_modules/happy-dom/src/window/Window.ts:215:132
 ❯ new GlobalWindow node_modules/.pnpm/happy-dom@12.0.0/node_modules/happy-dom/lib/window/GlobalWindow.js:11:9
 ❯ Object.setup node_modules/.pnpm/vitest@0.34.4_@vitest+ui@0.34.4_happy-dom@12.0.0_lightningcss@1.21.7/node_modules/vitest/dist/vendor-environments.c9c96bf7.js:559:17
[...]

PR #1070, opened 2023-09-15
#1069@patch: Fixes __dirname reference

ekwoka:
Solves #1069 

__dirname isn't defined in the global context in esm, so it needs to be generated from import.meta.url

### DIFF
--- a/packages/happy-dom/src/navigator/NavigatorUtility.ts
+++ b/packages/happy-dom/src/navigator/NavigatorUtility.ts
@@ -1,5 +1,8 @@
 import FS from 'fs';
 import Path from 'path';
+import { fileURLToPath, URL } from 'url';
+
+const __dirname = globalThis.__dirname ?? fileURLToPath(new URL('.', import.meta.url));
 
 /**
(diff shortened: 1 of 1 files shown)


capricorn86 (dismissed):

capricorn86:
I had to do a "git --amend" to fix the commit message, which caused the build to fail as there is a validation for it.

capricorn86:
`import.meta` is not allowed in CommonJS, so I changed the solution to only use `Path.resolve()`.

ekwoka:
That's an even better solution! Solving the root problem, not just a symptom.

capricorn86:
I realized that using `Path.resolve()` wouldn't work.

I ended up doing a new branch with a complicated solution where a "version.js" file is generated for both CJS and ESM with the version statically applied. This should also have better performance.

I have created a new PR for it here:
https://github.com/capricorn86/happy-dom/pull/1071

capricorn86:
Thank you again for contributing even though we ended up with another solution :slightly_smiling_face: 

I'm closing this now.

Issue #1082, opened 2023-09-20
Registered keys in GlobalRegistrator are not bound to global


capricorn86:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Device:**

**Additional context**

PR #1083, opened 2023-09-20
#1082@patch: Fixes issue related to functions not being bound to glob…

capricorn86:
…al when using GlobalRegistrator from @happy-dom/global-registrator.


valentin-nemcev:
@capricorn86 hey, so this breaks Node.TEXT_NODE and all other static props, I think it's just a JS thing. Probably you need `global[key] = typeof window[key] === 'function' && window[key].prototype != null ? window[key].bind(global) : window[key]`

CleanShot 2023-09-30 at 21 42 50@2x

CleanShot 2023-09-30 at 21 37 52@2x

PR #1093, opened 2023-09-24
#1092@patch: Makes PointerEvent implementation more specification compliant.

### LINKED ISSUE #1092: PointerEvent constructor does not comply with spec
Specification: https://w3c.github.io/pointerevents/#dom-pointerevent-constructor

Current implementation: https://github.com/capricorn86/happy-dom/blob/v12.1.6/packages/happy-dom/src/event/events/PointerEvent.ts

tkrotoff:
Specification: https://w3c.github.io/pointerevents/#dom-pointerevent-constructor

Fix #1092

### DIFF
--- a/packages/happy-dom/src/event/Event.ts
+++ b/packages/happy-dom/src/event/Event.ts
@@ -35,12 +35,12 @@ export default class Event {
 	 * @param type Event type.
 	 * @param [eventInit] Event init.
 	 */
-	constructor(type: string, eventInit: IEventInit = {}) {
+	constructor(type: string, eventInit: IEventInit | null = null) {
 		this.type = type;
 
-		this.bubbles = eventInit.bubbles ?? false;
-		this.cancelable = eventInit.cancelable ?? false;
-		this.composed = eventInit.composed ?? false;
+		this.bubbles = eventInit?.bubbles ?? false;
+		this.cancelable = eventInit?.cancelable ?? false;
+		this.composed = eventInit?.composed ?? false;
 	}
 
 	/**
--- a/packages/happy-dom/src/event/Event.ts
+++ b/packages/happy-dom/src/event/Event.ts
@@ -12,9 +12,9 @@ import IDocument from '../nodes/document/IDocument.js';
  * Event.
  */
 export default class Event {
-	public composed = false;
-	public bubbles = false;
-	public cancelable = false;
+	public composed: boolean;
+	public bubbles: boolean;
+	public cancelable: boolean;
 	public defaultPrevented = false;
 	public eventPhase: EventPhaseEnum = EventPhaseEnum.none;
 	public _immediatePropagationStopped = false;
@@ -22,7 +22,7 @@ export default class Event {
 	public _target: IEventTarget = null;
 	public _currentTarget: IEventTarget = null;
 	public timeStamp: number = performance.now();
-	public type: string = null;
+	public type: string;
 	public _isInPassiveEventListener = false;
 	public NONE = EventPhaseEnum.none;
 	public CAPTURING_PHASE = EventPhaseEnum.capturing;
@@ -35,14 +35,12 @@ export default class Event {
 	 * @param type Event type.
 	 * @param [eventInit] Event init.
 	 */
-	constructor(type: string, eventInit: IEventInit = null) {
+	constructor(type: string, eventInit: IEventInit = {}) {
 		this.type = type;
 
-		if (eventInit) {
-			this.bubbles = eventInit.bubbles || false;
-			this.cancelable = eventInit.cancelable || false;
-			this.composed = eventInit.composed || false;
-		}
+		this.bubbles = eventInit.bubbles ?? false;
+		this.cancelable = eventInit.cancelable ?? false;
+		this.composed = eventInit.composed ?? false;
 	}
 
 	/**
--- a/packages/happy-dom/src/event/events/CustomEvent.ts
+++ b/packages/happy-dom/src/event/events/CustomEvent.ts
@@ -5,7 +5,7 @@ import ICustomEventInit from './ICustomEventInit.js';
  *
  */
 export default class CustomEvent extends Event {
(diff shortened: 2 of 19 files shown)


packages/happy-dom/src/event/events/PointerEvent.ts
```diff
@@ -6,37 +6,46 @@ import IPointerEventInit from './IPointerEventInit.js';
...
+	public readonly predictedEvents: PointerEvent[] = [];
 
 	/**
 	 * Constructor.
 	 *
 	 * @param type Event type.
 	 * @param [eventInit] Event init.
 	 */
-	constructor(type: string, eventInit: IPointerEventInit = null) {
+	constructor(type: string, eventInit: IPointerEventInit = {}) {
```
tkrotoff, line 29: https://w3c.github.io/pointerevents/#dom-pointerevent-constructor

> constructor([DOMString](https://webidl.spec.whatwg.org/#idl-DOMString) type, optional [PointerEventInit](https://w3c.github.io/pointerevents/#dom-pointereventinit) eventInitDict = **{}**);

packages/happy-dom/src/event/events/CustomEvent.ts
```diff
@@ -5,20 +5,18 @@ import ICustomEventInit from './ICustomEventInit.js';
...
 	 * @param [eventInit] Event init.
 	 */
-	constructor(type: string, eventInit: ICustomEventInit = null) {
+	constructor(type: string, eventInit: ICustomEventInit = {}) {
 		super(type, eventInit);
 
-		if (eventInit) {
-			this.detail = eventInit.detail !== undefined ? eventInit.detail : null;
-		}
+		this.detail = eventInit.detail ?? null;
```
capricorn86, line 19: `eventInit.detail` can be a falsy value (e.g. 0 or '')

packages/happy-dom/src/event/events/PointerEvent.ts
capricorn86: The downside is that we will get multiple one liner if statements as `eventInit` can be null (see other comment) (e.g. `eventInit?.pointerId ?? 0`), but the benefit is that we will have all default values in one place of course. I guess it is fine.

packages/happy-dom/src/event/events/CustomEvent.ts
```diff
@@ -5,20 +5,18 @@ import ICustomEventInit from './ICustomEventInit.js';
  *
  */
 export default class CustomEvent extends Event {
-	public detail: object = null;
+	public detail: object;
```
capricorn86, line 8:
```suggestion
	public detail: object | string | number | boolean;
```

capricorn86 (changes requested):

I found some things that we should fix.

packages/happy-dom/src/event/events/PointerEvent.ts
capricorn86: The downside is that we will get multiple one liner if statements as `eventInit` can be null (see other comment) (e.g. `eventInit?.pointerId ?? 0`), but the benefit is that we will have all default values in one place of course, so I think you are right that it is better to go with that.

packages/happy-dom/src/event/events/CustomEvent.ts
tkrotoff: Should this be `any` like i TypeScript lib.dom.d.ts: https://github.com/microsoft/TypeScript/blob/695226ea5ac64195e4ab0180cd38f06010157ec0/src/lib/dom.generated.d.ts#L5891 and in the spec: https://dom.spec.whatwg.org/#interface-customevent

> readonly attribute [any](https://webidl.spec.whatwg.org/#idl-any) [detail](https://dom.spec.whatwg.org/#dom-customevent-detail);

packages/happy-dom/src/event/Event.ts
```diff
@@ -35,14 +35,12 @@ export default class Event {
 	 * @param type Event type.
 	 * @param [eventInit] Event init.
 	 */
-	constructor(type: string, eventInit: IEventInit = null) {
+	constructor(type: string, eventInit: IEventInit = {}) {
```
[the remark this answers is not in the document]
tkrotoff: Wow, good catch!

I can't understand/verify from the spec if `new Event('click', **null**)` is legit (of course it is since Chrome & Firefox allow it). Can you point the line that says that?

https://dom.spec.whatwg.org/#interface-event
> constructor(DOMString type, **optional** EventInit eventInitDict = **{}**)

The `EventInit eventInitDict = **{}**` part is misleading. I guess it's `optional` that means `undefined | null` and not just `undefined`

So I've added a commit that does what you ask:
> eventInit: IEventInit | null = null
> this.height = eventInit?.height ?? 1

Btw I strongly suggest you to enable [`strict` TypeScript mode](https://www.typescriptlang.org/tsconfig#strict) which turns on [`strictNullChecks`](https://www.typescriptlang.org/tsconfig#strictNullChecks).

packages/happy-dom/src/event/events/CustomEvent.ts
capricorn86: Yes, I guess that "any" is most correct. Not sure that the current TS config allows it, but maybe you can try to override it otherwise.

I'm thinking about going through the types in general in Happy DOM and make sure they follow the lib.dom.d.ts.

packages/happy-dom/src/event/Event.ts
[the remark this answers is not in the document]
capricorn86: I agree that `strictNullChecks` should be enabled.

I think that optional values in general in the spec can be null or undefined, but I'm not sure. I just tried in FF and Chrome and I assumed that they are following it correctly.

capricorn86 (dismissed):

capricorn86:
@tkrotoff I changed the type for `CustomEvent.detail` to "any" and added a unit test for `PointerEvent.width` and `PointerEvent.height`.

PR #1098, opened 2023-09-26
#1097@patch: Implements Symbol.toStringTag on EventTarget and AbortSignal.

### LINKED ISSUE #1097: Unable to fetch to third party using Remix with error: TypeError: Expected signal to be an instanceof AbortSignal
**Describe the bug**
While I was trying to implement unit testing for my authenticate loaders and actions on my Remix project, I could not launch my test using Happy-DOM. I tried the same tests using JSDOM, and they passed without error. 

After digging into the problem, I found the problem is that the `AbortSignal` and `EventTarget` classes don't implement the `Symbol.toStringTag` method. [Remix is using this method to detect if the signal attached to a Request is an AbortSignal or EventTarget](https://github.com/remix-run/web-std-io/blob/main/packages/fetch/src/utils/is.js#L95). 

As a result, [this check fails](https://github.com/remix-run/web-std-io/blob/main/packages/fetch/src/utils/is.js#L95) and throws an exception.

**To Reproduce**
Steps to reproduce the behavior (See Additional context for the link to the repository with the minimal code to reproduce it):
1. Create a loader that has a fetch to a third-party API
2. Create a test expecting the result of a loader
3. Launch the test, and it will fail with `TypeError: Expected signal to be an instanceof AbortSignal `

**Expected behavior**
I expected that it would work as on JSDOM.

**Additional context**
I made a [repository where you can find a minimal reproduction of the bug](https://github.com/Afsoon/happy-dom-reproduction) and a possible workaround until a fix lands in the repository.

~~If it has low priority, I can make a PR to fix it.~~ Nevermind, I had patched locally the package, I will open an PR ASAP with the fix.

This is the workaround 

```ts
const Signal = new AbortSignal();

const SignalProxied = new Proxy(Signal, {
  get: (target, prop, receiver) => {
    if (prop === Symbol.toStringTag) {
      return "AbortSignal";
    }

    return Reflect.get(target, prop, receiver);
  },
});

const controller = new AbortController();
const ControllerProxied = new Proxy(controller, {
  get: (target, prop, receiver) => {
    if (prop === "signal") {
      return SignalProxied;
    }

    return Reflect.get(target, prop, receiver);
  },
});

global.AbortController = new Proxy(global.AbortController, {
  construct: () => {
    return ControllerProxied;
  },
});
```

The workaround is based on proxify AbortController and AbortSignal to intercept the call to the method `Symbol.toStringTag`

Afsoon:
Fixes #1097

### DIFF
--- a/packages/happy-dom/src/event/EventTarget.ts
+++ b/packages/happy-dom/src/event/EventTarget.ts
@@ -20,6 +20,13 @@ export default abstract class EventTarget implements IEventTarget {
 	} = {};
 
 	/**
+	 * Return a default description for the EventTarget class.
+	 */
+	public get [Symbol.toStringTag](): string {
+		return 'EventTarget';
+	}
+
+	/**
 	 * Adds an event listener.
 	 *
 	 * @param type Event type.
--- a/packages/happy-dom/src/event/IEventTarget.ts
+++ b/packages/happy-dom/src/event/IEventTarget.ts
@@ -7,6 +7,11 @@ import IEventListenerOptions from './IEventListenerOptions.js';
  */
 export default interface IEventTarget {
 	/**
+	 * Return a default description for the EventTarget class.
+	 */
+	[Symbol.toStringTag]: string;
+
+	/**
 	 * Adds an event listener.
 	 *
 	 * @param type Event type.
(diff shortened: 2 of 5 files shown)


Afsoon:
@capricorn86 any guidance on why is failing the test? In my local isn't throwing an error, even after delete the `.turbo` folder.

capricorn86:
> ~~@capricorn86 any guidance on why is failing the test? In my local isn't throwing an error, even after delete the `.turbo` folder. ~~ So I assume that `UncaughtExceptionObserver.test` it's a flaky test, no?

I believe it has something to do with the turborepo cache in Github. I have tried to fix it, but it seems like it is still flaky.

PR #1103, opened 2023-09-30
#0@trivial: Add version badge.

btea:


capricorn86:
@btea I don't think it is necessary as the version is available under the list of packages and it is also available under "Releases" in Github.

btea:
👌

Issue #1023, opened 2023-08-21
React input element missing checked attribute


ArrayKnight:
**Describe the bug**
When an input of type checkbox or radio is provided the checked prop with a value of true, the resulting dom doesn't include the expected checked attribute.

**Expected behavior**
Elements should have checked attribute, tests should pass

**Screenshots**
<img width="1287" alt="image" src="https://github.com/capricorn86/happy-dom/assets/5431637/f041c7da-0d67-484e-ad9c-909bc3b01124">

**Device:**
 - OS: macOS 13.4.1 (22F82)
 - Browser [e.g. chrome, safari]: Vitest
 - Version [e.g. 22] 0.34.2

**Additional context**
Running the exact same tests using jsdom run and pass as expected

"@testing-library/dom": "^9.3.1",
"@testing-library/jest-dom": "^6.0.1",
"@testing-library/react": "^14.0.0",
"@testing-library/user-event": "^14.4.3",
"happy-dom": "^10.10.4",
"jsdom": "^22.1.0",
"typescript": "^5.1.6",
"vite": "^4.4.9",
"vitest": "0.34.2"

timkindberg:
Same thing happens for me, when clicking the checkbox label it doesn't work.

```js
// App.js

import './App.css';

function App() {
  return (
    <label data-testid="my-checkbox">
      <input name="terms_and_conditions" type="checkbox" value="false"/>
      I agree to the Terms of Use
    </label>
  );
}

export default App;
```

```js
// App.test.js

import { fireEvent, render, screen } from '@testing-library/react';
import App from './App';

test('allows clicking checkbox label', () => {
  // Validate that we are using HappyDOM, not JSDOM
  const isJSDOM = typeof navigator !== "undefined" && navigator.userAgent.includes("jsdom");
  expect(isJSDOM).toBe(false)
  expect(window.happyDOM).toBeDefined();

  render(<App />);
  const label = screen.getByTestId('my-checkbox');
  const checkbox = screen.getByLabelText('I agree to the Terms of Use');
  expect(checkbox).not.toBeChecked();

  // Clicking the input itself should work
  fireEvent.click(checkbox)
  expect(checkbox).toBeChecked();
  fireEvent.click(checkbox)
  expect(checkbox).not.toBeChecked();

  // Clicking Label should also work!
  fireEvent.click(label);
  expect(checkbox).toBeChecked(); // THIS LINE FAILS
});
```

Output:
```
    Received element is not checked:
[...]

PR #1108, opened 2023-10-02
#1023@minor: Adds support for dispatching a click event on the contro…

capricorn86:
…l element when clicking on a label.

### DIFF
--- a/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElement.ts
+++ b/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElement.ts
@@ -2,6 +2,9 @@ import HTMLElement from '../html-element/HTMLElement.js';
 import IHTMLElement from '../html-element/IHTMLElement.js';
 import IHTMLFormElement from '../html-form-element/IHTMLFormElement.js';
 import IHTMLLabelElement from './IHTMLLabelElement.js';
+import Event from '../../event/Event.js';
+import EventPhaseEnum from '../../event/EventPhaseEnum.js';
+import PointerEvent from '../../event/events/PointerEvent.js';
 
 /**
  * HTML Label Element.
@@ -40,7 +43,8 @@ export default class HTMLLabelElement extends HTMLElement implements IHTMLLabelE
 	public get control(): IHTMLElement {
 		const htmlFor = this.htmlFor;
 		if (htmlFor) {
-			return <IHTMLElement>this.ownerDocument.getElementById(htmlFor);
+			const control = <IHTMLElement>this.ownerDocument.getElementById(htmlFor);
+			return control !== this ? control : null;
 		}
 		return <IHTMLElement>(
 			this.querySelector('button,input:not([type="hidden"]),meter,output,progress,select,textarea')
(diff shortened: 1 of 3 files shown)


packages/happy-dom/src/nodes/html-label-element/HTMLLabelElement.ts
```diff
@@ -66,4 +70,23 @@ export default class HTMLLabelElement extends HTMLElement implements IHTMLLabelE
...
+	public override dispatchEvent(event: Event): boolean {
+		const returnValue = super.dispatchEvent(event);
+
+		if (
+			event.type === 'click' &&
+			(event.eventPhase === EventPhaseEnum.atTarget || event.eventPhase === EventPhaseEnum.bubbling)
+		) {
+			const control = this.control;
+			if (control && event.target !== control) {
+				control.dispatchEvent(new PointerEvent('click', { bubbles: true, cancelable: true }));
```
timkindberg, line 86: Is this what happens in native browsers as well? It dispatches a click event on the input as well?

timkindberg: Yup ok, looks like it is the correct behavior! https://codesandbox.io/s/quizzical-cerf-mp2xtk?file=/src/App.js

https://github.com/capricorn86/happy-dom/assets/168046/3ad2c2fc-98db-4f32-b655-5b35f44239f3

capricorn86: Took sometime to figure it out, but it should be correct now :slightly_smiling_face:

capricorn86: The label will have two click event dispatched on it if it is a parent to the input, as the event bubbles up to it.

capricorn86: Yes sure, it supports that as well.

packages/happy-dom/test/nodes/html-label-element/HTMLLabelElement.test.ts
```diff
@@ -76,4 +78,30 @@ describe('HTMLLabelElement', () => {
 			expect(element.form === form).toBe(true);
 		});
 	});
+
+	describe('dispatchEvent()', () => {
+		it('Dispatches a click event on the control element if it exists.', () => {
```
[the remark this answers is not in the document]
artursvonda: Nevermind me, understood what happened here. `span` is nested inside the label. Sorry for silly question.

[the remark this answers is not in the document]
capricorn86: @artursvonda what is it that broke? Can you file a bug for it?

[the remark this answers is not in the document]
artursvonda: Looks like issue might be with `@testing-library/user-event` dispatching click event twice on their own causing checkbox to be toggled twice. Still investigating and once I have repro, will creating issue on both repos.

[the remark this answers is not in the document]
capricorn86: Thank you!

Issue #1114, opened 2023-10-03
Refactor: `AsyncTaskManager.whenComplete()` should not rely on setTimeout


KevinBLT:
**Is your feature request related to a problem? Please describe.**
`AsyncTaskManager.whenComplete()` should not rely on setTimeout, so it doesn't artificially slow down tests or possible SSR solutions when async requests were required.

**Describe the solution you'd like**
Resolve as fast as possible

**Describe alternatives you've considered**
There is none.

**Additional context**
Pull request is coming.

PR #1115, opened 2023-10-03
#1114@minor: Remove `setTimeout` AsyncTaskManager

KevinBLT:
See #1114 information.

I checked that tests kept running and created a new test case for multiple and overlapping calls to whenComplete().

Motivation was the need for SSR with web components that needed fetch responses for a full SSR rendering and the timeout slowed down rendering.


capricorn86:
Thank you for contributing @KevinBLT! :star: 

We want to ensure that `whenAsyncComplete()` is triggered when all async tasks are done. The only way to ensure that is if we wait for new async tasks to come in, even though all current tasks have completed.

Consider this:
```javascript
const window = new Window();

document.write(`
       <script>
             let completedAsyncTasks = 0;
             setTimeout(() => {
                     completedAsyncTasks++;
                     console.log('Completed:', completedAsyncTasks);
                     setTimeout(() => {
                            completedAsyncTasks++;
                            console.log('Completed:', completedAsyncTasks);
                     });
             }, 10);
       </script>
`);

window.happyDOM.whenAsyncComplete().then(() => {
    console.log('All tasks completed');
});
```

The console will be:
```
Completed: 1
All tasks completed
Completed 2
```

What we can do is to change the wait time to either a `setTimeout()` that is as short as possible (0) or something like `process.nextTick`.

KevinBLT:
Hello hello,

yeah I can see that case! Should also add a test for that. I will try nextTick and see what will happen :-)

capricorn86:
Also, if you are concerned about performance when doing SSR, you should consider setting `disableErrorCapturing` to true and use `@happy-dom@uncaught-exception-observer` instead. Using try/catch which is used in Happy DOM for capturing errors slows down execution of the code inside the catch block a lot.

KevinBLT:
It is now working for the case you mentioned above.

Although i think it can always be failing when doing this: 
```html
<script>
setTimeout(async () => { 
  console.log('A');

  await Promise.resolve(); 
  await Promise.resolve();
  await Promise.resolve();
  await Promise.resolve();

  setTimeout(async () => { 
    console.log('B');

    setTimeout(async () => { 
      console.log('C');
    
      setTimeout(async () => console.log('D'));
    });
  });
});
</script>
```

I think it would need `Promise` to be overridden.

In the end, it can still not take async functions into account if no window function was called and properly registered within AsyncTaskManager. The ultimate best would be to ask the node vm instance if there are running async task or else there could always be awkward situations where it does not work.

In the end, for SSR this might be better with two microtasks queues but a clean solution would be to ask nodejs engine directly what is going on in there. We could merge this but I am not fully confident, that this is indeed better or safer than before. The example above would also break the previous solution with the code gathers enough promises to exceed the 10ms of the timeout. What do you think about that topic? Or is SSR with async functions not a plausible use case here?

capricorn86:
> It is now working for the case you mentioned above.
> 
> Although i think it can always be failing when doing this:
> 
> ```
> <script>
> setTimeout(async () => { 
>   console.log('A');
> 
>   await Promise.resolve(); 
>   await Promise.resolve();
>   await Promise.resolve();
>   await Promise.resolve();
> 
>   setTimeout(async () => { 
>     console.log('B');
> 
>     setTimeout(async () => { 
>       console.log('C');
>     
>       setTimeout(async () => console.log('D'));
>     });
>   });
> });
> </script>
> ```
> 
> I think it would need `Promise` to be overridden.
> 
> In the end, it can still not take async functions into account if no window function was called and properly registered within AsyncTaskManager. The ultimate best would be to ask the node vm instance if there are running async task or else there could always be awkward situations where it does not work.
> 
> In the end, for SSR this might be better with two micro tasks queued but a clean solution would be to ask nodejs engine directly what is going on in there. We could merge this but I am not fully confident, that this is indeed better or safer than before. The example above would also break the previous solution with the code gathers enough promises to exceed the 10ms of the timeout. What do you think about that topic? Or is SSR with async functions not a plausible usecase here?

I have looked into it.

The problem is that it doesn't seem to be possible to use `process.nextTick()` or `queueMicrotask()` as they get triggered before `then()` or `catch()` are resolved in Promises. I could only make it work by using `setImmediate()` which is executed after micro tasks.

I can see that you also have similar problems as multiple unit tests are failing:
https://github.com/capricorn86/happy-dom/actions/runs/6410223607/job/17404277071?pr=1115#step:11:63

I have created a PR which will improve performance quite a lot as it no longer waits 10ms and it doesn't have to wait for timeout callbacks at all. Have a look and say what you think.

My pull request:
#1119

KevinBLT:
Looks nice and sensible to me, didn't came to my mind with this `setImmediate()` (never used, I have to say :D).

So yeah! Should be nice! :-) 👍

PR #1117, opened 2023-10-04
#1113@patch: Fixes second parameter missing from MutationObserver callback.

### LINKED ISSUE #1113: MutationObserver callback should have MutationObserver as second argument
**Describe the bug**

MutationObserver callback should have MutationObserver as second argument.

**To Reproduce**

I have a test in my test suite that checks if MutationObserver callback gets called with correct arguments.

```js
    expect(listener).toHaveBeenCalledWith(
      [expect.any(MutationRecord)],
      expect.any(MutationObserver),
    );
```

and results in an error:

```
AssertionError: expected "spy" to be called with arguments: [ [ Any{ …(3) } ], Any{ …(3) } ]

Received: 

  1st spy call:

  Array [
    Array [
     Any<MutationRecord>,
    ],
-   Any<MutationObserver>,
  ]
```

**Expected behavior**

MutationObserver callback to have MutationObserver as second argument.

Source of this claim:

- https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
- TypeScript's lib.dom.d.ts:
    ```ts
    interface MutationCallback {
        (mutations: MutationRecord[], observer: MutationObserver): void;
    }
    ```
- JSDoc behavior

**Screenshots**

Commit with tests failing after replacing JSDom with Happy DOM:
https://github.com/wojtekmaj/react-hooks/commit/cc74f9a6bf1a897c63db9afbe585701b10162b7f

wojtekmaj:
Fixes #1113

### DIFF
--- a/packages/happy-dom/src/mutation-observer/MutationListener.ts
+++ b/packages/happy-dom/src/mutation-observer/MutationListener.ts
@@ -1,4 +1,5 @@
 import IMutationObserverInit from './IMutationObserverInit.js';
+import MutationObserver from './MutationObserver.js';
 import MutationRecord from './MutationRecord.js';
 
 /**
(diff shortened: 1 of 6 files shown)


capricorn86 (approved):

wojtekmaj:
Hmmm, looks like the release workflow failed, but the cause seems unrelated 🤔

PR #1118, opened 2023-10-04
#0@trivial: Use native actions/setup-node npm cache.

wojtekmaj:
This is an experiment to see if CI performance would improve if we were to use native npm cache instead of caching (quite heavy, especially when compared to cache) node_modules.


wojtekmaj:
Looks like it's not worth it

PR #1123, opened 2023-10-07
docs: fix item typo

goring:


capricorn86:
Thank you for contributing @goring! :star: 

It seems like the commit message is not following the [Commit Convention](https://github.com/capricorn86/happy-dom/blob/master/docs/contributing.md#commit-convention) for the project and therefore the build is failing.

capricorn86:
You should be able to do `git commit --amend -m "#1123@patch: Fixes typo in documentation."` to fix the problem.

Issue #1126, opened 2023-10-10
Invalid tree due to optional tags


itozyun:
**Describe the bug**
Hi, I am developing a [library](https://github.com/itozyun/html.json) to convert HTML to JSON with happy-dom.

I report an invalid tree result of the following test that omits the end tags.

**To Reproduce**

Add the following to /packages/happy-dom/test/xml-parser/XMLParser.test.ts.

~~~js
it('Handles unclosed tags (e.g. <dl>, <table>).', () => {
	const root = XMLParser.parse(
		document,
		`
		<div class="test" disabled>
			<dl>
				<dt>Test
				<dd>Test</dt></dt></dt></dt>
				<dt>Test</dt>
				<dd>Test
			</dl>
			<select>
				<option>Test 1
				<optgroup>
					<option>Test 2
					<option>Test 3</option>
					<option>Test 4
				<optgroup>
					<option>Test 5
					<option>Test 6
			</select>
			<ruby>明日
				<rp>(
				<rt>Ashita
				<rp>)
			</ruby>
			<table>
				<caption>Test
				<colgroup>
					<col>
					<col>
					<col>
				<thead>
					<tr>
						<th>Test 1</th><td>Test 2<td>Test 3
					<tr>
						<th>Test 4<td>Test 5<td>Test 6</th></th></th>
				<tbody>
					<tr>
						<th>Test 7<td>Test 8<td>Test 9</td>
					</tr>
					<tr>
						<th></td></td></td></td></td></td>Test 10<td>Test 11<td>Test 12
				<tfoot>
					<tr>
						<th>Test 13<td>Test 14<td>Test 15
			</table>
		</div>
		`
	);

	expect(new XMLSerializer().serializeToString(root).replace(/\s/gm, '')).toBe(
		`
		<div class="test" disabled="">
			<dl>
				<dt>Test</dt>
				<dd>Test</dd>
				<dt>Test</dt>
				<dd>Test</dd>
			</dl>
			<select>
				<option>Test 1</option>
				<optgroup>
					<option>Test 2</option>
					<option>Test 3</option>
					<option>Test 4</option>
				</optgroup>
				<optgroup>
					<option>Test 5</option>
					<option>Test 6</option>
				</optgroup>
			</select>
			<ruby>明日
				<rp>(</rp>
				<rt>Ashita</rt>
				<rp>)</rp>
			</ruby>
			<table>
				<caption>Test</caption>
				<colgroup>
					<col>
					<col>
[...]

PR #1127, opened 2023-10-10
#1126@patch: Add more unclosed tags support to XMLParser.

itozyun:
How does jsdom solve this? I don't know...


capricorn86:
Thank you for your contribution @itozyun! :star2: 

I need to look into the code a bit more before I can merge the PR.

capricorn86:
I believe that most of the logic here could have been solved with the "contentModel" property of `HTMLElementConfig`, so I will close this PR for now.

PR #1145, opened 2023-10-26
#1144@patch: Run tests against newer Node.js versions

### LINKED ISSUE #1144: Run tests against newer Node.js versions
The pull request workflow is currently running tests only against an EOL Node.js version, 16. Tests ideally should be run against the current LTS version (20) as well as the last LTS version (18) to ensure this library continues to properly support them as well.

IGx89:
Also fixed tests that were failing locally for me, for various reasons.

Closes #1144

### DIFF
--- a/packages/happy-dom/src/nodes/html-input-element/HTMLInputElementDateUtility.ts
+++ b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElementDateUtility.ts
@@ -5,24 +5,28 @@ export default class HTMLInputElementDateUtility {
 	/**
 	 * Returns iso week number from given date
 	 *
+	 * @see https://stackoverflow.com/a/6117889
 	 * @param date Date or number.
 	 * @returns Iso-week string.
 	 */
 	public static dateIsoWeek(date: Date | number): string {
-		date = new Date(date);
-		const day = (date.getUTCDay() + 6) % 7;
-		date.setUTCDate(date.getUTCDate() - day + 3);
-		const firstThursday = date.getTime();
-		date.setUTCMonth(0, 1);
-		if (date.getUTCDay() !== 4) {
-			date.setUTCMonth(0, 1 + ((4 - date.getUTCDay() + 7) % 7));
-		}
-		return (
-			date.getUTCFullYear() +
-			'-W' +
-			String(1 + Math.ceil((firstThursday - date.getTime()) / 604800000)).padStart(2, '0')
+		const parsedDate = typeof date === 'number' ? new Date(date) : date;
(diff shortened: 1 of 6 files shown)


packages/uncaught-exception-observer/test/UncaughtExceptionObserver.test.ts
```diff
@@ -61,25 +61,24 @@ async function itObservesUnhandledJavaScriptFetchRejections(): Promise<void> {
...
 
 	if (!(errorEvent instanceof window.ErrorEvent)) {
 		throw new Error('Error event not dispatched.');
 	}
 
-	if (
-		errorEvent.error.message !==
-		'Fetch to "https://localhost:3000/404.js" failed. Error: connect ECONNREFUSED 127.0.0.1:3000'
-	) {
+	if (!errorEvent.error.message.startsWith('Fetch to "https://localhost:3000/404.js" failed.')) {
```
IGx89, line 77: Node.js:
- 16: `Error: connect ECONNREFUSED 127.0.0.1:3000`
- 18: `Error: connect ECONNREFUSED ::1:3000`
- 20: `Error: `
Since the actual message seemed to be irrelevant to this test, I simplified the check to ignore that part.

packages/happy-dom/src/nodes/html-input-element/HTMLInputElementDateUtility.ts
```diff
@@ -14,8 +14,8 @@ export default class HTMLInputElementDateUtility {
 		date.setUTCDate(date.getUTCDate() - day + 3);
 		const firstThursday = date.getTime();
 		date.setUTCMonth(0, 1);
-		if (date.getDay() !== 4) {
```
capricorn86, line 17: It seems like we are using this:
https://stackoverflow.com/a/16591175/12415847

But we should use this (it seems like JSDOM is using this one as well):
https://stackoverflow.com/a/6117889/12415847

capricorn86 (changes requested):

capricorn86 (approved):
Thank you for your contribution @IGx89! :star2: 

I looked into `dateIsoWeek()` and It seems like we are using this:
https://stackoverflow.com/a/16591175/12415847

But we should probably use this (it seems like JSDOM is using this one as well):
https://stackoverflow.com/a/6117889/12415847

I have updated changed it now.

packages/uncaught-exception-observer/test/UncaughtExceptionObserver.test.ts
capricorn86: Yes, that should be fine.

PR #1157, opened 2023-11-06
#1156@patch: Add missing html elements from global scope.

### LINKED ISSUE #1156: Missing global HTML objects
Some global objects, that normally exist in the window object, are missing in missing-dom, for example HTMLOptGroupElement and HTMLOOptionElement.

lukasbach:
Fixes #1156


capricorn86:
Thank you for your contribution @lukasbach! :star2: 

Sorry for taking such long time before answering. This was fixed in #1141 as well, and the fix has been released.

I will close this one.

Issue #1163, opened 2023-11-19
Selection.anchorOffset should return 0 instead of null according to the spec


seanogdev:
**Describe the bug**
Trying to migrate to happy-dom and our test presumes that if the anchorNode is null,  then the anchorOffset is 0
**To Reproduce**

https://w3c.github.io/selection-api/#dom-selection-anchoroffset

**Expected behavior**
It matches spec

**Screenshots**

**Device:**

**Additional context**

capricorn86:
Thank you again for your contribution @seanogdev!

Your fix has been released:
https://github.com/capricorn86/happy-dom/releases/tag/v13.1.2

PR #1164, opened 2023-11-19
#1163@patch: Selection anchorOffset should return a number.

seanogdev:
The offsets should return a number https://w3c.github.io/selection-api/#dom-selection-anchoroffset


capricorn86 (dismissed):

Sorry for taking such a long time before reviewing this.

capricorn86 (approved):

PR #1169, opened 2023-11-24
#1168@patch: Normalize whitespace in processing of DOMTokenList.

### LINKED ISSUE #1168: Element.classList has empty string item when a class attribute contains multiple spaces
**Describe the bug**
Element.classList has empty string item when a class attribute contains multiple spaces. Multiple spaces should be compressed  and Element.classList should not have empty string item.

**To Reproduce**
1. Run following code:

```js
const {Window} = require('happy-dom');
const {document} = new Window();
element.className = 'a  b';
const list = [...element.classList.values()];
console.log(list); // ['a', '', 'b']
```

**Expected behavior**
The expected result of `list` is `['a', 'b']`. Other DOM implenetations including Google Chrome, Mozilla Firefox, and JSDOM return `['a', 'b']`.

**Device:**
 - OS: Windows 11
 - Browser: Google Chrome and Mozilla Fireefox
 - Version; happy-dom 12.10.3

takenspc:
Normalize whitespace before splitting text into tokens.

Fixes #1168.


packages/happy-dom/src/dom-token-list/DOMTokenList.ts
```diff
@@ -212,6 +204,18 @@ export default class DOMTokenList implements IDOMTokenList {
...
 
+	/**
+	 * Returns token list from attribute value.
+	 *
+	 * @see https://infra.spec.whatwg.org/#split-on-ascii-whitespace
+	 */
+	private _getTokenList(): string[] {
+		let attr = this._ownerElement.getAttribute(this._attributeName) ?? '';
+		attr = attr.replace(/^[\t\n\f\r ]+|[\t\n\f\r ]+$/g, '');
+		attr = attr.replace(/[\t\n\f\r ]+/g, ' ');
```
shirakaba, line 215: Hard to demonstrate as I’m typing on mobile, but String.prototype.split() accepts RegExp so this replace() and the next split() could both be reduced to one step.

takenspc: Thank you for the review. I have merged the two lines.

I learned a lot this time that String.prototype.split() accepts RegExp. Thank you again.

capricorn86 (dismissed):
Thank you for your contribution @takenspc! :star2: 

Sorry for taking such a long time before reviewing this.

I have made some improvements to the logic in `DOMTokenList.#getTokenList()` to improve its performance. I also merged the code and changed it to follow the new coding guidelines of using `Symbol()` or `#` for internal properties instead of "_" as prefix.

capricorn86 (approved):

PR #1170, opened 2023-11-29
#0@patch: Matches a partially invalid selector and throws an error.

btea:


capricorn86 (approved):

Issue #1176, opened 2023-12-08
customElements.getName() missing and invalid custom element name case handling.


mash-graz:
**Describe the bug**
1. The method `customElements.getName()` isn't implemented until now.

2. And while trying to fix this issues I had to figure out, that happy-dom makes some fundamental wrong assumptions about custom element naming. It internally translates all tag names to uppercase for case-insensitive lookups, but the specification to the contrary does only allow lowercase ASCII letters in valid custom element names anyway (https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name).

I'll try to send you a PR to fix these issues soon.

PR #1177, opened 2023-12-10
Implements customElements.getName() and fixes letter case handling.

mash-graz:
Adding the getName() method was a rather trivial task, but it doesn't make much sense without trying to fix the much more complicated letter case resp. case-sensitivity issues regarding the custom element registry, because otherwise we would return an invalid identifier.

The custom element specification requires names, that do not contain capital ASCII letters and case-sensitive letters of some Unicode ranges. It's closer the XML conventions than traditional HTML. To emphasize this difference I also used the variable name `localName` instead of `tagName` in the relevant code sections, because `element.tagName` will always use capital letters, while the corresponding `localName` mostly uses lowercase. 

Fixes: #1176


capricorn86:
I'm looking into this. Still some more things I want to check.

mash-graz:
> I'm looking into this. Still some more things I want to check.

There is no hurry. :)

Unfortunately I'm also rather busy with other jobs. Therefor also I couldn't review your extensive node-dependency-reduction-solution util now, but I will take a look as soon as possible.

capricorn86 (approved):

PR #1193, opened 2024-01-08
#1191@patch: Support for passing pseudo-selectors as argument of :not.

### LINKED ISSUE #1191: DOMException: The selector ":nth-of-type" is not valid.
image

In the case of nested selectors, it is not valid.

e.g.)
```css
p:not(:nth-of-type(1))
```

gdorsi:
Closes #1191


capricorn86 (approved):
Thank you for your contribution @gdorsi! :star2:

PR #1213, opened 2024-01-15
chore: [#1213] Replaces rm rf with a windows compatible script when building

btea:


capricorn86 (changes requested):
"rimraf" should be added as a dev dependency.

btea:
> rimraf pulls in 27 dependencies, so is generally nice to avoid: https://npmgraph.js.org/?q=rimraf
> 
> you can do this as a nicer alternative:
> 
> ```
> node -e "fs.rmSync('./cjs', { recursive: true, force: true })"
> ```

Thanks for the solution, it looks great. 👍

capricorn86 (approved):

Sorry that you had to wait for such a long time to get this in.

PR #1219, opened 2024-01-17
#1216@patch: Clone response body when cloning a response.

### LINKED ISSUE #1216: Cloned responses do not clone the underlying stream
**Describe the bug**
Doing a `fetch`, and then cloning the `Response` will give you 2 response objects.
In the browser, you can use these separately, and when you consume the Response, the other one is not consumed [(as described on MDN)](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone).
In happy-dom, consuming either of them will consume both, rendering the `.clone()` operation useless.

**To Reproduce**
```js
// test.mjs

const window = new Window();

const originalResponse = await window.fetch(
  "https://pokeapi.co/api/v2/pokemon/ditto"
);
const clonedResponse = originalResponse.clone();

console.log("ORIGINAL", await originalResponse.text()); // Will show the response
console.log("CLONED", await clonedResponse.text()); // Will return an empty string
```

- If you change the order of the two last lines, whichever comes first will work and the second one will not.
- Responses created from `Response.json({foo:"bar"})` do not have this bug
- If you immediately repeat the `fetch()`, both requests will work correctly (I suspect this is due to correctly implemented cloning for the internal cache).

**Expected behavior**
Both response objects return the correct response text even if the other one has been consumed.
You can copy the code (minus the `new Window()`) to your browser console and see it working correctly.

**Screenshots**
\-

**Device:**
node v20.11.0 on macOS Sonoma 14.2.1

**Additional context**
\-

cprecioso:
Closes #1216 

Simply clones the Response's body stream when cloning the response.


capricorn86 (approved):
Thank you for your contribution @cprecioso! :star2: 

The Fetch API was migrated to using `ReadableStream` instead of `Stream.Readable`. This caused some issues when I merged in the changes as the clone solution used `ReadableStream.tee()`, which requires the cloned streams to be consumed in parallel.

I added support for piping the underlying Node Stream, which solved the problem.

capricorn86 (approved):

PR #1224, opened 2024-01-18
#1223@patch: Relax `MutationObserver.observe` check for `options` argument.

### LINKED ISSUE #1223: `MutationObserver.observe` check for `options` argument is too strict
**Describe the bug**
`MutationObserver.observe` check for `options` argument is too strict and throws `DOMException` for the cases when it's valid according to the [DOM Standard](https://dom.spec.whatwg.org/#dom-mutationobserver-observe).

It should also throw `TypeError` and not `DOMException` in case when `options` object is invalid.

**To Reproduce**

1. Run the following code in 'happy-dom' environment
```js
let div = document.createElement('div');
let text = document.createTextNode('old');
let observer = new MutationObserver(() => console.log('mutated'));
observer.observe(div, { attributeFilter: ['style', 'class'] });
observer.observe(text, { characterDataOldValue: true });
```

**Expected behavior**
Should not throw `DOMException` or `TypeError`.

**Device:**
- Version `13.1.4`

**Additional context**
According to the [DOM Standard](https://dom.spec.whatwg.org/#dom-mutationobserver-observe) the following steps should be performed first before checking `options`.
1. If either `options["attributeOldValue"]` or `options["attributeFilter"]` exists, and `options["attributes"]` does not exist, then set `options["attributes"]` to `true`.
2. If `options["characterDataOldValue"]` exists and `options["characterData"]` does not exist, then set `options["characterData"]` to `true`.

Discovered while running tests against codebase with `VueUse` usage. Here `MutationObserver` is used with just `attributeFilter`: https://github.com/vueuse/vueuse/blob/84965f53edc73b221804de91e565ffdf71d1e28d/packages/core/useElementBounding/index.ts#L96-L98

romansp:
Fixes #1223 to better match [DOM Standard](https://dom.spec.whatwg.org/#dom-mutationobserver-observe).

Also added additional checks. All these throw `TypeError` in Chrome:
* `{ attributes: false, attributeOldValue: true }`
* `{ attributes: false, attributeFilter: ['class'] }`
* `{ characterData: false, characterDataOldValue: true }`


jajedidiah:
Screen Shot 2024-02-13 at 10 59 03 PM
@romansp any idea/timeline when this will be merged to master, ty

romansp:
@jajedidiah Unfortunately don't have an answer to that, awaiting for review from @capricorn86.

capricorn86 (approved):
Thank you for your contribution @romansp! :star2:

PR #1235, opened 2024-01-24
#1221@minor:  Add support for `:target` pseudo selector.

### LINKED ISSUE #1221: Add support for `:target` pseudo-selector
Hey :wave:,

I'v noticed in a component I'd like to test that happy-dom always returns `document` when calling `document.querySelector(':target')`.
After some digging I've found that happy-dom doesn't seem to support the [`:target` pseudo-selector](https://drafts.csswg.org/selectors/#the-target-pseudo) yet.

I've looked at the implementation of `querySelector` but I'm not feeling very comfortable to fix this myself as `querySelector` is such a performance critical method.

I've written a simple unit tests however to illustrate the expected behaviour of the `:target` selector 
 
```ts
it('Returns matched element matching :target', () => {
	window.location.hash = 'id'

	const div2 = document.createElement('div');

	div2.id = 'id';
	div.appendChild(div2);

	expect(div.querySelector(':target')).toBeNull();
	expect(document.querySelector(':target')).toBeNull();
	document.appendChild(div);

	expect(div.querySelector(':target')).toBe(div2);
	expect(document.querySelector(':target')).toBe(div2);

	window.location.hash = 'something-else';
	expect(div.querySelector(':target')).toBeNull();
	expect(document.querySelector(':target')).toBeNull();
})
```

Schleuse:
Fixes #1221 and implements `:target` pseudo selector.


Schleuse:
I've noticed that the `:target-within` implementation only checks direct children - I've marked the PR as draft and I will take a look again :sweat_smile:

Schleuse:
As `:target-within` seems to be still [very experimental in Browsers](https://caniuse.com/mdn-css_selectors_target-within) I've opted to remove the support for it

capricorn86 (approved):
Thank you for your contribution @Schleuse! :star2:

PR #1236, opened 2024-01-25
#1217@patch: Check if options are null.

### LINKED ISSUE #1217: [v13] Error: Cannot read properties of null (reading 'subtree')
**Describe the bug**

We use React and  [DevExtreme components](https://js.devexpress.com/). After updating to happy-dom 13, some of our tests are failing with the following message:

> TypeError: Cannot read properties of null (reading 'subtree')
 ❯ Function.removeChild node_modules/happy-dom/src/nodes/node/NodeUtility.ts:127:26
 ❯ Function.removeChild node_modules/happy-dom/src/nodes/element/ElementUtility.ts:122:15
 ❯ HTMLElement.removeChild node_modules/happy-dom/src/nodes/element/Element.ts:507:25
 ❯ node_modules/devextreme-react/cjs/core/component-base.js:150:183
 ❯ Popover.ComponentBase.componentWillUnmount node_modules/devextreme-react/cjs/core/component-base.js:150:77
 ❯ Popover.Component.componentWillUnmount node_modules/devextreme-react/cjs/core/component.js:68:47
 ❯ callComponentWillUnmountWithTimer node_modules/react-dom/cjs/react-dom.development.js:22857:14
 ❯ safelyCallComponentWillUnmount node_modules/react-dom/cjs/react-dom.development.js:22873:5
 ❯ commitDeletionEffectsOnFiber node_modules/react-dom/cjs/react-dom.development.js:24129:13

The problem can be reproduced with the following simple test:

```tsx
import { Popover } from "devextreme-react/popover";
import { render, screen } from "@testing-library/react";
import { test } from "vitest";

test("popover", async () => {
  render(
    <Popover visible>
      <div>Content</div>
    </Popover>
  );

  await screen.findByText("Content");
});
```

**To Reproduce**
1. Clone the [repro](https://github.com/rothsandro/repro.happy-dom-13-subtree)
2. Install dependencies: `npm i`
3. Run the tests via `npm run test`

**Expected behavior**
The tests should run successfully.

**Device:**
 - OS: macOS
 - Browser: Brave 120

**Additional context**

This only happens with happy-dom v13 and when [Vitest globals](https://vitest.dev/config/#globals) are enabled (even if we don't rely on globals but import the functions). It works when downgrading to happy-dom v12 or when disabling Vitest globals.

zachlankton:
Fixes #1217 

There are some instances where `observer.options` appears to be null or undefined.

In addition to the issue detailed in #1217 
I was also having issues with `childList` `characterData` and `attributes`

The change in this PR adds optional chaining to the affected code.

I believe this covers all code where `if (observer.options.*)` occurs.


zachlankton:
Hi @capricorn86 have you had a chance to peak at this? Pretty small change and fixes some bugs for me and @rothsandro

capricorn86 (approved):
Thank you @zachlankton for your contribution! :star2:

Issue #1181, opened 2023-12-15
`Object.keys(localStorage)` returns ['_store'] not the keys stored


knownasilya:
**Describe the bug**
`Object.keys(localStorage)` returns ['_store'] not the keys stored
You have to do `Object.keys(localStorage._store)` to get the keys. 
This also happens with for-in loop.

Using "happy-dom": "^12.10.3"

**To Reproduce**
1. Set values on localStorage
2. Run `Object.keys(localStorage)`
3. See incorrect result

**Expected behavior**
Object.keys(localStorage) should return keys that were set.

**Device:**
 - OS: macos m1 latest version

knownasilya:
Workaround:

```js
for (let i = 0; i < localStorage.length; i++) {
    const key = localStorage.key(i);
    // access each key
}
```

XiaoDaiGua-Ray:
I encountered the same situation as you, after emulating the storage of a value, I used `Object.keys(window.localStorage)` to get an empty array.

<img width="1376" alt="image" src="https://github.com/capricorn86/happy-dom/assets/51957438/cda1a15d-f9c7-4372-b298-4fa6e05146bb">

capricorn86:
Than you for reporting @knownasilya! and @XiaoDaiGua-Ray! :slightly_smiling_face: 

A big thanks to @frankdiw for contributing with a fix :star2: 

You can read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v14.3.1

PR #1254, opened 2024-02-05
fix: [#1181] Refactor storage mock

frankdiw:
FIxed: #1181


capricorn86 (approved):
Thank you for your contribution @frankdiw! :star2: 

I removed the added classes `LocalStorage` and `SessionStorage`, as `Storage` is the correct class to use for the properties `localStorage` and `sessionStorage` according to spec.

Mozilla docs:
 - [Window.localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)
 - [Window.sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)
 
 I also added some more tests to cover that adding properties to `Storage` won't break in the future.

capricorn86:
By the way, sorry that it took such a long time before merging!

PR #1284, opened 2024-03-01
Fix: Fragile cloneNode implementation

itutto:
### Sample to recreate the issue
```js
// sample.mjs
import * as happyDom from 'happy-dom';

const win = new happyDom.Window();
const doc = win.document;

class Test extends win.HTMLElement {
    attributeChangedCallback() {
        // This will fail because the cloned element does not have a tagname (yet).
        this.querySelectorAll('label');
        /**
         * SelectorItem.js:42
            if (this.tagName !== '*' && this.tagName !== element[PropertySymbol.tagName].toUpperCase()) {
         */
    }
}
Test.observedAttributes = ['my-test'];

win.customElements.define('x-test',Test);

const template = doc.createElement('template');
template.innerHTML = `<template><x-test my-test="fail"></x-test></template>`;
template.content.cloneNode(true)
```
Run
```shell
node sample.mjs
```

### What is the fix?
I propose to move the tagName, localName and namespaceUri initialization before the attribute clone.


capricorn86 (approved):
Thank you for your contribution @itutto! :star2:

Issue #1285, opened 2024-03-01
Node.contains doesn't handle undefined args


rofly:
**Describe the bug**

Calling Node.contains with undefined argument raises a TypeError exception.

`"TypeError": Cannot read properties of undefined (reading 'Symbol(parentNode)')`

**To Reproduce**

Run the following code snippet

```javascript
const happyDom = require("happy-dom");

const happyWindow = new happyDom.Window();
const parentElement = happyWindow.document.createElement('div');
const childElement = happyWindow.document.createElement('div');

parentElement.appendChild(childElement);

let nonInitializedElement;

parentElement.contains(nonInitializedElement);
```

**Expected behavior**

The same behavior that happens in the browser.
Node.contains should return false when it's called with `undefined`

PR #1286, opened 2024-03-01
fix: [#1285] Prevent contains from accessing properties of undefined args

rofly:
This was the first solution that came to mind to solve https://github.com/capricorn86/happy-dom/issues/1285.


capricorn86 (approved):
Thank you @rofly for your contribution! :star2:

Issue #1138, opened 2023-10-19
Wrong select value when option with empty value is selected


modiho:
**Describe the bug**

A select element reports the text content of an option that has an empty value as its value, instead of the empty string.

**To Reproduce**

Execute the following in a Node.js repl:

```javascript
let {GlobalRegistrator} = require('@happy-dom/global-registrator')

GlobalRegistrator.register();

document.body.innerHTML = `<form>
  <select name="size">
    <option value="">Please choose</option>
    <option value="s">S</option>
    <option value="m">M</option>
  </select>
</form>`

document.querySelector('select').value // => "Please choose"
```

**Expected behavior**

```javascript
document.querySelector('select').value // => "" (empty string)
```

**Screenshots**

**Device:**

 - Node.js v18.14.1

karpiuMG:
@capricorn86 i see the fix was merged - let's close the issue then

PR #1293, opened 2024-03-10
fix: [#1138] Modify option node to return empty string even if the value is empty string

t-wataru:
This might fix "Wrong select value when option with empty value is selected #1138 ".


capricorn86 (approved):
Thank you @t-wataru for your contribution! :star2:

Issue #1308, opened 2024-03-14
If multiple custom property fallbacks occur, `getPropertyValue` returns an empty string


odanado:
**Describe the bug**
If multiple custom property (CSS Variables) fallbacks occur, `getPropertyValue` returns an empty string.
`var()` falls back to the second argument's value if no custom property is defined.
```css
.one {
  /* Red if --my-var is not defined */
  color: var(--my-var, red);
}
.two {
  /* pink if --my-var and --my-background are not defined */
  color: var(--my-var, var(--my-background, pink));
}
```
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#defining_fallbacks_in_the_var_function

If the fallback is one time, `getPropertyValue` works correctly, but if it is two or more times, it returns an empty string.
ref: https://github.com/odan-sandbox/happy-dom-getComputedStyle-with-css-variables-sandbox/blob/bcffcab32a5c6299ea41535bcfd9cb7659df3354/src/index.js#L26-L30

**To Reproduce**
1. Go to https://github.com/odan-sandbox/happy-dom-getComputedStyle-with-css-variables-sandbox
2. Clone this repository
3. run `npm i`
4. run `node src/index.js`

**Expected behavior**
This case returns `"pink"`
https://github.com/odan-sandbox/happy-dom-getComputedStyle-with-css-variables-sandbox/blob/bcffcab32a5c6299ea41535bcfd9cb7659df3354/src/index.js#L26-L27

This case returns `"blue"`
https://github.com/odan-sandbox/happy-dom-getComputedStyle-with-css-variables-sandbox/blob/bcffcab32a5c6299ea41535bcfd9cb7659df3354/src/index.js#L29-L30

**Device:**
 - Node: v18.12.1
 - Version: 13.8.4

PR #1309, opened 2024-03-14
fix: [#1308] Fix a bug of multiple fallbacks of custom property

odanado:
Fix: https://github.com/capricorn86/happy-dom/issues/1308

The previous code was incorrect in its use of [RegExp.prototype.exec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec), and the regular expression matching was performed only once.

`RegExp.prototype.exec` is difficult to use correctly because it is a stateful API.
I modified the logic using [String.prototype.matchAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) as described in MDN.

> In addition, [String.prototype.matchAll()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) helps to simplify matching multiple parts of a string (with capture groups) by allowing you to iterate over the matches.
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#description

### DIFF
--- a/packages/happy-dom/src/css/declaration/element-style/CSSStyleDeclarationElementStyle.ts
+++ b/packages/happy-dom/src/css/declaration/element-style/CSSStyleDeclarationElementStyle.ts
@@ -366,14 +366,9 @@ export default class CSSStyleDeclarationElementStyle {
 	private parseCSSVariablesInValue(value: string, cssVariables: { [k: string]: string }): string {
 		const regexp = new RegExp(CSS_VARIABLE_REGEXP);
 		let newValue = value;
-		let matches;
+		let match;
 
-		while ((matches = newValue.matchAll(regexp)) !== null) {
-			const match = matches.next().value;
-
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Thank you for contributing @odanado! :star2: 

`matchAll()` will execute a find all and return an iterator for each loop in the while statement, which will not have very good performance.

I instead corrected the `exec()` to run on the modified value, which solved the problem.

capricorn86:
Sorry @odanado, I was a bit fast. `exec()` won't work, so I changed it to this implementation using `match()` instead (which should also have good performance):
https://github.com/capricorn86/happy-dom/pull/1312

Not sure why the unit tests passed in the PR :thinking:

Issue #1315, opened 2024-03-16
Implement the getSetCookie method of Headers


ushiboy:
**Is your feature request related to a problem? Please describe.**
In the happy-dom environment, the test fails because the getSetCookie method of Headers does not exist.

```
src/sample.test.ts > Headers.getSetCookie()
TypeError: headers.getSetCookie is not a function
 ❯ src/sample.test.ts:3:18
      1| test("Headers.getSetCookie()", () => {
      2|   const headers = new Headers();
      3|   expect(headers.getSetCookie()).toEqual([]);
       |                  ^
      4| 
```

**Describe the solution you'd like**
I'd like to see Headers getSetCookie implemented according to the specification.
see: https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie

**Describe alternatives you've considered**
None.

**Additional context**
None.

PR #1316, opened 2024-03-16
feat: [#1315] Implement the getSetCookie method of Headers

ushiboy:
#1315 

Implemented the getSetCookie method of Headers.


capricorn86 (approved):
Thank you for contributing @ushiboy! :star2: 

I think that is safer to just add support for multiple values to Headers instead of splitting the cookie string, so I changed the logic in Headers and removed the utility used for splitting.

I hope you are fine with the solution :slightly_smiling_face:

ushiboy:
Thanks @capricorn86 

I agree that changing the data structure would be a more appropriate approach. 
I hope that method works well.

Issue #1347, opened 2024-03-22
localStorage.setItem  non-string value(such as 1, {})  not behavior as expected


dr2009:
**Describe the bug**
```js
// happy-dom@14.3.1
localStorage.setItem('a', {})  // {}, expect '[object Object]'
localStorage.setItem('a', 1)  // 1, expect '1'
```

**To Reproduce**

**Expected behavior**
```js
localStorage.setItem('a', {})  // '[object Object]'
localStorage.setItem('a', 1)  // '1'
```

**Screenshots**

**Device:**

**Additional context**
happy-dom@14.3.1

PR #1348, opened 2024-03-22
fix: [#1347] LocalStorage.setItem non-string value should to be string

dr2009:


capricorn86 (approved):
Thank you for your contribution @dr2009! :star2:

Issue #825, opened 2023-03-12
`iframe.sandbox` should return a DOMTokenList instead of a string


saschanaz:
**Describe the bug**
As the title says.

https://github.com/capricorn86/happy-dom/blob/13bcfe77ae9a202a93ce8fddcf4a7c1b43ecde16/packages/happy-dom/src/nodes/html-iframe-element/HTMLIFrameElement.ts#L121-L123

**To Reproduce**
On the REPL console:

```js
dom = await import("happy-dom")
(new dom.Window()).document.createElement("iframe").sandbox
```

**Expected behavior**

A DOMTokenList should be returned.

**Screenshots**

**Device:**
 - OS: Windows
 - Browser: Node.js
 - Version: 19.6.0

**Additional context**

See https://html.spec.whatwg.org/multipage/iframe-embed-object.html#htmliframeelement for the spec.

capricorn86:
Thank you for reporting @saschanaz! :slightly_smiling_face: 

Big thanks to @jeffwcx to make a contribution :star: 

Read more about the release here:
https://github.com/capricorn86/happy-dom/releases/tag/v14.4.0

PR #1375, opened 2024-03-31
feat: [#825] Attribute iframe.sandbox should return a DOMTokenList

jeffwcx:
According to the specifications:

https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-sandbox
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox

The sandbox attribute of the HTMLIFrameElement should return a DOMTokenList.

Additionally, although sandbox works like a regular DOMTokenList, it is checked after the attribute is set. If there are tokens that do not comply with the specification, an error message will be printed (but no error will be thrown).

There is a bug in DOMTokenList where, after setting the value, operations like sandbox[0] do not take effect. So we need an update of the indices.

### DIFF
--- a/packages/happy-dom/src/PropertySymbol.ts
+++ b/packages/happy-dom/src/PropertySymbol.ts
@@ -158,3 +158,4 @@ export const navigator = Symbol('navigator');
 export const screen = Symbol('screen');
 export const sessionStorage = Symbol('sessionStorage');
 export const localStorage = Symbol('localStorage');
+export const sandbox = Symbol('sandbox');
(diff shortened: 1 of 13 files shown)


jeffwcx:
Hi, @capricorn86 can you take a look?

capricorn86 (approved):
Thank you for contributing @jeffwcx! :star: 

I made some fixes to the solution. Attributes are updated through a `NamedNodeMap`. For this element the class `HTMLIFrameElementNamedNodeMap` is used and this is where the indices should be updated and the validation should happen.

PR #1379, opened 2024-04-04
feat: Adds support for application/x-www-form-urlencoded in Request.formData

tt-public:
Currently `Request.formData` only handles "multipart/form-data", but it seems that it should also support "application/x-www-form-urlencoded" according to the specs. https://fetch.spec.whatwg.org/#dom-body-formdata

Current behavior:
```typescript
const params = new URLSearchParams();
params.append("d", "data");

const req = new Request("https://example.com", { method: "POST", body: params });

for (const [key, value] of req.headers.entries()) {
    console.log(`${key}: ${value.split(";")[0]}`);
}
// content-type: application/x-www-form-urlencoded

(async () => {
    for (const [key, value] of (await req.formData()).entries()) {
        console.log(`${key}: ${value}`);
    }
    // DOMException [InvalidStateError]: Failed to build FormData object: The "content-type" header isn't of type "multipart/form-data".
})();
```

The current implementation of `Request.formData` in happy-dom directly calls a method which seems to be based on `toFormData` in node-fetch. [[source](https://github.com/capricorn86/happy-dom/blob/75041b205dc7a24861599ce08262c8f30241afb3/packages/happy-dom/src/fetch/multipart/MultipartFormDataParser.ts#L12)] However, `Request.formData` in node-fetch only calls `toFormData` after handling "application/x-www-form-urlencoded" [[source](https://github.com/node-fetch/node-fetch/blob/8b3320d2a7c07bce4afc6b2bf6c3bbddda85b01f/src/body.js#L110)], which probably should be reflected in this project too.

Thank you for the awesome project. I hope this helps!

### DIFF
--- a/packages/happy-dom/src/fetch/Request.ts
+++ b/packages/happy-dom/src/fetch/Request.ts
@@ -304,29 +304,50 @@ export default class Request implements Request {
 	 * @returns FormData.
 	 */
 	public async formData(): Promise<FormData> {
-		if (this.bodyUsed) {
-			throw new DOMException(
-				`Body has already been used for "${this.url}".`,
-				DOMExceptionNameEnum.invalidStateError
-			);
-		}
+		const type = this[PropertySymbol.contentType];
 
-		(<boolean>this.bodyUsed) = true;
+		if (/multipart/i.test(type)) {
+			if (this.bodyUsed) {
+				throw new DOMException(
+					`Body has already been used for "${this.url}".`,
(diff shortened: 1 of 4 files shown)


capricorn86 (approved):
Thank you for contributing @tt-public! :star:

I added some unit tests and synced the logic between `Request.formData()` and `Response.formData()`.

I'm not sure why this was missed as `Response.formData()` had the logic for handling "application/x-www-form-urlencoded".

PR #1381, opened 2024-04-04
fix: [#1363] Fix bug with missing fallback value

### LINKED ISSUE #1363: Incorrect return value of `getPropertyValue` for nested custom properties without fallback value
**Describe the bug**
An incorrect return value of `getPropertyValue` for nested custom properties without fallback value.

```js

const window = new Window({ url: "https://localhost:8080" });

const div = window.document.createElement("div");

const style = window.document.createElement("style");

style.textContent = `
  div {
    --my-color1: pink;
    --my-color2: red;

    border-color: var(--my-color1);

    color: var(--my-color1, var(--my-color2, blue));

    background-color: var(--my-color1, var(--my-color2));
  }
`;

window.document.head.appendChild(style);
window.document.body.appendChild(div);

// OK: expected: pink, actual: pink
console.log(window.getComputedStyle(div).getPropertyValue("border-color"));

// with fallback value
// OK: expected: pink, actual: pink
console.log(window.getComputedStyle(div).getPropertyValue("color"));

// without fallback value
// NG: expected: red, actual: "" (empty string)
console.log(window.getComputedStyle(div).getPropertyValue("background-color"));
```

**To Reproduce**
1. Run `git clone https://github.com/odan-sandbox/happy-dom-getComputedStyle-with-css-variables-sandbox`
2. Run `npm ci`
3. Run `node src/repro-non-fallback.js`

**Expected behavior**
`--my-color1` is defined, so the `background-color` should be pink.

**Device:**
 - Node.js: v18.12.1
 - happy-dom: 14.3.6

odanado:
Close #1363 

Custom properties without fallback values can be converted to custom properties with fallback values (using `SINGLE_CSS_VARIABLE_REGEXP`).

e.g.
- `var(--known)`
  - In this case, you can resolve the value of the custom property by examining the value of `--known`.
- `var(--unknown, var(--known)` 
  - This case can be converted to a custom property with a fallback value of the form `var(--unknown, known-color)` by resolving the `--known` value.
- `var(--unknown, var(--unknown, var(--known)))`
  - This case can also be converted to a custom property with a fallback value of the form `var(--unknown, var(--unknown, known-color))` by resolving the value of `--known`.

After converting to a custom property for which a fallback value exists, the value can be found using a regular expression that handles the custom property in this case (using `CSS_VARIABLE_REGEXP`).

### DIFF
--- a/packages/happy-dom/src/css/declaration/element-style/CSSStyleDeclarationElementStyle.ts
+++ b/packages/happy-dom/src/css/declaration/element-style/CSSStyleDeclarationElementStyle.ts
@@ -20,6 +20,7 @@ import MediaQueryList from '../../../match-media/MediaQueryList.js';
 import WindowBrowserSettingsReader from '../../../window/WindowBrowserSettingsReader.js';
 
 const CSS_MEASUREMENT_REGEXP = /[0-9.]+(px|rem|em|vw|vh|%|vmin|vmax|cm|mm|in|pt|pc|Q)/g;
+const CSS_VARIABLE_REGEXP = /var\( *(--[^), ]+)\)|var\( *(--[^), ]+), *(.+)\)/;
 
 type IStyleAndElement = {
 	element: Element | ShadowRoot | Document;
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Thank you for your contribution @odanado! :star: 

I simplified the solution a bit by fixing the original RegExp, which will only require one white loop.

PR #1383, opened 2024-04-05
feat: [#1124] Implement inert attribute

### LINKED ISSUE #1124: Support `HTMLElement#inert` property
Hi, 
according to the HTML Spec (https://html.spec.whatwg.org/multipage/interaction.html#dom-inert) the HTMLElement has the `inert` property which reflects its `inert` boolean attribute and the inert state of the DOM element (https://html.spec.whatwg.org/multipage/interaction.html#inert) (no focus, click, etc).

happy-dom seems not yet to support this. e.g.

```js

const window = new Window();
document.body.innerHTML = '<div class="container"></div>';

container.inert = true;

console.log(container.outerHTML); // will print `<div class="container"></div>` instead of `<div class="container" inert></div>`
```

cheers

odanado:
Close #1124 

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert


capricorn86 (approved):
Thank you for your contribution @odanado! :star:

Issue #1398, opened 2024-04-10
Add support for the srcdoc attribute of iframe


jeffwcx:
The current iframe implementation lacks support for srcdoc.

PR #1399, opened 2024-04-10
feat: [#1398] Add support for iframe srcdoc

jeffwcx:
Implement the iframe srcdoc attribute, which should operate similarly to the src attribute but with higher priority, without the need to initiate a fetch request.

Also, I have fixed several bugs: 
+ When remove the src or srcdoc attributes, the iframe should be reloaded.
+ For non-cross-domain iframes, after being connected to the Document, their contentWindow’s parent/top properties should not be null.

References:
+ [W3C-the-iframe-element.html#attr-iframe-src](https://www.w3.org/TR/2010/WD-html5-20100624/the-iframe-element.html#attr-iframe-src)

### DIFF
--- a/packages/happy-dom/src/browser/types/IGoToOptions.ts
+++ b/packages/happy-dom/src/browser/types/IGoToOptions.ts
@@ -14,4 +14,8 @@ export default interface IGoToOptions extends IReloadOptions {
 	 * Referrer policy.
 	 */
 	referrerPolicy?: IRequestReferrerPolicy;
+	/**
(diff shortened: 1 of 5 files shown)


jeffwcx:
@capricorn86 can you take a look?

capricorn86 (approved):
Thank you @jeffwcx for your contribution! :star2: 

I've added a fix to isolate the implementation to the iframe implementation, as "srcdoc" is not really related to navigating a Browser frame.

Issue #1161, opened 2023-11-14
element.getBoundingClientRect().toJSON() does not exist


dragonofdeath:
**Describe the bug**
element.getBoundingClientRect().toJSON() throws as its not implemented. 

getBoundingClientRect() returns [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect) which inherits its properties from [DOMRectReadOnly](https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly) and it has `toJSON` method, which is listed in mdn and is supported by all browsers. In my jsdom based tests I was using `jsdom-pollyfill` to solve the usecase  and the implementaction code it is was using is something like this https://github.com/remirror/remirror/blob/main/packages/jest-remirror/src/jsdom-polyfills.ts#L90-L92 Would you mind PR adding this method?

capricorn86:
Thank you for reporting @dragonofdeath  ! :slightly_smiling_face: 

Big thanks to @domakas for contributing with a fix :star: 

https://github.com/capricorn86/happy-dom/releases/tag/v14.8.1

PR #1402, opened 2024-04-12
fix: [#1161] Implement DOMReact and DOMReactReadOnly interfaces

domakas:
Resolves https://github.com/capricorn86/happy-dom/issues/1161

Implements `DOMReact` and `DOMReactReadOnly` classes according to spec https://drafts.fxtf.org/geometry/#DOMRect


capricorn86 (approved):
Thank you for your contribution @domakas! :star2: 

I changed the protected properties to be Symbol properties instead, so that they won't be accessible by the consumer. I also added a conversion to Number() according to spec in the constructor.

Issue #1349, opened 2024-03-22
Expose document.forms.namedItem function


martialanouman:
**Is your feature request related to a problem? Please describe.**
I am creating a Remix app that uses @conform-to/react to manage the forms.

The new v1.0.4 version is using document.forms.namedItem() function but it is not implemented by happy-dom. That does not allow me to test form interaction with happy-dom

Here is the error i got:
```shell
TypeError: document.forms.namedItem is not a function
 ❯ getFormElement node_modules/.pnpm/@conform-to+dom@1.0.4/node_modules/@conform-to/dom/form.mjs:266:27
 ❯ Object.submit node_modules/.pnpm/@conform-to+dom@1.0.4/node_modules/@conform-to/dom/form.mjs:327:24
 ❯ onSubmit node_modules/.pnpm/@conform-to+react@1.0.4_react@18.2.0/node_modules/@conform-to/react/context.mjs:205:28
 ❯ submitHandler node_modules/.pnpm/react-router-dom@6.22.3_react-dom@18.2.0_react@18.2.0/node_modules/react-router-dom/index.tsx:1264:19
 ❯ HTMLUnknownElement.callCallback node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js:4164:14
 ❯ HTMLUnknownElement.dispatchEvent node_modules/.pnpm/happy-dom@14.3.1/node_modules/happy-dom/src/event/EventTarget.ts:222:42
 ❯ HTMLUnknownElement.dispatchEvent node_modules/.pnpm/happy-dom@14.3.1/node_modules/happy-dom/src/nodes/element/Element.ts:1174:29
 ❯ HTMLUnknownElement.dispatchEvent node_modules/.pnpm/happy-dom@14.3.1/node_modules/happy-dom/src/event/EventTarget.ts:131:9
 ❯ HTMLUnknownElement.dispatchEvent node_modules/.pnpm/happy-dom@14.3.1/node_modules/happy-dom/src/nodes/element/Element.ts:1174:29
 ❯ Object.invokeGuardedCallbackDev node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js:4213:16
```

**Describe the solution you'd like**
That `document.forms.namedItem` should be a function, not undefined

**Describe alternatives you've considered**
I can't test it as this is how that library implements the useForm hook.

**Additional context**
[...]

PR #1404, opened 2024-04-12
fix: [#1349] Return forms as html collection

jean-leonco:
Return `document.forms` as `HTMLCollection` instead of `NodeList`.

https://developer.mozilla.org/en-US/docs/Web/API/Document/forms
image

Related to #1349.


Amorim33 (approved):

capricorn86 (approved):
Thank you for your contribution @jean-leonco! :star2:

PR #1407, opened 2024-04-13
fix: [1406] Fixes issue with insertBefore and comment reference node

### LINKED ISSUE #1406: insertBefore with a comment node causes duplication of node
## Description

Using `insertBefore` with a comment node results in copying the target node instead of moving it

```js
// Create container
const container = document.createElement("div");
document.body.appendChild(container);

// Create child
const child = document.createElement("p");
child.textContent = "Hi";
container.appendChild(child);

// Create comment node
const comment = document.createComment("");
container.appendChild(comment);

// Query nodes, it returns "['Hi']", as expected
console.log(document.querySelectorAll("p").map(item => item.textContent));

// Try to move node
container.insertBefore(child, comment);

// Query nodes, "['Hi']" expected, but it returns "['Hi', 'Hi']"
console.log(document.querySelectorAll("p").map(item => item.textContent));
```

## Steps to reproduce

1. Create a DOM element and add it to the container
2. Create a comment node and add it to the same container
3. Attempt to move the element using insertBefore, using the comment node as the reference node

Expected result:

The target node should be moved or remain untouched

Actual result:

The target node is duplicated

## Reproduction

I have created a repository with a minimal reproduction of the issue

[https://github.com/mdafanasev/happy-dom-insert-before-comment-issue](https://github.com/mdafanasev/happy-dom-insert-before-comment-issue)

## Environment

@happy-dom/global-registrator: 14.7.1

NodeJS: 20.11.0

mdafanasev:
Resolves #1406.

I found that when the reference node is not an Element, such as a comment, the insertion correctly occurs in the `childNodes` array, but the target node is not removed from `childNodes`; it is only removed from `children`. Therefore, I have added code that also removes the target node from `childNodes`.

### DIFF
--- a/packages/happy-dom/src/nodes/element/ElementUtility.ts
+++ b/packages/happy-dom/src/nodes/element/ElementUtility.ts
@@ -171,6 +171,13 @@ export default class ElementUtility {
 						parentNodeChildren.splice(index, 1);
 					}
 				}
+				const parentChildNodes = (<Element>ancestorNode)[PropertySymbol.childNodes];
+				if (parentChildNodes) {
+					const index = parentChildNodes.indexOf(newNode);
+					if (index !== -1) {
+						parentChildNodes.splice(index, 1);
+					}
+				}
 			}
 
 			const ancestorNodeChildren = <HTMLCollection<HTMLElement>>(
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Thank you for your contribution @mdafanasev! :star2: 

In case you are interested, i'm working on a fix that removes `ElementUtility` and refactors the code to simplify the logic to avoid these kind of bugs :slightly_smiling_face: 
https://github.com/capricorn86/happy-dom/pull/1359/files#diff-f84cace34f3e54afcf796c8ae3f1f853f7abeb0451a8d52e2213db8857ae58dcR127

PR #1408, opened 2024-04-14
feat: Add `fromRect` method

btea:


capricorn86 (approved):

This was already implemented in #1402. However, I have added the additional test that you added.

PR #1412, opened 2024-04-22
feat: [#1411] Add queryCommandSupported method.

### LINKED ISSUE #1411: Add support for document.queryCommandSupported()
https://stackoverflow.com/questions/78225627/dealing-with-the-deprecated-document-querycommandsupported

I think we can start with a simple implementation that just returns `true`.

btea:
close #1411


capricorn86 (approved):

PR #1421, opened 2024-04-28
ci: test on node 22

btea:
Node 16 is already eof and version 22 is released, so we can update it.


capricorn86:
@btea we need to change the "engines" property in the "package.json" files as well. However, this would be considered a breaking change and might be something better to combine with other breaking changes in a major release.

btea:
Indeed, we should change the `engine` field to `>=18.0.0`.

capricorn86:
@btea, I fixed this now in https://github.com/capricorn86/happy-dom/pull/1359. As it was a breaking change, I wanted to release it as part of other major changes.

PR #1437, opened 2024-05-14
fix: [#1336] Adds check for if Window and MutationObserver has been d…

capricorn86:
…estroyed when triggering listeners


thaggie:
Thanks, that was fast! 

After creating the ticket I build the library locally to a) test that approach (which works fine) and b) try and work out the root cause.

The root cause was that we had old code in our jest setup that was providing it's own versions of `requestAnimationFrame` / `cancelAnimationFrame` - removing these stopped the issue for us.

PR #1465, opened 2024-06-17
fix: [#1464] Fix react click handler with `preventDefault` on anchor not preventing navigation

### LINKED ISSUE #1464: React `onClick` handler with `preventDefault()` on an anchor tag does not prevent navigation
**Describe the bug**

An `<a href="https://example.com">Click me</a>` tag with a click handler that invokes `event.preventDefault()` should prevent the subsequent navigation.

When doing this with pure DOM APIs:

```ts
const anchor = document.createElement('a');
anchor.href = 'http://example.com';
anchor.textContent = 'Click me';
anchor.addEventListener('click', (e) => e.preventDefault());
document.body.appendChild(anchor);

anchor.click();
expect(location.href).not.toBe('http://example.com');
```

This passes.

When doing the same, but passing through React (and its synthetic event handling):

```tsx
render(
  <a href="http://example.com" onClick={(e) => e.preventDefault()}>
    Click me
  </a>
);

screen.getByRole('link').click();
expect(location.href).not.toBe('http://example.com');
```

This fails, but should pass.

**To Reproduce**

https://stackblitz.com/edit/stackblitz-starters-x99sjj?file=package.json

The "Pure DOM"-based test passes
The React-based one fails

**Expected behavior**
The React test should pass

**Additional context**
In real DOM and in JSDom this works as expected

amitdahan:
Closes #1464

### DIFF
--- a/packages/happy-dom/src/nodes/html-anchor-element/HTMLAnchorElement.ts
+++ b/packages/happy-dom/src/nodes/html-anchor-element/HTMLAnchorElement.ts
@@ -500,8 +500,7 @@ export default class HTMLAnchorElement extends HTMLElement implements IHTMLHyper
 		if (
 			event.type === 'click' &&
 			event instanceof MouseEvent &&
-			(event.eventPhase === EventPhaseEnum.atTarget ||
-				event.eventPhase === EventPhaseEnum.bubbling) &&
+			event.eventPhase === EventPhaseEnum.none &&
 			!event.defaultPrevented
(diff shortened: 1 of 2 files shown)


amitdahan:
Was really not sure about the eventual solution, but it seemed to do the trick and all other tests remain passing.

capricorn86 (approved):
Thank you for your contribution @amitdahan! :star2:

Issue #1468, opened 2024-06-25
Add support for the `AbortSignal.any()` static method


ezzatron:
**Is your feature request related to a problem? Please describe.**
[AbortSignal.any()](https://developer.mozilla.org/docs/Web/API/AbortSignal/any_static) is now widely available, and it would be great to have this available in `happy-dom` too.

**Describe the solution you'd like**
I am creating this issue so I have an issue number to use in the commit message for a PR I intend to submit to add `AbortSignal.any()`.

**Describe alternatives you've considered**
Currently I polyfill this manually in every project that needs it.

**Additional context**
N/A

PR #1469, opened 2024-06-25
feat: [#1468] Add AbortSignal.any() static method

ezzatron:
This PR adds the [`AbortSignal.any()` static method](https://developer.mozilla.org/docs/Web/API/AbortSignal/any_static).

This method is extremely useful for combining multiple signals together. An example would be adding a timeout to a manually aborted signal:

```ts
const timeout = AbortSignal.timeout(10000);
const controller = new AbortController();

const signal = AbortSignal.any([timeout, controller.signal]);

// Will abort after 10 seconds, or when controller.abort() is called
await fetch("https://example.org/, { signal });
```


capricorn86 (approved):
Thank you for your contribution @ezzatron! :star2:

Sorry that you had to wait for so long time before getting this in.

capricorn86:
@ezzatron This PR got blocket because some of the integration tests failed. The integration tests loads Github.com, and apparently they released something that broke Happy DOM. I'm working on it, but perhaps it would be good to integration test a more stable target in the future.

capricorn86 (approved):

Issue #1470, opened 2024-06-25
Add support for the `AbortSignal.timeout()` static method


ezzatron:
**Is your feature request related to a problem? Please describe.**
[AbortSignal.timeout()](https://developer.mozilla.org/docs/Web/API/AbortSignal/timeout_static) is now widely available, and it would be great to have this available in `happy-dom` too.

**Describe the solution you'd like**
I am creating this issue so I have an issue number to use in the commit message for a PR I intend to submit to add `AbortSignal.timeout()`.

**Describe alternatives you've considered**
Currently I polyfill this manually in every project that needs it.

**Additional context**
There's a feature of `AbortSignal.timeout()` which is documented on MDN that interacts with the browser "stop" button. I'm not sure if there is a way to simulate this kind of interaction with `happy-dom`:

> The signal aborts with a `TimeoutError` [`DOMException`](https://developer.mozilla.org/docs/Web/API/DOMException) on timeout, or with `AbortError` [`DOMException`](https://developer.mozilla.org/docs/Web/API/DOMException) due to pressing a browser stop button (or some other inbuilt "stop" operation). This allows UIs to differentiate timeout errors, which typically require user notification, from user-triggered aborts that do not.

PR #1471, opened 2024-06-25
feat: [#1470] Add AbortSignal.timeout() static method

ezzatron:
This PR adds the [`AbortSignal.timeout()` static method](https://developer.mozilla.org/docs/Web/API/AbortSignal/timeout_static).

This method is useful for initiating async operations that should abort after a pre-determined amount of time:

```ts
// Will abort after 10 seconds
await fetch("https://example.org/, { signal: AbortSignal.timeout(10000) });
```

I could use some advice on whether it's also possible in `happy-dom` to emulate this interaction with the browser "stop" button mentioned on MDN (I haven't attempted to cover the functionality in this PR):

> The signal aborts with a `TimeoutError` [`DOMException`](https://developer.mozilla.org/docs/Web/API/DOMException) on timeout, or with `AbortError` [`DOMException`](https://developer.mozilla.org/docs/Web/API/DOMException) due to pressing a browser stop button (or some other inbuilt "stop" operation). This allows UIs to differentiate timeout errors, which typically require user notification, from user-triggered aborts that do not.


efredin:
Would be great to see this PR progress.

capricorn86:
@ezzatron This PR got blocket because some of the integration tests failed. The integration tests loads Github.com, and apparently they released something that broke Happy DOM. I'm working on it, but perhaps it would be good to integration test a more stable target in the future.

ezzatron:
@capricorn86 No worries! I panicked for a second and thought I had added a bad test. But you're talking about a test that was already in the suite, right?

capricorn86:
@capricorn86 yes exactly, it's already in the master branch :sweat_smile:

capricorn86 (approved):

capricorn86:
@ezzatron this PR has finally been merged :slightly_smiling_face: 

I had to fix, so that we use "setTimeout()" from the Window object, so that it is possible to cancel the timeout during teardown etc.

PR #1473, opened 2024-06-27
fix: [#1472] Fix child combinator matching in nested element

### LINKED ISSUE #1472: `element.matches()` returns incorrect result for nested selector
**Describe the bug**

`element.matches(".k > .k")` returns incorrect result.

**To Reproduce**

Run the following code, or open https://stackblitz.com/edit/stackblitz-starters-dngcid?file=index.mjs,package.json and run `node index.mjs`.

```ts
import * as happyDom from 'happy-dom';

function test(win) {
  let document = win.document;
  document.body.innerHTML = `
    <div>
      <div class="k a">
        <div>
          <div class="k b">
          </div>
        </div>
      </div>
    </div>`;

  const a = document.querySelector('.a');
  const b = document.querySelector('.b');

  const selector = '.k > .k';

  // Expected "k a false", actual "k a false"
  console.log(a?.className, a?.matches(selector));

  // Expected "k b false", actual "k b true"
  console.log(b?.className, b?.matches(selector));
}

test(new happyDom.Window());
```

**Expected behavior**

`b.matches(selector)` should return `false`, but it returns `true`.

 
**Device:**
 - OS: macOS
 - Browser Chrome
 - Version 14.12.3

ocavue:
Closes #1472


capricorn86 (approved):
Thank you for your contribution @ocavue! :star2: 

Sorry that you had to wait for such a long time.

Your fix worked, but I've updated the logic to not use "targetElement" and "currentElement" in `matchSelector()`, as it felt a bit confusing when I tried to understand the code.

Issue #1439, opened 2024-05-15
Input type checkbox indeterminate state not working


MarvinKubus:
**Describe the bug**
A checkbox with the indeterminate state being true should go to false when clicked, but this is not happening with happy-dom, it does with jsdom.

**To Reproduce**
Minimal example: https://stackblitz.com/edit/vitest-dev-vitest-rzc6un?file=test%2Fbasic.test.ts
Switch the happy-dom to jsdom there to see the test succeed.

**Expected behavior**
When clicking a checkbox element, I expect the indeterminate state to switch to false.

malko:
This is the JSDOM behavior but not the default behavior. You can try to make a minimal file example like this one:
```html
<!DOCTYPE html>
<html><body>
  <input type="checkbox" />
  <script type="text/javascript">
    document.querySelector('input').indeterminate = true
  </script>
</body></html>
```

And test it with chrome or firefox, with chrome it constantly  make it true when clicked, and with firefox reloading the page will make it false or true changing the value on each reload.
The indeterminate state can only be set by javascript and is purely visual. It is the checked state that should determinate the new state when clicked not the indeterminate state at all.

So if your input is checked and indeterminate it should when clicked be unchecked,
if it was unchecked and indeterminate it should be checked when clicked.

malko:
There's still a problem with the happy-dom implementation, as triggering a click event on the element will not change the indeterminate state at all, but is should be passed to false.

malko:
Also I can see the actual implementation sets an attribute indeterminate on the input element but it should not as indeterminate is not meant to be an input attribute.

malko:
will propose a PR about this in few days

malko:
PR on the way :)

malko:
Should be solved by merged #1475 @capricorn86 , @MarvinKubus

capricorn86:
Thank you again @malko! :star2:

PR #1475, opened 2024-07-02
fix: [#1439] HTMLInputElement's Indeterminate property doesn't behave properly

malko:
This PR will make indeterminate property behave more properly:
- It will not set indeterminate as an attribute but only as a property
- It will correctly sets indeterminate state to false when HTMLInputElement type="checkbox" is clicked


malko:
MDN related documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes

capricorn86 (approved):
Thank you for your contribution @malko! :star2: 

Sorry that you had to wait for such a long time. It has been a busy summer.

PR #1484, opened 2024-07-14
fix: throw error when append parameter type incorrect

btea:


capricorn86 (approved):

I made some additional fixes, so that the error class from the Window instance is used. Otherwise the BrowserExceptionObserver will not be able to identify where the error came from.

Issue #1494, opened 2024-07-23
innerHTML and outterHTML case that differs from browser behavior


weibangtuo:
**Code**

```js
test('mytest', () => {
  const div = document.createElement('div')

  div.innerText = '<b>a</b>'
  expect(div.innerHTML).toBe('&lt;b&gt;a&lt;/b&gt;')
  expect(div.outerHTML).toBe('<div>&lt;b&gt;a&lt;/b&gt;</div>')
})
```

**Expected behavior**
In a typical browser environment, this test case should pass. But it didn't.
```
AssertionError: expected '<b>a</b>' to be '&lt;b&gt;a&lt;/b&gt;' // Object.is equality

- Expected
+ Received

- &lt;b&gt;a&lt;/b&gt;
+ <b>a</b>
```

**Device**
`happy-dom` version: `14.12.3`

capricorn86:
Thank you for reporting @weibangtuo! :slightly_smiling_face: 

I believe this issue has been fixed now in v16.0.0.

PR #1498, opened 2024-07-26
fix: [#1494] Escape content in getInnerHTML

syi0808:
According to the https://html.spec.whatwg.org/#serialising-html-fragments, in innerHTML, textnode should be escaped like https://html.spec.whatwg.org/#escapingString 

The actual browser behavior of the modified existing test is also as follows.
스크린샷 2024-07-27 오전 12 47 56

Fix: #1494


capricorn86:
Thank you for your contributuion @syi0808! :star2: 

I believe this was fixed in [happy-dom@15.0.0](https://github.com/capricorn86/happy-dom/releases/tag/v15.0.0). I will therefore close the PR.

Issue #1414, opened 2024-04-22
Selector starting with new line throws error


rothsandro:
**Describe the bug**
If selector that starts with a new line, Happy DOM throws an error "Invalid selector".

```tsx
test("selector", () => {
  document.querySelector("\ninput");
});
```

**To Reproduce**
I forked the happy-dom repro and added a test in `QuerySelector.test.ts`.
Clone the repro, switch to the `bug/selector-with-new-line` branch and run the tests.

Repro: https://github.com/rothsandro/happy-dom/tree/bug/selector-with-new-line

Test: https://github.com/rothsandro/happy-dom/blob/bug/selector-with-new-line/packages/happy-dom/test/query-selector/QuerySelector.test.ts#L1090-L1098

```
it('Returns an element when selector starts with a new line', () => {
  const div = document.createElement('div');
  const div2 = document.createElement('div');

  div2.className = 'baz';
  div.appendChild(div2);

  expect(div.querySelector('\n.baz') === div2).toBe(true);
});
```

**Expected behavior**
The selector should work because it's valid and works in the browser.

**Device:**
 - OS: Windows 11

**Additional context**
A third-party component library that we use in our application uses such a selector. Because of that we cannot test parts of our app because it would result in this error.

capricorn86:
Thank you for reporting @rothsandro! :slightly_smiling_face: 

Big thanks to @syi0808 for providing with a fix!

Issue #1461, opened 2024-06-13
broken querySelectorAll when the string contains new lines


masylum:
**Describe the bug**
When the selector contains new lines, `querySelectorAll` returns all sorts of things that it shouldn't.

**To Reproduce**

```
document.querySelectorAll(
  `pre, code, script, style, object, embed,
    iframe, form, input, textarea, button,
    select, option, nav`
)
```

and then compare with

```
document.querySelectorAll(
  'pre, code, script, style, object, embed, iframe, form, input, textarea, button, select, option, nav'
)
```

**Expected behavior**
it should be the same

capricorn86:
Thank you for reporting @rothsandro! :slightly_smiling_face: 

There is a fix in now.

Big thanks to @syi0808 for providing with the fix!

PR #1499, opened 2024-07-27
fix: [#1414] Normalize selector parameter

syi0808:
To keep the browser behavior, I submit this PR. (However, [dom spec](https://drafts.csswg.org/css-syntax-3/#css-parse-something-according-to-a-css-grammar) doesn't seem to detail the implementation.)

Fixes: #1414 #1461


capricorn86 (approved):
Thank you for your contribution @syi0808! :star2: 

It seems like it should be supported to use new lines and tabs as white-space character (e.g. `.a\n.b`), so I added support for it as well.

PR #1500, opened 2024-08-04
feat: Support Blob.stream().

PlopAndRun:
https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream

Some libraries, like zip.js, use this method internally. Therefore it is impossible right now to test code that uses such libraries.


capricorn86 (approved):
Thank you for your contribution @PlopAndRun! :slightly_smiling_face:

PR #1523, opened 2024-08-31
chore: [1523] Add vitest extensions

btea:
Convenient and quick execution of debugging test cases.

image


capricorn86 (approved):

PR #1533, opened 2024-09-10
Fix incorrect handling of non-node items inserted into the DOM

BenjaminAster:
Happy-DOM's current handling of inserting items into the DOM that are not nodes via `.prepend()`, `.append()`, `.before()`, `.after()`, `.replaceWith()` or `.replaceChildren()` is incorrect and might even allow for XSS vulnerabilities:
If an item is a string, Happy-DOM parses the string as HTML and inserts the parsed HTML. However, the actual behavior should be to add text nodes with the string as the text content, which is then automatically sanitized.
```js
element.append("<script>");
console.log(element.innerHTML);
// Expected result: "&lt;script&gt;"
// Actual result: "<script></script>"
```
Additionally, if an item is neither a string nor a node, the browser converts it to a string automatically, but Happy-DOM errors.
```js
element.append(123, true, {});
console.log(element.textContent);
// Expected result: "123true[object Object]"
// Actual result: [throws an error]
```
Note that the extra `String()` wrapper before passing it to `document.createTextNode` is necessary due to the special behavior of `undefined` in the DOM methods vs. the `new Text()` constructor.
```js
element.append(new Text(undefined));
console.log(element.textContent);
// Expected result: "";
// BUT:
element.append(undefined);
console.log(element.textContent);
// Expected result: "undefined";
```
This PR (hopefully) fixes this to match the actual behavior of browsers. Let me know if I'm missing something and if I should add tests.

### DIFF
--- a/packages/happy-dom/src/nodes/child-node/ChildNodeUtility.ts
+++ b/packages/happy-dom/src/nodes/child-node/ChildNodeUtility.ts
@@ -1,8 +1,5 @@
 import DOMException from '../../exception/DOMException.js';
 import * as PropertySymbol from '../../PropertySymbol.js';
-import XMLParser from '../../xml-parser/XMLParser.js';
-import DocumentFragment from '../document-fragment/DocumentFragment.js';
-import Document from '../document/Document.js';
 import Node from '../node/Node.js';
 import IParentNode from '../parent-node/IParentNode.js';
 import IChildNode from './IChildNode.js';
@@ -36,15 +33,10 @@ export default class ChildNodeUtility {
 		}
 
 		for (const node of nodes) {
-			if (typeof node === 'string') {
-				const newChildNodes = (<DocumentFragment>(
-					XMLParser.parse(<Document>childNode[PropertySymbol.ownerDocument], node)
-				))[PropertySymbol.nodeArray];
-				while (newChildNodes.length) {
-					parent.insertBefore(newChildNodes[0], childNode);
-				}
-			} else {
+			if (node instanceof Node) {
 				parent.insertBefore(node, childNode);
+			} else {
+				parent.insertBefore(parent[PropertySymbol.ownerDocument].createTextNode(String(node)), childNode);
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


BenjaminAster:
Alright, so apparently Happy-DOM uses the `.replaceWith()` method for the `innerHTML` setter, which then expects the wrong unsanitizing behavior... :/ Plus, the failing tests also seem to imply that the `instanceof Node` check isn't working reliably, or something...?

Anyways, I'll look into this this probably tomorrow; don't merge this yet.

capricorn86:
Thank you for contributing @BenjaminAster! :slightly_smiling_face: 

Looking forward for the fix!

I believe that `instanceof Node` should be a reliable way of detecting if it is a `Node`. The tests are probably failing because the test the current behavior.

BenjaminAster:
@capricorn86 Ok, I have now modified the tests and the `outerHTML` setter, and it should be working now with the tests actually passing. GitHub still reports some integration-tests failing; I do not know what this means.

---

Btw, unrelated to this:
When I `git commit` on my machine, it reports the following errors:
```
2024-09-11 18:55:44.588 [info] husky - DEPRECATED

Please remove the following two lines from .husky/pre-commit:

#!/usr/bin/env sh
. "$(dirname -- "$0")/_/husky.sh"

They WILL FAIL in v10.0.0

C:\[MY_PERSONAL_PATH_TO]\happy-dom\node_modules\.bin\happy-lint-changed:2
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
          ^^^^^^^

SyntaxError: missing ) after argument list
    at wrapSafe (node:internal/modules/cjs/loader:1469:18)
    at Module._compile (node:internal/modules/cjs/loader:1491:20)
    at Module._extensions..js (node:internal/modules/cjs/loader:1691:10)
    at Module.load (node:internal/modules/cjs/loader:1317:32)
    at Module._load (node:internal/modules/cjs/loader:1127:12)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:217:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:166:5)
    at node:internal/main/run_main_module:30:49

Node.js v22.7.0
husky - pre-commit script failed (code 1)
```
, which I can circumvent by running `git commit` with `--no-verify`, but maybe this is something that should be inspected further? I have no idea about any of this, so I doubt I can help much here 😬.

capricorn86 (approved):
Thank you for your contribution @BenjaminAster! :star2: 

Not sure how this got implemented wrong, but great that it got fixed.

capricorn86:
The failing unit test was because of some hiccup. It loads real websites such as github.com and npmjs.com into Happy DOM, which fails randomly from time to time.

BenjaminAster:
Thank you! And no worries, it's easy to get confused with all the hundreds of DOM methods and what they do exactly, and this error probably stayed undetected for a while because it's very rare to insert text that contains HTML tags.

Issue #1534, opened 2024-09-11
`details` HTML element works differently in happy-dom compared to js-dom


mikedidomizio:
**Describe the bug**

I was working on something that uses the [details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element and I noticed that testing with react-testing-library behaved differently between one project of ours and another.

I checked versions of vitest (and other vitest dependencies), testing-library and although the stack trace didn't give me any insights, I think I've narrowed it down to happy-dom.

The details element when closed, is supposed to have had the `open` attribute stripped from it, because `open=""` and `open="false"` is actually open.

js-dom seems to work that way, but not happy-dom.

| [js-dom](https://testing-playground.com/#markup=DwEwlgbgfKCmAuBDMAbAzjNBXAtjxATgJ5Q5EAE2ehRwA9FfsVAPbwAWsB5ICy69XklQZBkKEA) after closed | [happy-dom](https://testing-playground.com/#markup=DwEwlgbgfKCmAuBDMAbAzgAgPYAdYDsBeAImJjQFcBbKxAJwE8oqGNKb6HgB6d2xqFngALWHQwgEydD0lJUaGN3DQgA) after closed |
|--------|--------|
| ![CleanShot 2024-09-11 at 08 45 53](https://github.com/user-attachments/assets/86d1d9e8-a05d-4939-a402-375660ebab87) | ![CleanShot 2024-09-11 at 08 45 40](https://github.com/user-attachments/assets/4226a265-dc80-4b58-9136-fa0a5fba88c9) | 

**To Reproduce**

1. I've created two CodeSandbox examples to show the difference (albeit in Next.js), 

- [js-dom](https://codesandbox.io/p/devbox/next-vitest-details-jsdom-mltyv9?file=%2Fapp%2F__tests__%2Fpage.test.tsx%3A35%2C5-35%2C38)

- [happy-dom](https://codesandbox.io/p/devbox/next-vitest-details-happy-dom-pvlsjv?file=%2Fapp%2F__tests__%2Fpage.test.tsx%3A35%2C38&workspaceId=266b9164-50d3-428a-95e5-b7e4618ca0b5)

2. Open the terminal window
3. Run `npm run test`

**Expected behavior**
- Both happy-dom and js-dom pass

**Screenshots**

https://github.com/user-attachments/assets/22cc3eb1-383d-42d4-b01c-d29d1105286a

**Device:**

[...]

PR #1536, opened 2024-09-12
fix: [#1534] Fix click event not removing open attribute

mikedidomizio:
Solves https://github.com/capricorn86/happy-dom/issues/1534

I'm not exactly sure this is the correct approach to take as I'm not familiar with the project setup, but maybe we can figure it out.

This change makes it so that a click event on the `<details>` element will trigger the removal or adding of the `open` attribute.

Before clicking on `<details>` would not remove the attribute or add it either.


capricorn86 (approved):

capricorn86:
Thank you for your contribution @mikedidomizio! :star: 

I have added some unit tests and added some additional rules according to spec.

PR #1539, opened 2024-09-16
fix: [#1538] Always return Promise<Blob> from ClipboardItem.getType() method

### LINKED ISSUE #1538: ClipboardItem.getType() should return `Promise<Blob>`, not `Promise<Blob | string>`
**Describe the bug**

According to [the spec](https://w3c.github.io/clipboard-apis/#dom-clipboarditem-gettype), `ClipboardItem.getType()` should return a blob, even when the item was created with a plain string. In particular, point 8.1.2.1.1 in section 7.2.3 of the spec details the expected behavior:

- `7.2.3.` getType(type)
  - `8.` For each representation in itemTypeList:
      - `1.` If representation’s [MIME type](https://w3c.github.io/clipboard-apis/#representation-mime-type) is mimeType and representation’s [isCustom](https://w3c.github.io/clipboard-apis/#representation-iscustom) is isCustom, then:        
          - `2.` [React](https://webidl.spec.whatwg.org/#dfn-perform-steps-once-promise-is-settled) to representationDataPromise:
              - `1.` If representationDataPromise was fulfilled with value v, then:
                  - **`1.` If v is a [`DOMString`](https://webidl.spec.whatwg.org/#idl-DOMString), then follow the below steps:**
                      - **`1.` Let dataAsBytes be the result of [UTF-8 encoding](https://encoding.spec.whatwg.org/#utf-8-encode) v.**
                      - **`2.` Let blobData be a [`Blob`](https://w3c.github.io/FileAPI/#dfn-Blob) created using dataAsBytes with its [`type`](https://w3c.github.io/FileAPI/#dfn-type) set to mimeType, [serialized](https://mimesniff.spec.whatwg.org/#serialize-a-mime-type).**
                      - **`3.` Resolve p with blobData.**
                  - `2.` If v is a [`Blob`](https://w3c.github.io/FileAPI/#dfn-Blob), then follow the below steps:
                      - `1.` Resolve p with v.

I believe this bug is also causing unexpected clipboard behavior when using `happy-dom` with `@testing-library/user-event`, which expects the return value from `ClipboardItem.getType()` to always be a `Promise<Blob>`.

**To Reproduce**

```js
const i = new ClipboardItem({ "text/plain": "test" });
const b = await i.getType("text/plain");
console.log(b instanceof Blob); // should be true
```

**Expected behavior**

`ClipboardItem.getType()` should always return a `Promise<Blob>`, even if it means converting a string to a blob.

**Screenshots**

N/A

**Device:**

 - OS: N/A
 - Browser: N/A
 - Version: 15.7.4

**Additional context**

N/A

ezzatron:
This PR changes `ClipboardItem.getType()` to always return a `Promise<Blob>`, even when the item was created with a plain string. This is intended to fix #1538.


marchaos:
Any chance we can get this merged? Looks like there's a failing build, but not sure it's related to the changes here.

packages/happy-dom/src/clipboard/ClipboardItem.ts
marchaos, line 1: You need to import Blob here: 

```
import Blob from '../file/Blob.js';
```

See https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/clipboard/Clipboard.ts

Otherwise this uses Blob in the global scope, which is node when I run tests.

capricorn86 (approved):

capricorn86:
@ezzatron Thank you for contributing! :star: 

PR #1546, opened 2024-09-23
Use `globalThis` instead of `global`

mattallty:
`global` is a legacy, node.js-only global variable. 

I think `globalThis` should be used instead, which will make global-registrator portable on other runtimes like Cloudflare workers.


capricorn86 (approved):

capricorn86:
Thank you for your contribution @mattallty! :star: 

PR #1555, opened 2024-09-26
fix: [jest-environment] Allow custom export conditions

aearly:
When switching to use `happy-dom` from `jsdom` in my project's Jest setup, I noticed that the export conditions stopped working. It looks like the jest-environment needs to replicate [this logic](https://github.com/jestjs/jest/blob/main/packages/jest-environment-jsdom-abstract/src/index.ts#L125-L137) from the abstract environment to respect any `customExportConditions` settings.

- Also fixed some warnings in the tests.


capricorn86 (approved):
Thank you for your contribution @aearly! :star: 

Great that you fixed the warnings as well!

As changing the default value for `customExportConditions` would be a breaking change, we will have to wait with it until there is a major release.  The tests for Vue failed as it used the browser version instead of the Node version for example.

I have changed the default value to the same as before in a commit.

Issue #1557, opened 2024-09-28
TypeError: Cannot set property srcObject of [object Object] which has only a getter


zeid0ne:
**Describe the bug**
I cant set the srcObject props of a dom element "generated" by happi-dom

**To Reproduce**
https://stackblitz.com/edit/vitest-dev-vitest-ve6adl

**Expected behavior**
I can set srcObject to a stream or anything else

**Device:**
System:
    OS: Windows 11 10.0.22631
    CPU: (16) x64 AMD Ryzen 7 7745HX with Radeon Graphics
    Memory: 15.30 GB / 31.69 GB
 Browsers:
    Edge: Chromium (125.0.2535.92)
    Internet Explorer: 11.0.22621.3527
Happy-dom: 15.7.4

**Additional context**
I am using vitest. Thats working well with jsdom.

zeid0ne:
can i close this ?

PR #1560, opened 2024-10-06
fix: [#1557] TypeError: Cannot set property srcObject had only a getter

ewesemann:
Changed the srcObject of the HTMLMediaElement to include a setter. See issue [#1557].  I came across this issue as well and thought it shouldn't be to bad to fix.


ewesemann:
Looks like the failing unit test does so randomly when tested on my machine.

capricorn86 (approved):
Thank you for your contribution @ewesemann! :star: 

I have added validation for if the object to be set is of type `MediaStream` as a commit.

capricorn86:
> Looks like the failing unit test does so randomly when tested on my machine.

There are some flaky tests that are based on timers. I'm trying to create a list of them and will hopefully fix it soon.

Issue #1568, opened 2024-10-21
Location.assign and Location.replace should accept `URL` object for `url`


silverwind:
Browsers convert the argument to these to string, so happy-dom should do too.

https://developer.mozilla.org/en-US/docs/Web/API/Location/assign
https://developer.mozilla.org/en-US/docs/Web/API/Location/replace

https://github.com/microsoft/TypeScript/blob/e6ef279403d86440600c866d53839a3e695220d3/src/lib/dom.generated.d.ts#L15107
https://github.com/microsoft/TypeScript/blob/e6ef279403d86440600c866d53839a3e695220d3/src/lib/dom.generated.d.ts#L15119

PR #1569, opened 2024-10-21
fix: [#1568] Accept stringable objects in Location.assign and Location.replace

silverwind:
Fixes: https://github.com/capricorn86/happy-dom/issues/1568
Related: https://github.com/mdn/content/pull/36432

As Typescript does not have a suitable type for stringable things, I think having `URL` is the best we can do and it matches the Typescript DOM types as well.


capricorn86 (approved):
Thank you for your contribution @silverwind! :star: 

I agree that `URL` is probably the best we can do, specially if the Typescript DOM types uses it.

PR #1574, opened 2024-11-02
chore: [husky] The current husky version no longer supports the install command

btea:


capricorn86 (approved):

PR #1575, opened 2024-11-02
fix: [#1565] Strictly check the type of `blob` parameters

### LINKED ISSUE #1565: FileReader never throws errors
**Describe the bug**
```
const reader = new FileReader();
reader.readAsDataURL(file);

reader.onload = function () {
  resolve(reader.result);
  };
reader.onerror = function (error) {
  reject(error); // never goes here
};
```

The following test failed:
```
it('should reject with an error when passed an invalid file', async () => {
    const invalidFile = {};
    await expect(fileToBase64(invalidFile)).rejects.toThrowError();
  });
  
  - Expected: 
[Error: rejected promise]

+ Received: 
"data:application/octet-stream;base64,"
```

**Expected behavior**
To throw an error. For example, Mozilla throws the error:
Uncaught TypeError: FileReader.readAsDataURL: Argument 1 does not implement interface Blob.

**Screenshots**

**Device:**
 - OS: MacOS 14.6.1
 - Browser Mozilla
 - Version 15.7.4

**Additional context**
Vitest tests.
The test provided passes on `jsdom` environment

btea:
fix #1565


capricorn86 (dismissed):

I changed to use TypeError from Window, as BrowserExceptionObserver can then detect the origin of the error

btea:
Thanks! 👍

PR #1595, opened 2024-11-09
chore: [husky] Update vite

btea:
The current vitest version does not match the @vitest/ui version


capricorn86 (approved):

PR #1597, opened 2024-11-10
fix: [#1586] Prevent always wrapping the URL in quotation marks when set in style

iradonov:


capricorn86 (changes requested):
Thank you for contributing @iradonov! :star: 

According to spec, quotation marks should always be added.

If you run the following code in the browser, it will behave the same:
```javascript
var el = document.createElement('div');
el.style = 'background-image: url(test.jpg)'
// Outputs url("test.jpg")
console.log(el.style.backgroundImage)
```

Issue #1611, opened 2024-11-13
Support TextEncoder & TextDecoder


Ayc0:
**Is your feature request related to a problem? Please describe.**

On the global scope, `TextEncoder` & `TextDecoder` are available (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/TextEncoder)).

But when I run this code:

```js
const h = require('happy-dom');
const w = new h.Window();
w.TextEncoder // undefined
```

This poses an issue for tests running in Jest with `@happy-dom/jest-environment`, testing code using `TextEncoder`

**Describe the solution you'd like**
Implement those 2 globals by exposing those from `node:util`

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1612, opened 2024-11-13
fix: [#1611] Implement window.TextEncoder & window.TextDecoder

Ayc0:
Fixes https://github.com/capricorn86/happy-dom/issues/1611


capricorn86 (approved):

Issue #1615, opened 2024-11-18
DOMParser does not recognise BODY and thus creates it twice


OlaviSau:
**Describe the bug**
DOMParser is not compliant with the browser implementation. This breaks frameworks like Angular from setting the innerHTML as an attribute, breaking most renders using Angular.
This occurs because BODY is not recognised and this causes body to be created as an extra element in HappyDOM.

**To Reproduce**
`(new window.DOMParser()).parseFromString("<body><x></x>Example Text", "text/html").body.innerHTML`
HappyDOM: `<body><x></x>Example Text</body>`
**Expected behavior**
Chrome: `<x></x>Example Text`

**Additional context**
tagName BODY should not be created as a standard node.
https://github.com/capricorn86/happy-dom/blob/afd256b2e4f0260adb22432c1a354f558cda6623/packages/happy-dom/src/dom-parser/DOMParser.ts#L79
https://github.com/capricorn86/happy-dom/blob/afd256b2e4f0260adb22432c1a354f558cda6623/packages/happy-dom/src/dom-parser/DOMParser.ts#L87

@capricorn86

OlaviSau:
@capricorn86 I have created a naive fix for this, please take a look and make improvements.
https://github.com/capricorn86/happy-dom/pull/1616

aralroca:
Now I receive this error:

```js

21 |     document.body.innerHTML = '';
                  ^
TypeError: null is not an object (evaluating 'document.body.innerHTML = ""')
```

Is that normal? `document.body.innerHTML = ""` is allowed in browsers

capricorn86:
> Now I receive this error:
> 
> ```js
> 21 |     document.body.innerHTML = '';
>                   ^
> TypeError: null is not an object (evaluating 'document.body.innerHTML = ""')
> ```
> 
> Is that normal? `document.body.innerHTML = ""` is allowed in browsers

"body" should only become null if you have removed the body element. When using `document.write()` or `DOMParser` it should always be populated with an `HTMLBodyElement`.

How did you end up with "body" becoming null?

aralroca:
@capricorn86 I'm using in this case on `afterEach`:

https://github.com/brisa-build/brisa/blob/66249054e2223b7965c16126ec517ce71d6ea3b8/packages/brisa/src/utils/rpc/resolve-rpc/index.test.ts#L21

But if I comment this line, I got similar error on other lines like this one:

[...]

PR #1617, opened 2024-11-19
fix: [#1615] Fixes problems related to parsing <html>, <head> and <body> using DOMParser

capricorn86:


packages/happy-dom/test/dom-parser/DOMParser.test.ts
```diff
@@ -98,11 +98,19 @@ describe('DOMParser', () => {
...
 		});
 
-		it('parses SVGs', () => {
+		it('Parses SVGs', () => {
 			const newDocument = domParser.parseFromString(DOMParserSVG, 'image/svg+xml');
 			expect(new XMLSerializer().serializeToString(newDocument).replace(/[\s]/gm, '')).toBe(
 				DOMParserSVG.replace(/[\s]/gm, '')
 			);
 		});
+
```
OlaviSau, line 107: If you're planning on handling all the cases - then i suggest these as well:

```
export const DOMParserBODYSibling = '<body><example></example>Example Text</body><body></body>';
export const DOMParserBODYChildren = '<body><body></body><example></example>Example Text</body>';
		it('recognises does not duplicate body when BODY is a child', () => {
			const newDocument = domParser.parseFromString(DOMParserBODYChildren, 'text/html');
			expect(newDocument.body.innerHTML).toBe('<example></example>Example Text');
		});

		it('recognises does not duplicate body when BODY is a sibling', () => {
			const newDocument = domParser.parseFromString(DOMParserBODYSibling, 'text/html');
			expect(newDocument.body.innerHTML).toBe('<example></example>Example Text');
		});
```

packages/happy-dom/src/config/HTMLElementConfig.ts
```diff
@@ -128,11 +138,13 @@ export default <
 	},
 	col: {
 		className: 'HTMLTableColElement',
-		contentModel: HTMLElementConfigContentModelEnum.noDescendants
+		contentModel: HTMLElementConfigContentModelEnum.noDescendants,
+		permittedParents: ['colgroup']
 	},
 	colgroup: {
 		className: 'HTMLTableColElement',
```
OlaviSau, line 145: The class name is the same as the previous?

packages/happy-dom/src/html-serializer/HTMLSerializer.ts
```diff
@@ -0,0 +1,162 @@
...
+		switch (root[PropertySymbol.nodeType]) {
+			case NodeTypeEnum.elementNode:
+				const element = <Element>root;
+				const prefix = element[PropertySymbol.prefix];
+				const localName = element[PropertySymbol.localName];
+				const config = HTMLElementConfig[element[PropertySymbol.localName]];
+				const tagName = prefix ? `${prefix}:${localName}` : localName;
+
+				if (config?.contentModel === HTMLElementConfigContentModelEnum.noDescendants) {
+					return `<${tagName}${this.getAttributes(element)}>`;
```
OlaviSau, line 72: I noticed that for AREA element it doesn't delete the element, it just places the element after it. Test case would be:
```
const parser = new DOMParser();
expect(
  parser.parseFromString("<area><span></span></area>", "text/html").body.innerHTML
).toBe("<area><span></span>")
```
For the col element it does something different, instead of placing it AFTER, it places it before.
Test case:
```
const parser = new DOMParser();
expect(
  parser.parseFromString("<table><col><span></span></col></table>", "text/html").body.innerHTML
).toBe("<span></span><table><colgroup><col></colgroup></table>")
```
It seems you need descendantsAfter / descendantsBefore instead of noDescendants.

packages/happy-dom/test/dom-parser/DOMParser.test.ts
capricorn86: I have added the test cases, thank you! :slightly_smiling_face: 

This turned out to be a lot more complicated than what I first thought and I had to refactor quite a bit of the logic, but I have hopefully covered all cases reported and discovered when reading the spec and tested in the browser.

packages/happy-dom/src/config/HTMLElementConfig.ts
capricorn86: This is correct.

If you run the following code in the browser you will see that both are using the `HTMLTableColElement` class:
```javascript
// Outputs "HTMLTableColElement"
console.log(document.createElement('col').constructor.name);

// Outputs "HTMLTableColElement"
console.log(document.createElement('colgroup').constructor.name);
```

packages/happy-dom/test/dom-parser/DOMParser.test.ts
OlaviSau: It's awesome though that you managed to get this through - impressive work :)

capricorn86: Thank you @OlaviSau!

Issue #1551, opened 2024-09-24
getElementsByClassName returns wrong results when given multiple classes


hombrew:
**To Reproduce**
```javascript
import { Window } from "happy-dom";

const window = new Window();
const document = window.document;

document.body.innerHTML = `
<div class="class1 class2">
  <div class="class2">
    <span>hi</span>
  </div>
</div>`;

const arr1 = Array.from(document.getElementsByClassName("class1 class2"));
console.log(arr1.length); // output: 0 - wrong
```

**Expected behavior**
in this case getElementsByClassName should return 1 element as shown [here](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName)

**Device:**
 - OS: OSX Sonoma 14.6.1
 - Browser Chrome
 - Version 129.0.6668.58

**Additional context**
based on the same code, comma separated classes work as expected:

```javascript
const arr2 = Array.from(document.getElementsByClassName("class1, class2"));
console.log(arr2.length); // output: 1 - ok
```

capricorn86:
Thank you for reporting @hombrew! :slightly_smiling_face: 

Big thanks to @kylehalleman for contributing with a fix :star: 

The fix has been released in v16.1.0.

PR #1618, opened 2024-11-20
fix: Allow getElementsByClassName to accept a whitespace-separated list

kylehalleman:
`getElementsByClassName` can accept a whitespace-separated list of classNames (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName#names) and [spec](https://dom.spec.whatwg.org/#example-5ffcda00)).

The current implementation prepends `.` to the className in `querySelectorAll` (i.e., `classOne classTwo    classThree` -> `.classOne classTwo    classThree`). This PR splits the className string by any amount of whitespace, then prepends a `.` to each separated className, then joins to make a single query to `querySelectorAll` (e.g., `classOne classTwo    classThree` -> `.classOne.classTwo.classThree`).

Added test to confirm this new functionality.


personalyisus:
This would be a solution for #1551

capricorn86 (approved):
Thank you for your contribution @kylehalleman! :star2: 

I have changed the solution a bit to improve performance by only using a `replace()`.

PR #1624, opened 2024-11-25
feat: [1623] Add missing scrollBy functionality

OlaviSau:
Just filling in the method :)


OlaviSau:
@capricorn86 should hopefully be a simple one. LMK if you want improvements :)

capricorn86 (dismissed):
Thank you for your contribution @OlaviSau! :star: 

Sorry that it took some time to get this in.

I have added some unit tests and also found that "x" and "y" where reversed when not providing an object as argument.

capricorn86 (approved):
I also noticed that scrollBy() was missing on Window as well, so I've also added it there now

PR #1632, opened 2024-12-04
fix: [#1631] Set fetch credentials to same-origin for xhr when withCredentials is false

### LINKED ISSUE #1631: XMLHttpRequest doesn't send Authorization header when withCredentials is false even if the request is same origin
**Describe the bug**
XMLHttpRequest doesn't send Authorization header when withCredentials is false even if the request is same origin

**To Reproduce**
1. set url to some url `http://localhost:8123`
2. call an xmlhttprequest to the same origin with `withCredentials: false` and `Authorization` header
```
const req = new XMLHttpRequest();
req.addEventListener("load", reqListener);
req.open("GET", "http://localhost:8123/);
req.setRequestHeader("Authorization", "Basic asdf");
req.send();
```
3. `Authorization` header get's erased

**Expected behavior**
`Authorization` header is sent as per spec https://xhr.spec.whatwg.org/#the-withcredentials-attribute

**Additional context**

saitonakamura:
To more accurately emulate xhr spec behaviour https://xhr.spec.whatwg.org/#the-withcredentials-attribute

It looks like `credentials` should be set to `same-origin` when using xhr with `withCredentials` false, because xhr will only stop sending credentials if the request is not the same origin, and `send` method of fetch will call `getRequestHeaders` which performs this kind of check. I might be missing something, some kind of behaviour that will break with `same-origin` and was working correctly with `omit`, but I struggle to find it.

I'm not sure if my approach to tests is good enough. I tried to do this via unit test first, but every test there mocks the `send` implementation (and the `send` method calls `getRequestHeaders`) so I decided that I'd rather write an integration test and a unit test will check what kind of `credentials` option is passed to fetch. So let me know if you have better idea of how to refactor the tests if it's needed

Fixes #1631


capricorn86 (approved):
Thank you for your contribution @saitonakamura! :star: 

The tests looks good and seem to cover all scenarios.

PR #1639, opened 2024-12-18
Update NodeUtility.ts to fix: TypeError: Cannot read properties of undefined (reading 'length')

jackytank:
image


capricorn86 (changes requested):
Thank you for your contribution @jackytank! :star: 

I believe that the root cause for the problem is that the value sent in is not of type node, and therefore the property `[PropertySymbol.nodeArray]` is not available on the object.

In that case, the correct fix is to throw an exception when the value is not of type 'Node'.

We also need unit tests for it.

I can perhaps look into it more later tonight.

PR #1640, opened 2024-12-18
fix: Remove redundant querySelector type check

chocolateboy:
Remove a type check which is covered by (and shadows) subsequent checks.

### DIFF
--- a/packages/happy-dom/src/query-selector/QuerySelector.ts
+++ b/packages/happy-dom/src/query-selector/QuerySelector.ts
@@ -92,7 +92,9 @@ export default class QuerySelector {
 		}
 
 		if (typeof selector === 'symbol') {
-			throw new window.TypeError(`Cannot convert a Symbol value to a string`);
+			throw new window.TypeError(
+				`Failed to execute 'querySelectorAll' on '${node.constructor.name}': Cannot convert a Symbol value to a string`
+			);
 		}
 
 		selector = String(selector);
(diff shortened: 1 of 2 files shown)


btea (approved):

packages/happy-dom/src/query-selector/QuerySelector.ts
```diff
@@ -205,12 +205,6 @@ export default class QuerySelector {
 			);
 		}
```
capricorn86, line 207: We should cover the case if selector is a Symbol.

The correct error for Symbol should be:
`TypeError: Failed to execute 'querySelector' on 'Document': Cannot convert a Symbol value to a string`

```suggestion
		if (typeof selector === 'symbol') {
			throw new window.TypeError(
				`Failed to execute 'querySelector' on 'Document': Cannot convert a Symbol value to a string`
			);
		}
```

capricorn86 (changes requested):
Thank you for your contribution @chocolateboy! :star: 

We should cover the case if selector is a Symbol.

packages/happy-dom/src/query-selector/QuerySelector.ts
chocolateboy: That check is already included below:

https://github.com/capricorn86/happy-dom/blob/7a5aa6249e7b1fb61aae6ecb64e2ce9384b7e66b/packages/happy-dom/src/query-selector/QuerySelector.ts#L214-L216

chocolateboy: I've updated that error message to include the node type, as per the other checks.

capricorn86 (approved):
Looks good :slightly_smiling_face:

Issue #1502, opened 2024-08-07
Intercept network resource requests / defining resource loader


SebJansen:
**Is your feature request related to a problem? Please describe.**
JSDOM has the ability to define custom resource loaders, to avoid the need for network requests, by going through the filesystem instead. This saves precious milliseconds. 

JSDOM resource loaders also have access to the requesting HTML element; eg if a style element is requesting a CSS file, the loader can get it from the filesystem instead and then replace the <link rel="stylesheet"> element with a <style> element with the CSS inlined to prevent the client needing to make an additional network request for the CSS file upon first page being served.

**Describe the solution you'd like**
Some way to intercept network requests and instead serve them from the filesystem. Perhaps this is already possible with HappyDOM but merely undocumented. 

**Describe alternatives you've considered**
Overriding window.fetch requires some properties like browser frame, which are not exposed during eg new Window(). Perhaps with a little guidance it is possible to document a way to override fetch.

mariusrak:
Hi, i also very much need this. I do requests only inside my own server and do not need to go through network. So this intercepting fetch request would greatly increase performance and security.

OlaviSau:
Hi :) We also would greatly appreciate this, again currently using patch-package to get around this.

OlaviSau:
https://github.com/capricorn86/happy-dom/pull/1556 Just an example of something that could work.

OlaviSau:
@SebJansen I discussed it with a colleague and although file urls would work, the caching would be a bit tricky as files can change at any moment. It is probably wiser to implement a full interceptor instead.

davidortner-ipex:
Hi! :wave: 
I agree that something is needed, but I'm not sure what the best solution would be yet :sweat_smile:  Another feature I would like is to make it possible to simulate a local web-server that is served from a "public" directory. It should also be easy to add transformers for resources loaded.

If you feel that this should be prioritized, I recommend adding a :+1: to the ticket. There are a lot of feature requests to sort between and the thumb indicate that several has voted for it.

capricorn86:
Hi! :wave: 
I agree that something is needed, but I'm not sure what the best solution would be yet :sweat_smile:  Another feature I would like is to make it possible to simulate a local web-server that is served from a "public" directory. It should also be easy to add transformers for resources loaded.

If you feel that this should be prioritized, I recommend adding a :+1: to the ticket. There are a lot of feature requests to sort between and the thumb indicate that several has voted for it.

[...]

PR #1662, opened 2025-01-03
feat: [#1502] Allow fetch to be intercepted and modified

OlaviSau:
A proposal for implementation.
The intercept option is structured a bit deep, but it is for a reason. Since there exists a SyncFetch as well I could not come up with a good name that would clearly imply that the interceptor only applies for async requests and not for sync requests.
I considered the following options:
`{ fetch: { beforeSend: ... } }`
This is confusing as there is SyncFetch and Fetch. The types wouldn't match either for the response object. In addition if the options were to be a class, the method would be lost during setup in quite a confusing way.
`{ fetch: { beforeAsync: ... } }` 
This doesn't clarify before what and has the issue of losing the methods.
`{ fetch: { asyncInterceptor: { beforeSend: ... } } }` 
The use might assume that the interceptor is allowed to be async and applies both to SyncFetch and Fetch. Even if we change it to asyncFetchInterceptor it retains that problem.

Methods can be lost during spreading in BrowserSettingsFactory.createSettings(options?.settings) as methods from classes are not enumerable by default.

I also considered whether we could use the interceptor for setting the options instead of using the declarative way - it would allow for dynamically setting some options on fetch like disableCache, redirectCount or disableSameOriginPolicy. The downside is that it would require exposing the Fetch type to some extent. I'll leave that decision to @capricorn86.
The window is exposed due to the need for window for creating the response object as the response object cannot be created independently of the window. That would cause problems when creating the interceptor before the Window, which should be the default use case. It might also be useful if the interceptor is used for multiple windows.

If this approach is suitable - I can add a hooks for afterResponse, afterError and other cases that the users would like.


capricorn86:
I think your approach is good in general @OlaviSau.

I agree that there are many approaches with their pros and cons and it is not easy to say which one is the best :sweat_smile:

After thinking about it, my suggestion is:

_IFetchInterceptor.ts_
```typescript
export default interface IFetchInterceptor {
   beforeRequest?: (window: BrowserWindow, request: Request) => Promise<Response | void>;
   beforeSyncRequest?: (window: BrowserWindow, request: Request) => ISyncResponse | void;
}
```

_IBrowserSettings.ts_
```typescript
export default interface IBrowserSettings {
   fetch: {
      interceptor?: IFetchInterceptor
   };
}
```

I'm not sure "beforeRequest" is better than "beforeSend", but looking at some other libs it seems like it is common to use "beforeRequest" and "afterResponse".

I think it will make the code easier to understand and maintain if all interceptor logic is within the IFetchInterceptor interface (both for async and sync). I'm thinking that async is the default fetch behavior, so it might not need to be prefixed with "Async".

OlaviSau:
Thanks for the response ❤️ 
I think beforeRequest might be better if it's more common, now that I think about it - beforeRequest is probably more intuitive than beforeSend thinking from the outside context rather than inside the fetch class.

In terms of remove the the asyncFetch - I think that also makes sense.
There are 2 other changes you're suggesting though that I would like to get your reasoning behind window before the request. I was thinking that the window will primarily be useful when the user wants to return a response, but since that's not a guaranteed case - the request would be first.
The cases I am thinking about.
1. The user wants to do logging. Potentially require any number of parameters with window / request / response being unknown.
2. The user wants to change the request. Requires request.
3. The user wants to provide a response before the request is made. Requires request and window.
4. The user wants to change the response. Requires response and most likely the request, but not always (header addition).
5. The user wants to replace the response based on the response. Requires response, request and window.

It seems that it varies quite a bit what the user needs, perhaps I should use a context object instead?

I thought a bit about the extra layer and now I actually think it's fine that they are all in the same interceptor as the user can use composition on their class if they want to split functionality / handle data persistence in a shared manner. That allows me to remove one layer from the settings so I am happy about that :)

OlaviSau:
I added sync request for now as well. Will add response hooks soon.

capricorn86:
> Thanks for the response ❤️ I think beforeRequest might be better if it's more common, now that I think about it - beforeRequest is probably more intuitive than beforeSend thinking from the outside context rather than inside the fetch class.
> 
> In terms of remove the the asyncFetch - I think that also makes sense. There are 2 other changes you're suggesting though that I would like to get your reasoning behind window before the request. I was thinking that the window will primarily be useful when the user wants to return a response, but since that's not a guaranteed case - the request would be first. The cases I am thinking about.
> 
> 1. The user wants to do logging. Potentially require any number of parameters with window / request / response being unknown.
> 2. The user wants to change the request. Requires request.
> 3. The user wants to provide a response before the request is made. Requires request and window.
> 4. The user wants to change the response. Requires response and most likely the request, but not always (header addition).
> 5. The user wants to replace the response based on the response. Requires response, request and window.
> 
> It seems that it varies quite a bit what the user needs, perhaps I should use a context object instead?
> 
> I thought a bit about the extra layer and now I actually think it's fine that they are all in the same interceptor as the user can use composition on their class if they want to split functionality / handle data persistence in a shared manner. That allows me to remove one layer from the settings so I am happy about that :)

Glad to hear that you agree :smile: 

I get your point. I know that many libraries could send the parameters in that order.

I see it as the injection pattern where the context is the most important parameter that is always sent, and request is the second as it is within the context. This way we know that the first parameter will always be the same, instead of not knowing where the window parameter is.

Example:
beforeRequest?: (request: Request,  __window: BrowserWindow__, options?: Object) => Promise<Response | void>;
afterResponse?: (request: Request, response: Response, __window: BrowserWindow__, options?: Object => Promise<Response | void>;
beforeError?: (error: Error, request: Request, response: Response | null, __window: BrowserWindow__, options?: Object => Promise<Response | void>;

This is how most logic works for other Happy DOM functionality as much logic needs to know which Window context it is part of, specially when dealing with multiple Window instances running at the same time. The consumer will most likely need to know which Window at some point. It could for example be for checking the origin of the request as it can potentially be an iframe or another page.

I hope I explained my thinking a little bit :sweat_smile:

packages/happy-dom/src/fetch/Fetch.ts
```diff
@@ -39,7 +40,7 @@ const LAST_CHUNK = Buffer.from('0\r\n\r\n');
  */
 export default class Fetch {
 	private reject: (reason: Error) => void | null = null;
-	private resolve: (value: Response | Promise<Response>) => void | null = null;
+	private resolve: (value: Response | Promise<Response>) => Promise<void> = null;
```
OlaviSau, line 43: Could this cause any issues?

OlaviSau:
@capricorn86 I have added the response methods as well. I am noticing that there is quite a bit of duplication in the tests for the Fetch / SyncFetch now. Should I try to eliminate some of it or is it okay?

OlaviSau:
One other thing I was thinking about was perhaps splitting the afterResponse into two - beforeCache and afterCache. The naming would be tricky though so I just decided to leave it with the current plan.

packages/happy-dom/src/fetch/Fetch.ts
```diff
@@ -383,7 +402,14 @@ export default class Fetch {
 						});
 				}
 				this.#browserFrame[PropertySymbol.asyncTaskManager].endTask(taskID);
```
capricorn86, line 404: We need to end the task for the "asyncTaskManager" after the intercepted response has been handled

packages/happy-dom/src/fetch/Fetch.ts
capricorn86: It is only used internally, so it should be fine

capricorn86 (changes requested):
Looks good! :slightly_smiling_face:  Just a small finding

packages/happy-dom/src/fetch/Fetch.ts
OlaviSau: Fixed it :)

OlaviSau:
> > @capricorn86 I have added the response methods as well. I am noticing that there is quite a bit of duplication in the tests for the Fetch / SyncFetch now. Should I try to eliminate some of it or is it okay?
> 
> As long as the code is still easy to read and understand after removing any duplication you can try to eliminate some of it. To me it is better with some duplication and more code if it makes it easier to understand it.

I thought about it a bit and I could replace the mockModule part with a function that would return a promise that then returns the requestArgs. If possible I would like to do this in a secondary PR for easier review.

This is what I had in mind.
```
	async function mockNetwork(
		schema: 'http' | 'https',
		responseText: string,
		extra: Record<string, string>
	): Promise<{ url: string; options: HTTP.RequestOptions }> {
		return new Promise((resolve) => {
			mockModule(schema, {
				request: (url: string, options: HTTP.RequestOptions) => {
					return {
						end: () => {},
						on: (event: string, callback: (response: HTTP.IncomingMessage) => void) => {
							async function* generate(): AsyncGenerator<string> {
								yield responseText;
							}

							const response = <HTTP.IncomingMessage>Stream.Readable.from(generate());
							Object.assign(response, extra);
							if (event === 'response') {
								callback(response);
							}
							resolve({ url, options });
						},
						setTimeout: () => {}
					};
				}
			});
		});
	}
```

capricorn86:
> > > @capricorn86 I have added the response methods as well. I am noticing that there is quite a bit of duplication in the tests for the Fetch / SyncFetch now. Should I try to eliminate some of it or is it okay?
> > 
> > 
> > As long as the code is still easy to read and understand after removing any duplication you can try to eliminate some of it. To me it is better with some duplication and more code if it makes it easier to understand it.
> 
> I thought about it a bit and I could replace the mockModule part with a function that would return a promise that then returns the requestArgs. If possible I would like to do this in a secondary PR for easier review.
> 
> This is what I had in mind.
> 
> ```
> 	async function mockNetwork(
> 		schema: 'http' | 'https',
> 		responseText: string,
> 		extra: Record<string, unknown>
> 	): Promise<{ url: string; options: HTTP.RequestOptions }> {
> 		return new Promise((resolve) => {
> 			mockModule(schema, {
> 				request: (url: string, options: HTTP.RequestOptions) => {
> 					return {
> 						end: () => {},
> 						on: (event: string, callback: (response: HTTP.IncomingMessage) => void) => {
> 							async function* generate(): AsyncGenerator<string> {
> 								yield responseText;
> 							}
> 
> 							const response = <HTTP.IncomingMessage>Stream.Readable.from(generate());
> 							Object.assign(response, extra);
> 							if (event === 'response') {
> 								callback(response);
> 							}
> 							resolve({ url, options });
> 						},
> 						setTimeout: () => {}
> 					};
> 				}
> 			});
> 		});
> 	}
> ```
> 
> The consumption would look something like
> 
> ```
> const requestArgsPromise = mockNetwork(
> 	'http',
> 	'some text',
> 	{
> 		statusCode: 200,
> 		statusMessage: 'OK',
> 		headers: {},
> 		rawHeaders: [
> 			'content-type',
> 			'text/html',
> 			'content-length',
> 			String(responseText.length)
> 		]
> 	}
> );
> // fetch
> const requestArgs = await requestArgsPromise;
> ```
> 
> I tested it with some tests, seems that the GH workflow doesn't always run all tests, so I caught a bug I added at some point by using the interceptor without optional access.

Sounds good! :slightly_smiling_face:

capricorn86 (approved):

Issue #1629, opened 2024-11-29
Support for DOMPurify


capricorn86:
We should look into what tests that are failing in DOMPurify to get Happy DOM working there:
https://github.com/cure53/DOMPurify/pull/878

Related to:
https://github.com/capricorn86/happy-dom/issues/1403

StephenHaney:
Just for visibility, we ran into this today. Our SVG parsing uses `DOMPurify` and our test setup was using `happy-dom`. If appears if there is a `<style>` tag, the result of DOMPurify chops off all other elements after it when running in `happy-dom`. I couldn't find another workaround other than swapping to `jsdom` for the test runner. Hope it's a helpful report!

dubzzz:
Sorry for the ping on this issue, but I wanted to know what was the status of it. It seems it has been opened for quite some time. I wanted to know if a fix was planned or not or if the idea of fixing was not anymore planned. Thanks in advance for your answer, positive or not.

PR #1666, opened 2025-01-07
fix: [#1629] Multiple fixes to DOM parsing

capricorn86:


packages/happy-dom/test/xml-parser/XMLParser.test.ts
```diff
@@ -187,8 +187,8 @@ describe('XMLParser', () => {
 
 			expect((<HTMLElement>result.children[0].children[0]).textContent).toBe('');
 			expect((<HTMLElement>result.children[0].children[1]).textContent).toBe('');
-			expect((<HTMLElement>result.children[0].children[0]).innerHTML).toBe('<b></b>');
-			expect((<HTMLElement>result.children[0].children[1]).innerHTML).toBe('<b></b>');
+			expect((<HTMLElement>result.children[0].children[0]).innerHTML).toBe('<b/>');
```
OlaviSau, line 190: Is this correct?

capricorn86: It is correct as we are within an `XMLDocument`:

```javascript
var domParser = new DOMParser();
var result = domParser.parseFromString(
				`<div>
					<script><b></b></script>
					<style><b></b></style>
				</div>`,
				'application/xml'
			);
console.log(result.children[0].children[0].innerHTML);
```

TrevorBurnham:
I took a look at this PR and extracted some of these improvements from it to #1980.

PR #1677, opened 2025-01-08
chore: [#0] Improve type accuracy, strict mode compatibility and JSDoc documentation

wojtekmaj:
Fixes several easy to fix issues with TypeScript types accuracy, strict mode compatibility and JSDoc documentation.

While in non-strict mode, which you currently use, you will see no difference, you're 10% closer to enabling strict mode now!

Before, in strict mode:
> Found 2243 errors. Watching for file changes.

After, in strict mode:
> Found 2041 errors. Watching for file changes.

Most common issues:
- `null` not accepted where some value or undefined was accepted
- `(arg: any) => void | null` does NOT mean the same as `((arg: any) => void) | null` - the former means "a function that returns nothing or returns null", the latter - "a function that returns nothing, or no function (null)"


capricorn86 (approved):
Thank you for your contribution @wojtekmaj! :star: 

Many good catches. Enabling strict mode would be great. Hopefully it can be prioritized soon.

PR #1681, opened 2025-01-09
feat: [1502] Reduce duplication in tests

OlaviSau:
I'll start off with some examples before fully committing as there a lot of tests. If you're happy with these changes feel free to merge this and I'll continue with the refactor. By the way, perhaps you should allow for refactor type in commit message.

### DIFF
--- a/packages/happy-dom/test/fetch/Fetch.test.ts
+++ b/packages/happy-dom/test/fetch/Fetch.test.ts
@@ -3,22 +3,18 @@ import Response from '../../src/fetch/Response.js';
 import Headers from '../../src/fetch/Headers.js';
 import DOMException from '../../src/exception/DOMException.js';
 import DOMExceptionNameEnum from '../../src/exception/DOMExceptionNameEnum.js';
-import AbortController from '../../src/fetch/AbortController.js';
 import HTTP, { ClientRequest } from 'http';
 import Net from 'net';
 import Stream from 'stream';
 import Zlib from 'zlib';
 import { TextEncoder } from 'util';
 import Blob from '../../src/file/Blob.js';
-import FormData from '../../src/form-data/FormData.js';
 import { URLSearchParams } from 'url';
 import '../types.d.js';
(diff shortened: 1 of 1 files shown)


packages/happy-dom/test/fetch/Fetch.test.ts
```diff
@@ -3,22 +3,18 @@ import Response from '../../src/fetch/Response.js';
 import Headers from '../../src/fetch/Headers.js';
 import DOMException from '../../src/exception/DOMException.js';
 import DOMExceptionNameEnum from '../../src/exception/DOMExceptionNameEnum.js';
-import AbortController from '../../src/fetch/AbortController.js';
```
OlaviSau, line 6: These were unused.

packages/happy-dom/test/fetch/Fetch.test.ts
```diff
@@ -30,6 +26,42 @@ const PLATFORM =
 	process.arch;
 
 describe('Fetch', () => {
+	async function mockNetwork(
```
capricorn86, line 29: I think we could move this to `test/setup.ts`, but it doesn't have to be done in this PR

capricorn86 (approved):
Sorry for taking some time to review. I think it looks good! :slightly_smiling_face: It definitely reduces duplicated code.

I had one comment, but it doesn't have to be done in this PR. Let me know what you think.

packages/happy-dom/test/fetch/Fetch.test.ts
OlaviSau: Hmm - In that case mockAsyncNetwork for now? It's actually something to understand as to if would be possible to make SyncFetch consistent with Fetch so mockNetwork would apply to both of them 😁

capricorn86: Yes, mockAsyncNetwork would work :slightly_smiling_face:

capricorn86:
I merged this now. We can move the function to the test setup later.

Issue #1683, opened 2025-01-10
getByRole doesn't work - CSS subsequent sibling combinator (~) 


karpiuMG:
**Describe the bug**
I have an input which is styled with using subsequent sibling combinator (~). With `happy-dom` my test fails but in `jsdom` it works. With these styles the element role - checkbox - is not seen. It seems like the space after input in styles make a difference:

```
// fails input~[doesntMatter]
it('with subsequent-sibling combinator ~', () => {
  render(
    <>
      <style>{`
        input~[doesntMatter] {
          display: none;
        }
      `}</style>
      <input type="checkbox" />
    </>
  );

  expect(screen.getByRole('checkbox')).toBeInTheDocument();
});

// works input ~ [doesntMatter]
it('with subsequent-sibling combinator ~', () => {
  render(
    <>
      <style>{`
        input ~ [doesntMatter] {
          display: none;
        }
      `}</style>
      <input type="checkbox" />
    </>
  );

  expect(screen.getByRole('checkbox')).toBeInTheDocument();
});
````

[Example on Stackblitz](https://stackblitz.com/edit/github-5r4uqn-ax1xemm5?file=src%2FApp.test.jsx)

it seems to be related to
@capricorn86 in https://github.com/capricorn86/happy-dom/issues/1122#issuecomment-1998078528_

PR #1686, opened 2025-01-12
feat: [#1683] Adds support for subsequent sibling combinator to query selectors and match

karpiuMG:
…uerySelector


karpiuMG:
@capricorn86 can you help me with the failing build (18)?

capricorn86:
> ~@capricorn86 can you help me with the failing build (18)?~
> 
> rerun helped

Yeah, there are some tests that are timer based and are bit flaky that needs to be fixed.

capricorn86:

I made a commit to improve the performance of the loop a bit, added support for `matches()` and fixes to make sure that the cache is cleared if any element is changed.

Issue #320, opened 2021-12-20
Cannot use import statement outside a module


jakobrosenberg:
Am I required to bundle dependencies or is there a way to support dynamic imports with `happy-dom`?

capricorn86:
Hi @jakobrosenberg! :slightly_smiling_face: 

Yes this is a feature I have been thinking about adding as I also have a need for it. Not sure exactly when I will get the time to add it, but I will do it as soon as possible.

jakobrosenberg:
That sounds amazing!

If there's anything I can do to help, let me know. 🙂

jakobrosenberg:
Any news on this? Still happy to help.

capricorn86:
@jakobrosenberg sorry for a late answer. It has been a lot in my private and working life.

This is not that easy to implement, but definitely something I want to do when the major bugs has been cleared up.

denniseffing:
We just ran into this issue as well. Due to this issue, we are not able to write unit tests using `vitest` and `@testing-library/dom`. Not even `jsdom` supports this https://github.com/jsdom/jsdom/issues/2475 😢 .

capricorn86:
There is now support for this in v17.0.0 🙂

PR #1705, opened 2025-01-26
BREAKING CHANGE: [#320] Adds support for ECMAScript modules

capricorn86:


Spixmaster:
Thank you very much for your work. I am really impressed by the high rate of changes to the source code.

PR #1707, opened 2025-01-29
fix: [1706] Handle non string values gracefully for removeAttribute o…

OlaviSau:
…n elements

### DIFF
--- a/packages/happy-dom/src/nodes/element/Element.ts
+++ b/packages/happy-dom/src/nodes/element/Element.ts
@@ -33,6 +33,9 @@ import NodeFactory from '../NodeFactory.js';
 import HTMLSerializer from '../../html-serializer/HTMLSerializer.js';
 import HTMLParser from '../../html-parser/HTMLParser.js';
 import IScrollToOptions from '../../window/IScrollToOptions.js';
+import { AttributeUtility } from '../../utilities/AttributeUtility.js';
+import DOMException from '../../exception/DOMException.js';
+import DOMExceptionNameEnum from '../../exception/DOMExceptionNameEnum.js';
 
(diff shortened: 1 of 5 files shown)


packages/happy-dom/src/nodes/element/Element.ts
```diff
@@ -793,6 +793,9 @@ export default class Element
 	 * @param name Name.
 	 */
 	public removeAttribute(name: string): void {
+		if (typeof name !== 'string') {
```
capricorn86, line 796: The name is actually not removed, it is stringified.

You can test it with this:
```javascript
var div = document.createElement('div');

div.setAttribute(null, 'test');

// Outputs '<div null="test"></div>'
console.log(div.outerHTML);

div.removeAttribute(null);

// Outputs '<div></div>'
console.log(div.outerHTML);
```

I think it is best if we add the stringify using `String()` to `NamedNodeMap` to methods such as `getNamedItem()`.

capricorn86 (changes requested):
Nice contribution!

packages/happy-dom/src/nodes/element/Element.ts
```diff
@@ -669,6 +672,13 @@ export default class Element
 	 * @param value Value.
 	 */
 	public setAttribute(name: string, value: string): void {
+		if (AttributeUtility.isValidAttributeName(name)) {
```
OlaviSau, line 675: This is a breaking change even though it's more correct this way.

OlaviSau:
One note - there is no need to implement a check on NamedNodeMap setNamedItem because Attr constructor is not available in the first place, thus an invalid name cannot be created unless the user explicitly goes out of their way to do so.

OlaviSau:
This is quite a significant breaking change even though it's more correct this way. Perhaps we should add a strict mode flag and not enable it by default and have a transition period?

PR #1713, opened 2025-01-31
fix: [1710]  - Ensure querySelector returns the first item that appears in the DOM for grouped selectors

### LINKED ISSUE #1710: QuerySelector returning incorrect results when using a grouped selector
**Describe the bug**
My team is trying to replace JSDOM with happy-dom and we are hitting some unexpected differences in behavior with regards to query selector. In particular, we have a minimal repro of a query selector that does not match the output of JSDOM or chromium.

**To Reproduce**
This unit test in QuerySelector.test.ts captures the issue. 

```typescript
	it('Supports grouped selectors', () => {

			div.innerHTML = `
				<div class>
            		<h1><span>Here is a heading</span></h1>
            		<div class="a">
              			<span>With a child span</span>
            		</div>
          		</div>
			`;

			expect(div.querySelector('.a,h1')).toBe(div.children[0].children[0])
		})
```

**Expected behavior**
The behavior here is that the div with class a is being returned, instead of the h1 which is what is returned in jsdom and in the browser.

**Screenshots**

**Device:**
Node 22 on Windows, MacOS, and Linux

**Additional context**

christiango:
Fixes #1710

Added a test case to query selector that in the current build of happy-dom fails due to it returning the first item that matches the grouped selector instead of the first element in the DOM that matches any selector in the group. This matches the behavior of Chromium browsers as well as JSDOM.


packages/happy-dom/src/query-selector/QuerySelector.ts
```diff
@@ -466,6 +443,44 @@ export default class QuerySelector {
...
+	 * @param cachedItem Cached item.
+	 
+	 */
+	private static insertMatchesInOrder(
+		items: Element[],
+		node: Element | Document | DocumentFragment,
+		selector: string,
+		cachedItem: ICachedQuerySelectorAllItem | ICachedQuerySelectorItem
+	): void {
+		const groups = SelectorParser.getSelectorGroups(selector);
```
christiango, line 461: My main concerns with this change are the following:
* I'm new to this project so I'm afraid this may have some unexpected consequences due to the complex nature of the caches required to make these perform well
* I'm not sure how much this impacts performance of the library all up, particularly in less complicated query selector scenarios than the one described in the bug.

capricorn86 (changes requested):
Thank you for contributing @christiango! :star: 

I believe that this fix would have a big impact on performance.

I think the best way to solve this bug is to add support for returning `DocumentPositionAndElement` to `findFirst()` (similar as what findAll() does). Then `querySelector()` collects the matches for all groups and sorts them by position (similar to what querySelectorAll() does).

christiango:
Great suggestion, I'll give that a shot!

capricorn86 (approved):
Nice work! :slightly_smiling_face:

PR #1721, opened 2025-02-10
fix: [1720] Fix bug where nested query selector is not returning the correct result when there are multiple matching selectorGroups

### LINKED ISSUE #1720: queryResult is not returning the correct results when there are nested tags
**Describe the bug**
When running query selector all on HTML with nested matches, queryResult is not returning the correct match. queryResultAll returns the correct results at array position 0

**To Reproduce**

Create HTML 	 
```html
<div>
              <blockquote>
                <div class="a">
                  <ul>
                    <li><span>Item 1</span></li>
                    <li><span>Item 2</span></li>
                  </ul>
                </div>
              </blockquote>
            </div>
```
Run query
```javascript
querySelector('.a,BLOCKQUOTE');
```

The returned result is the div with class a instead of the blockquote element

**Expected behavior**
The blockquote element is returned.

**Device:**
Repros on all endpoints
**Additional context**

christiango:
Found and fixed a minor bug where we still aren't quite returning the same result in querySelect() and querySelectAll()[0] when there are grouped selectors. The fix was to pass through the updated document position to recursive calls, which was missed during #1713.

Fixes #1720

### DIFF
--- a/packages/happy-dom/src/query-selector/QuerySelector.ts
+++ b/packages/happy-dom/src/query-selector/QuerySelector.ts
@@ -601,7 +601,8 @@ export default class QuerySelector {
 									rootElement,
 									[nextElementSibling],
 									selectorItems.slice(1),
-									cachedItem
+									cachedItem,
+									position
 								);
 								if (match) {
 									return match;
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Ah, we missed that one :sweat_smile:

Issue #1724, opened 2025-02-11
Add support for the `prefers-reduced-motion` media feature


mwdiaz:
**Is your feature request related to a problem? Please describe.**
We have React components that rely on a media query to remove/reduce nonessential animations CSS transitions based on this setting, and it would be great if our unit tests for these components could use this as well (especially since it simplifies things like faking timers or waiting for elements to appear/disappear in the DOM).

**Describe the solution you'd like**
I have a working implementation that we're using locally with `pnpm patch` and I'm going to submit a PR for that so it can be reviewed and hopefully added to the project.

**Describe alternatives you've considered**
I have a proposed implementation.

PR #1725, opened 2025-02-11
feat: [#1724] Add prefers-reduced-motion support

mwdiaz:
As mentioned in #1724, this is an implementation we're using locally via `pnpm patch` and it's been working well. I modeled the change after work done previously in #354, so hopefully I'm not too off base. I did consider making the `prefersReducedMotion` setting a boolean, but I think it makes more sense to accept the valid string values for the setting, which is also more future proof if they ever decide to extend the spec to accept more values.

Thanks for all the work on this project!


capricorn86 (approved):
Thank you for your contribution @mwdiaz! :star: 

I pushed in a fix to check if settings exist as it can potentially be null during tear down. It seems like it was missed before as well.

PR #1732, opened 2025-02-18
refactor: Node API replaces cp command

btea:
The node api replaces the cp command, which is not compatible with Windows.


capricorn86 (approved):

Issue #1689, opened 2025-01-16
Dataset value is returned as undefined when it is set to empty string


pdanpdan:
**Describe the bug**
Happy-Dom evaluates a dataset key as undefined when it is set to an empty string.
The browser evaluates it correctly to empty string.

**To Reproduce**
1. `document.body.dataset.test = ''`
2. `console.log(document.body.dataset.test)`
3. It will output `undefined` while it should output `''`

**Expected behavior**
It should be evaluated to empty string.

**Additional context**
It is a regression introduced in v16.0.0 - latest working version is v15.11.7.

pdanpdan:
A relevant regression test would be

```js
element.setAttribute('data-test-empty', '');
const dataset = element.dataset;
expect(dataset.testEmpty).toBe('');
expect(dataset.nonExistentKey).toBeUndefined();
```

PR #1738, opened 2025-02-23
fix: [#1689] Fix dataset get proxy handler - return an attribute only…

karpiuMG:
… when it was found


capricorn86 (approved):

PR #1744, opened 2025-02-24
fix: [#1563] Prevent disabled inputs from being focussed or blurred

### LINKED ISSUE #1563: Calling focus on a disabled input should not fire focus event or change focus of active element
**Describe the bug**
As per the spect (see https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled#attribute_interactions), calling focus() on an input element that is disabled should fire focus or change the active element. 

**To Reproduce**
Discovered when attempting to convert tests from jsdom to happy-dom

```
    it('input should not have focus if disabled', () => {
        props.disabled = true;
        renderComponent();

        screen.getByTestId('input').focus();

        expect(screen.getByTestId('input')).not.toHaveFocus();
    });
```

**Expected behavior**
Calling focus() when an input is disabled should not dispatch any events

karpiuMG:
resolve #1563


capricorn86 (approved):

PR #1748, opened 2025-02-26
fix: [#1747] Event phases should be on both class and instance

### LINKED ISSUE #1747: Event phases should be on both class and instance
**Describe the bug**

Event phases (`Event.NONE`, `Event.CAPTURING_PHASE`, `Event.AT_TARGET`, and `Event.BUBBLING_PHASE`) should be accessible on both class and instance. However, they are only set as instance properties (e.g. `new Event().NONE`) in current implementation.

**To Reproduce**
1. Use `Event.CAPTURING_PHASE`
2. TypeScript would think its type/value is `1`, but at runtime it is `undefined`

**Expected behavior**
```ts
const e = new Event("click");
expect(Event.NONE).toEqual(e.NONE);
expect(Event.CAPTURING_PHASE).toEqual(e.CAPTURING_PHASE);
expect(Event.AT_TARGET).toEqual(e.AT_TARGET);
expect(Event.BUBBLING_PHASE).toEqual(e.BUBBLING_PHASE);
```

alan910127:
Closes #1747.

### DIFF
--- a/packages/happy-dom/src/event/Event.ts
+++ b/packages/happy-dom/src/event/Event.ts
@@ -12,10 +12,15 @@ import Document from '../nodes/document/Document.js';
  * Event.
  */
 export default class Event {
-	public NONE = EventPhaseEnum.none;
-	public CAPTURING_PHASE = EventPhaseEnum.capturing;
-	public AT_TARGET = EventPhaseEnum.atTarget;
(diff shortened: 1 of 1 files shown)


capricorn86 (approved):
Thank you for your contribution @alan910127! :star:

PR #1751, opened 2025-02-27
feat: [#990] Add support for multiple selectors in :not pseudo-class

### LINKED ISSUE #990: `querySelectorAll` ignores `:not()` pseudo-class with multiple selectors
Enter the following code into https://npm.runkit.com/happy-dom

```js
var happyDom = require("happy-dom")

const window = new happyDom.Window();
document.body.innerHTML = `<ul class="list">
  <li class="list-item"></li>
  <li class="list-item"></li>
  <li class="list-item"></li>
  <li class="other-item"></li>
  <li></li>
</ul>`;

const lastItem = document.querySelectorAll('ul > li:not(.list-item, .other-item)');

console.log('lastItem', lastItem.length); // prints 2 but should be 1
```

`querySelectorAll` will only respect the first selector inside of the `:not` pseudo-class (`.list-item`) and ignores all subsequent selectors (`.other-item`).

**Expected behavior**
It respects the complete behaviour, especially comma-separated selector lists, of the [not()](https://developer.mozilla.org/en-US/docs/Web/CSS/:not) pseudo-class.

karpiuMG:
close #990

### DIFF
--- a/packages/happy-dom/src/query-selector/SelectorItem.ts
+++ b/packages/happy-dom/src/query-selector/SelectorItem.ts
@@ -248,7 +248,12 @@ export default class SelectorItem {
 			case 'root':
 				return element[PropertySymbol.tagName] === 'HTML' ? { priorityWeight: 10 } : null;
 			case 'not':
-				return !pseudo.selectorItems[0].match(element) ? { priorityWeight: 10 } : null;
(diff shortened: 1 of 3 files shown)


capricorn86 (approved):

PR #1753, opened 2025-02-28
fix: [#1151] Skips svg elements when innerText is used

### LINKED ISSUE #1151: svg element innertText is undefined
**Describe the bug**
svg element innertText is undefined, it should be a string.

**To Reproduce**
```javascript

const window = new Window({
	url: 'https://localhost:8080',
	width: 1024,
	height: 768
});

const div = window.document.createElement('div');
div.innerHTML = '<div><span><svg></svg></span>123<div>';
window.document.body.appendChild(div);
console.log(div.innerText);
```
actual log: ```undefined123```

**Expected behavior**
expected log: ```123```

**Device:**
 - OS: mac
 - Browser: node
 - Version: 16.20.1
 - happy-dom version: 12.10.3

**Additional context**
in https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-element/HTMLElement.ts#L101, I think the judgment here should be whether it's HTMLElement or not, only htmlelement has innerText

karpiuMG:
+ fix: [#1151] Add newline if it's a block/flex element and there's more content coming after

for skipping svg element a change is needed for recognizing tag names as HTML Parser parse svg elements in a way that svg tagName is lowercase - which shouldn't be the case for HTML document context
ref. https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName#value

I leave it to you @capricorn86 - maybe you could add the change in your DOM Parsing PR #1666  

close #1151


capricorn86:
Thank you for your contribution @karpiuMG :star: 

I found a small thing.

capricorn86 (approved):
I made a performance fix and removed the comment :slightly_smiling_face:

PR #1754, opened 2025-03-01
feat: [#1184] Add support for form method='dialog'

### LINKED ISSUE #1184: method="dialog" form submission does not close the dialog element (modal)
**Describe the bug**

> When the form (of `method="dialog"`) is inside a [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog), closes the dialog and causes a submit event to be fired on submission, without submitting data or clearing the form.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#method

**To Reproduce**

```javascript

const window = new Window();

// Copy below and paste it in a browser console.

document.body.innerHTML = `
<dialog>
  <form method="dialog">
    <button>Close</button>
  </form>
</dialog>`;

const dialog = document.querySelector('dialog');
console.log(dialog.open); // false

dialog.showModal();
console.log(dialog.open); // true

const button = document.querySelector('button');
button.click();
console.log(dialog.open); // true - should be false

const form = document.querySelector('form');
form.submit();
console.log(dialog.open); // true - should be false

```

**Expected behavior**

Dialog should have been closed on the button press.

**Device:**

OS: macOS 14.2.1
CPU: (10) arm64 Apple M1 Pro
Memory: 85.06 MB / 16.00 GB

Chrome: 120.0.6099.129
Edge: 120.0.2210.91
Safari: 17.2.1

karpiuMG:
feat: [#1184] Close modal on pressing ESC key

close #1184


packages/happy-dom/src/nodes/html-dialog-element/HTMLDialogElement.ts
```diff
@@ -65,15 +66,31 @@ export default class HTMLDialogElement extends HTMLElement {
 		this.removeAttribute('open');
 		this.returnValue = returnValue;
 		if (wasOpen) {
+			this.#removeKeydownListener();
 			this.dispatchEvent(new Event('close'));
 		}
 	}
 
+	#keydownHandler = (event: KeyboardEvent): void => {
```
capricorn86, line 74: I'm not able to replicate this behavior in the browser.

I believe that the native behavior can't be replicated programmatically by dispatching events. The only way to implement the possibility to trigger this behavior is to implement this in the [Browser Page](https://github.com/capricorn86/happy-dom/wiki/BrowserPage).

```javascript
var element = document.createElement('dialog');

document.body.appendChild(element);

element.showModal();

// Outputs "true"
console.log(element.open);

document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', code: 'Escape', keyCode: 27 }));
document.dispatchEvent(new KeyboardEvent('keyup', { key: 'Escape', code: 'Escape', keyCode: 27 }));

// Outputs "true"
console.log(element.open);
```

packages/happy-dom/src/nodes/html-dialog-element/HTMLDialogElement.ts
```diff
@@ -65,15 +66,31 @@ export default class HTMLDialogElement extends HTMLElement {
...
+		if (event.key === 'Escape') {
```
capricorn86, line 75: In order to support clicking a button, we might be able to override `dispatchEvent()` and check if the event is "click" and the target is a button or input with type button.

capricorn86 (changes requested):

capricorn86 (dismissed):
I have added some fixes to work according to spec :slightly_smiling_face:

PR #1755, opened 2025-03-01
fix: [#1698] Handle invalid indeces on childNodes

### LINKED ISSUE #1698: NodeList gets the first element when you pass and empty array as index
**Describe the bug**
When you use an empty array as an index of a NodeList it returns the first element instead of undefined.

**To Reproduce**
We have the following function:
```javascript
const getActiveTabTrackId = () => {
    const bar = document.querySelector(".container");
    const active = bar && Object.keys(bar.childNodes).filter((key) => bar.childNodes[key].classList.contains('active');
    return bar && bar.childNodes[active] ? bar.childNodes[active].getAttribute('data-track-id') : null;
}
``` 

And we are testing it using:
```javascript
test('returns null if no active tab are present', () => {
    document.body.innerHTML = '<div class="container">' +
    '<div class="tab" data-track-id="one"></div>' +
    '<div class="tab" data-track-id="two"></div>' +
    '</div>';

    expect(getActiveTabTrackId()).toBe(null);
})
``` 

When we run it, test fails because NodeList its returning the first element instead of undefined

**Expected behavior**
`document.querySelector(".container").childNodes[[]]` should return undefined instead of the first element.

**Device:**
 - OS: macos
 - Browser: happy-dom and jest
 - Version: 16.6.0

**Additional context**
We are using Jest and @happy-dom/jest-environment

karpiuMG:
close #1698

### DIFF
--- a/packages/happy-dom/src/nodes/node/NodeList.ts
+++ b/packages/happy-dom/src/nodes/node/NodeList.ts
@@ -38,6 +38,9 @@ class NodeList<T extends Node> {
 					methodBinder.bind(property);
 					return target[property];
 				}
+				if (property === '') {
+					return undefined;
+				}
 				const index = Number(property);
 				if (!isNaN(index)) {
 					return items[index];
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):

PR #1757, opened 2025-03-01
fix: [#1463] Element.contentEditable should be synced with the "contenteditable" attribute

### LINKED ISSUE #1463: `HTMLElement.contentEditable` should apply the `contenteditable` attribute.
**Describe the bug**
When changing the `HTMLElement.contentEditable` property, the element does not get the `contenteditable` attribute.

**To Reproduce**
Repository to reproduce: https://github.com/mozesstumpf/contentEditable-attribute-should-change
```
test("ContentEditable attribute should change", () => {
	const div = document.createElement("div");

	div.contentEditable = "true";

	// should be passed
	expect(div.getAttribute("contenteditable")).toBe("true");
});
```

**Expected behavior**
The `contenteditable` attribute should be added to the element when the `HTMLElement.contentEditable` changed.

karpiuMG:
…r is invoked

close #1463 

I've added error throwing to mirror Chrome's behavior


capricorn86 (approved):
Thank you for you contribution @karpiuMG! :star: 

I have pushed in some fixes to the implementation, so that it works according to spec.

PR #1759, opened 2025-03-01
feat: [#1427] Support submitter in formData constructor and exclude s…

### LINKED ISSUE #1427: FormData constructed from a <form> includes buttons by default
**Describe the bug**
When constructing a `FormData` instance from a `HTMLFormElement` containing a button with a `name` and `value`, that button's value is always included in the resulting `FormData`.

This isn't the case for any browser I've tested (Chromium, Firefox, WebKit). A submit button's value is only included in form submissions when that button is the one that triggered the form submission, so it makes sense to exclude button values by default.

The [XMLHttpRequest Spec](https://xhr.spec.whatwg.org/#dom-formdata) defines an optional `submitter` parameter that can be used to include a submit button's value in the resulting form data instead.

**To Reproduce**
```ts
describe('FormData', () => {
  describe('Constructor', () => {
    it('Only includes button values when they are passed in as submitter', () => {

      input.name = 'input';
      input.value = 'testing';

      button.name = 'button';
      button.value = 'buttonValue';

      form.append(input);
      form.append(button);

      const formDataWithoutButton = new window.FormData(form);
      const formDataWithButton = new window.FormData(form, button);

      expect(formDataWithoutButton.get('input')).toBe('testing');
      expect(formDataWithoutButton.get('button')).toBeNull();

      expect(formDataWithButton.get('input')).toBe('testing');
      expect(formDataWithoutButton.get('button')).toBe('buttonValue');
    });
  });
});
```

**Expected behavior**
Constructing a new FormData instance from the form without passing in the button as submitter shouldn't include the `button`'s value.

Additionally it would be nice if the `submitter` parameter could be supported by happy-dom.

**Device:**

**Additional context**
Looks like the `submitter` parameter has been in the spec since the end of January 2023 [(ref)](https://github.com/whatwg/xhr/issues/262), and implemented in most major browsers [since April 2023](https://caniuse.com/mdn-api_formdata_formdata_submitter).

FWIW jsdom doesn't support the `submitter` parameter either, but it _does_ exclude button values by default.

karpiuMG:
…ubmit buttons from formData by default

close #1427 

I've checked behavior of formData in Chrome and indeed buttons are not included by default in the formData - that's why existing tests were to change

### DIFF
--- a/packages/happy-dom/src/form-data/FormData.ts
+++ b/packages/happy-dom/src/form-data/FormData.ts
@@ -4,6 +4,9 @@ import File from '../file/File.js';
 import HTMLInputElement from '../nodes/html-input-element/HTMLInputElement.js';
 import HTMLFormElement from '../nodes/html-form-element/HTMLFormElement.js';
 import BrowserWindow from '../window/BrowserWindow.js';
+import HTMLElement from '../nodes/html-element/HTMLElement.js';
+import DOMException from '../exception/DOMException.js';
+import HTMLButtonElement from '../nodes/html-button-element/HTMLButtonElement.js';
 
 type FormDataEntry = {
 	name: string;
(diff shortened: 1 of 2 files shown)


packages/happy-dom/src/form-data/FormData.ts
```diff
@@ -25,12 +28,45 @@ export default class FormData implements Iterable<[string, string | File]> {
...
 	 * @param [form] Form.
+	 * @param [submitter] A submit button that is a member of the form.
 	 */
-	constructor(form?: HTMLFormElement) {
+	constructor(form?: HTMLFormElement, submitter?: HTMLInputElement | HTMLButtonElement) {
 		if (!form) {
 			return;
 		}
 
+		if (submitter && submitter instanceof HTMLElement) {
```
capricorn86, line 38: Looking at JSDOM, it seems like it should be implemented a bit different. We will have to look into the spec a bit more.

Link to JSDOM implementation:
https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/xhr/FormData-impl.js

capricorn86 (changes requested):

capricorn86:

This has been fixed now in #1860. I chose to go forward with it as it as first glance seemed simpler, but looking at the spec your PR covered it as well and handled errors which was nice.

I'm closing this PR now. Sorry that this took such a long time.

PR #1772, opened 2025-03-11
fix: [#1772] Set 0 instead of undefined as default in setTimeout to prevent Bun from logging a "TimeoutNaNWarning"

lekoala:
Without this, I get a TimeoutNaNWarning: NaN is not a number. (with bun)

It doesn't seem a bad thing to actually set the timer to 0 for all anyway, but it might very well be something that bun needs to fix


capricorn86 (approved):
Thank you for your contribution @lekoala! :star: 

No, it doesn't hurt to add it :slightly_smiling_face:

PR #1780, opened 2025-03-28
fix: [#1779] Allow empty URL string in Request

### LINKED ISSUE #1779: Fetch an empty URL fails
**Describe the bug**
When executing `fetch("", ...)`, the `Request` constructor throws the error `TypeError: Failed to contruct 'Request': 1 argument required, only 0 present.` due to the empty string being falsy in the [`constructor` assertion](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/fetch/Request.ts#L69).

**To Reproduce**
1. Navigate to or create a Browser Page
2. Run `window.fetch("")`

**Expected behavior**
Calling `window.fetch("")` should be equivalent `window.fetch(document.baseURI)`.

**Screenshots**
Not applicable.

**Device:**
All devices.

**Additional context**
My use case is running a Next.js app using Server Actions, [which are run using `fetch("", ...)`](https://github.com/vercel/next.js/blob/ca4a18789d37dbd01ca3c88091d1f86134072dee/packages/next/src/client/components/router-reducer/reducers/server-action-reducer.ts#L90). However, the problem itself is not specific to Next.js.

elierotenberg:
* Allow empty request URL string, e.g. when running `window.fetch("")`.
* Fixes #1779

### DIFF
--- a/packages/happy-dom/src/fetch/Request.ts
+++ b/packages/happy-dom/src/fetch/Request.ts
@@ -66,7 +66,7 @@ export default class Request implements Request {
 			);
 		}
 
-		if (!input) {
+		if (typeof input !== `string` && !input) {
 			throw new window.TypeError(
 				`Failed to contruct 'Request': 1 argument required, only 0 present.`
 			);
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Thank you for your contribution @elierotenberg ! :star: 

I had to fix the unit tests as it can't resolve "" to be relative to "about:blank". In this case it should fail.

Issue #1782, opened 2025-03-30
Implement XMLHttpRequst.overideMimeType


maxmil:
**Is your feature request related to a problem? Please describe.**
XMLHttpRequest doesn't currently implement the web standard [overrideMimeType](https://xhr.spec.whatwg.org/#the-overridemimetype()-method) . 

I am using a library that calls this API and the fact that it is missing raises an error.

**Describe the solution you'd like**
To add an implementation of it

**Describe alternatives you've considered**
I've currently patched XMLHttpRequest in my client code by dynamically adding the method but I'd prefer not to have to do this.

maxmil:
Created PR https://github.com/capricorn86/happy-dom/pull/1783

PR #1783, opened 2025-03-30
feat: [#1782] Add overrideMimeType to XMLHttpRequest

maxmil:
Implement this web standard according to the specs.

https://xhr.spec.whatwg.org/#the-overridemimetype()-method


capricorn86 (approved):
Thank you for your contribution @maxmil! :star: 

Sorry for taking such a long time before merging.

Issue #1775, opened 2025-03-17
Anchor tags should navigate when they receive a bubbled click event


maxmil:
**Describe the bug**
From the [WHATWG living spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) on the content model

> Content model: Transparent, but there must be no interactive content descendant.

I might be wrong but I understand this to mean that if you wrap non interactive elements inside an anchor tag event bubbling should mean that the anchor tag navigates when a click event bubbles up to it.

**To Reproduce**
```
const browser = new Browser();
const page = browser.newPage();
page.content = '<html><body><a href="https://example.com"><div></div></a></body></html>';

const div = page.mainFrame.document.body.querySelector('div')!;
div.click();
await page.waitForNavigation();
```

**Expected behavior**
The above script should complete successfully however it hangs because the anchor tag ignores the bubbling event.

It looks from the [code](https://github.com/capricorn86/happy-dom/blob/b61762e732872651af11f0c07c12a90850ac830f/packages/happy-dom/src/nodes/html-anchor-element/HTMLAnchorElement.ts#L414) that this was an explicit choice so perhaps I'm missing some context?

maxmil:
If it's wasn't an explicit choice I think the fix is pretty simple.

I've added a PR https://github.com/capricorn86/happy-dom/pull/1784

PR #1784, opened 2025-03-30
fix: [#1775] Anchor tag handles bubbling clicks

maxmil:
From the spec anchor tags should handle bubbling events.

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element

This is also the behaviour of Chrome and Firefox.


capricorn86 (approved):
Thank you for your contribution @maxmil! ⭐

PR #1786, opened 2025-04-01
fix: Overwrite existing URL query string on form submit when method is "GET"

rslabbert:
Aligns with the [HTML spec](https://html.spec.whatwg.org/#constructing-form-data-set) ("4. Let entry list be a new empty entry list.") and observed behaviour in browsers. Previously the existing query string would be appended to.


capricorn86 (approved):

Issue #1774, opened 2025-03-16
Form submit does not honour enctype


maxmil:
**Describe the bug**
When a form is submitted, if the form's method is not _get_ the form data is always encoded as multipart form data.

**To Reproduce**
```TypeScript
const browser = new Browser({
  settings: {
    fetch: {
      interceptor: {
        beforeAsyncRequest: async ({ request, window }) => {
          return new window.Response(await request.text());
        },
      },
    },
  },
});

const page = browser.newPage();
page.url = 'https://example.com';
page.content = `
  <html>
    <body>
      <form action="#" method="post">
        <input name="field1" type="text" value="foo">
        <input name="field2" type="text" value="bar">
      </form>
    </body>
  </html>`;

const form = page.mainFrame.document.body.querySelector('form');
form?.submit();
await page.waitForNavigation();

console.log(page.mainFrame.document.body.innerHTML);
```

Output: 

```
------HappyDOMFormDataBoundary0.h8oieomkqnh
Content-Disposition: form-data; name="field1"

foo
------HappyDOMFormDataBoundary0.h8oieomkqnh
Content-Disposition: form-data; name="field2"

bar
------HappyDOMFormDataBoundary0.h8oieomkqnh--
```

**Expected behavior**
Since the default encoding for the form element is `application/x-www-form-urlencoded` we  would expect the form data to be url encoded.

**Additional context**
I've had a quick look at the sources and think I could create a PR for this. I just want to check first that there is not a reason for the current behaviour. Also, I'm new to Happy DOM so I may be missing a trick.

Spixmaster:
I hope this is related to https://github.com/capricorn86/happy-dom/issues/1761.

PR #1787, opened 2025-04-02
fix!: [#1774] Handle enctype for form submissions

rslabbert:
Forms now correctly handle `<form enctype="...">`, `<button formenctype="...">`, and `<input type="button" formenctype="...">`.

Limitations:
1. `text/plain` is not supported.
2. `<input type="image">` is not supported.

BREAKING CHANGE: The default encoding for form submissions is now `application/x-www-form-urlencoded` instead of `multipart/form-data`

PR Notes:
1. I've opted to change the `formData` field on `BrowserFrameNavigator.navigate` to allow passing in `URLSearchParams` instead of adding a new field or renaming the field as that seemed to be the simplest approach. That being said, this also now adds that as an option to the `IHistoryItem` field which might also result in a breaking change in types, so I'm happy to revisit this.
2. We could reuse the logic that creates the `URLSearchParams` in the `GET` submission and here, but that might be better done after #1786 has landed.
3. Submitting a form that contains `<input type="file">` without specifying `enctype="multipart/form-data"` will submit the form with `application/x-www-form-urlencoded` resulting in a `file=<filename>` body without any file contents which doesn't make much sense but does align with browser behaviour.

### DIFF
--- a/packages/happy-dom/src/browser/utilities/BrowserFrameNavigator.ts
+++ b/packages/happy-dom/src/browser/utilities/BrowserFrameNavigator.ts
@@ -39,7 +39,7 @@ export default class BrowserFrameNavigator {
 		url: string;
 		goToOptions?: IGoToOptions;
 		method?: string;
-		formData?: FormData | null;
+		formData?: FormData | URLSearchParams | null;
 		disableHistory?: boolean;
 	}): Promise<Response | null> {
 		const { windowClass, frame, url, formData, method, goToOptions, disableHistory } = options;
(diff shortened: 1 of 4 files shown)


capricorn86:

It's a bit complicated. I will soon revisit this PR again and try to understand what the best approach is that follows the spec.

rslabbert:
@capricorn86 I've rebased this onto the latest master. Let me know if you need any more information on the change or if it you'd like to explore an alternate approach.

Issue #1788, opened 2025-04-04
File constructor replaces slashes in filename with colons (e.g. foo/bar.jpg → foo:bar.jpg)


dyabol:
Hi,  
I encountered an inconsistency when using the `File` constructor in happy-dom compared to how browsers behave.

### Code example:

```js
const file = new File([], "foo/bar.jpg");
console.log(file.name);
```

In happy-dom, the result is:
```
foo:bar.jpg
```

Whereas in real browsers (Chrome, Firefox), the filename is:
```
foo/bar.jpg
```

### Specification reference

According to the [MDN documentation for the File constructor](https://developer.mozilla.org/en-US/docs/Web/API/File/File#filename), passing a file path (such as "foo/bar.jpg") is a legitimate use of the filename parameter.

Replacing / with : is non-standard and may lead to confusion or bugs when simulating file inputs in tests.

### Expected behavior

The slash / should be preserved in the filename (foo/bar.jpg).

### Environment
happy-dom version: 17.4.4

Node.js version: 20.15.1

Test environment: Vitest + happy-dom

THX

PR #1789, opened 2025-04-04
fix [#1788]: preserve slashes in filename and match browser behavior

dyabol:
The File constructor was incorrectly replacing slashes (/) in the filename with colons (:), resulting in names like "foo:bar.jpg" instead of the expected "foo/bar.jpg".

This commit updates the behavior to align with browser implementations:
- Slashes in the provided filename are preserved.

Reference: 
https://github.com/jsdom/jsdom/commit/2ab99ad18d58585e4958aca0911ed142fac80b71
https://developer.mozilla.org/en-US/docs/Web/API/File/File#filename

### DIFF
--- a/packages/happy-dom/src/file/File.ts
+++ b/packages/happy-dom/src/file/File.ts
@@ -37,7 +37,7 @@ export default class File extends Blob {
 
 		super(bits, options);
 
-		this.name = name.replace(/\//g, ':');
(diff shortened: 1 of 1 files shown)


capricorn86 (approved):
Thank you for your contribution @dyabol! ⭐

Issue #1785, opened 2025-03-30
React 19.1's new useId with unicode characters `«` and `»` break selectors


terrymun:
**Describe the bug**
The latest React 19.1 upgrade breaks unit tests that are using happy-dom under the hood, as the `useId` hook now returns a unicode ID like `«r123»` instead of `:r123:` in 19.0 and before: https://github.com/facebook/react/pull/32001

**To Reproduce**
1. Upgrade React to 19.1
2. Attempt to perform `querySelector` using an ID generated from `useId`, e.g. `document.querySelector('#«r123»')`
3. Notice that an error will be thrown:

```text
DOMException: Invalid selector: "#«r123»"
❯ Function.getSelectorGroups ../../node_modules/happy-dom/src/query-selector/SelectorParser.ts:226:10
❯ Function.querySelectorAll ../../node_modules/happy-dom/src/query-selector/QuerySelector.ts:118:33
❯ HTMLDocument.querySelectorAll ../../node_modules/happy-dom/src/nodes/document/Document.ts:1498:24
```

I am encountering this error whien using `happy-dom` in my vitest config to run unit tests.

A fix _might_ be pretty straightforward, seeing that it is also resolved when React introduced colons in their `useId` string: https://github.com/capricorn86/happy-dom/issues/877 (fixed in #879), but I could be very mistaken. 

**Expected behavior**
happy-dom should not throw an error when unicode characters in HTML `id` attribute is present, especially the `«` and `»` characters returned by React 19.1's updated `useId` hook.

**Device:**
 - OS: macOS 15.3.2
 - happy-dom: 17.4.4

terrymun:
@davilima6 What I've done is to manually patch it using `pnpm patch`:

```txt
diff --git a/lib/query-selector/SelectorParser.js b/lib/query-selector/SelectorParser.js
index 9a0dae8f490940ade33847504d7a8ca7d22e06d5..239fd8adb3f8825a815b6bc17c36a891471640aa 100644
--- a/lib/query-selector/SelectorParser.js
+++ b/lib/query-selector/SelectorParser.js
@@ -22,7 +22,7 @@ import DOMException from '../exception/DOMException.js';
  * Group 16: Pseudo element (e.g. "::after", "::-webkit-inner-spin-button").
  * Group 17: Combinator.
  */
[...]

PR #1794, opened 2025-04-08
feat: [#1785] Add support for unicode characters from React 19.1 `useId` hook

terrymun:
This PR resolves the bug reported in #1785.

In React 19.1, the `useId` hook starts using unicode characters `«` and `»`. This might be used as a class name or an ID (even though the latter is the most common use-case). The introduction of these new characters causes the attribute value to fail the check agains the pattern `SELECTOR_REGEXP`.

This PR adds support for these two characters inside the `SELECTOR_REGEXP` regular expression for class and ID selectors (capturing group 3 and group 4).


terrymun:
@capricorn86 THe tests are failing but it does not seem to be related to the changes made in this PR (the failure seems to be coming from `fetch()`). Could you guide me to how to ensure the builds are passing?

The latest release (v17.47) is still having issues with React's new ID pattern, and I think this PR will rectify the issue. Right now I'm relying on using `pnpm patch` to manually patch the selectors.

capricorn86 (approved):
Thank you for your contribution @terrymun! :star: 

Sorry for taking such a long time before merging. I've been working on a feature and there a lot going on in my private life.

I had to fix the unit tests as the change in `QuerySelectorHTML` caused several tests to fail.

terrymun:
@capricorn86 please don't apologize! Personal life comes first before everything else 😄 thanks for taking time to look into the PR and making the changes needed to pass the unit tests, I really appreciate it.

Issue #1795, opened 2025-04-08
Replace angular.io with angular.dev


BojanKogoj:
The old domain (https://angular.io) is only used for older versions, for current versions https://angular.dev/ is used.

PR #1796, opened 2025-04-08
chore: [#1795] Replace old Angular domain angular.io with angular.dev

BojanKogoj:
Fix for https://github.com/capricorn86/happy-dom/issues/1795


capricorn86 (approved):
Thank you for your contribution @BojanKogoj

Issue #1803, opened 2025-04-14
When using MutationObserver to observe element removal, previousSibling and nextSibling get a value of null


uxuip:
**Describe the bug**
When using MutationObserver to observe element removal, previousSibling and nextSibling get a value of null

**To Reproduce**
[online test](https://stackblitz.com/edit/vitejs-vite-hqdnaztn?file=happy-dom.test.ts)

**Expected behavior**
Getting the correct Sibling

PR #1805, opened 2025-04-17
fix: [#1803] Fixes issue where the sibling obtained when MutationObserver monitors node removal was null

uxuip:
Fixes: https://github.com/capricorn86/happy-dom/issues/1803

The test content is consistent with browser behavior:
> image


capricorn86 (approved):
Thank you for your contribution @uxuip! :star:

PR #1851, opened 2025-06-30
fix: `<link rel="modulepreload">`

iam-medvedev:
Currently, when using relative paths in `<link rel="modulepreload" href="/dist/..." />`, we will get a stable error from `new URL`. 

I moved the URL creation below, similar to `preloadResource`. In addition, there is no actual point in creating a `URL` if the `disableJavaScriptFileLoading` or `disableJavaScriptEvaluation` options are enabled.


Spixmaster:
A test would prevent that issue from happening.

iam-medvedev:
@Spixmaster sure, done

capricorn86 (approved):
Thank you for your contribution @iam-medvedev! :star: 

PR #1857, opened 2025-07-21
chore: [#1856] Change IterableIterator return type to ArrayIterator

### LINKED ISSUE #1856: Iterators should be ArrayIterator
Iterators in this library are declared to return IterableIterator<T>, but they can be ArrayIterator<T>.
Indeed, lib.dom.iterable.d.ts also declares every iterator as ArrayIterator or something like that.
ArrayIterator in lib.esnext.iterator.d.ts inherits IteratorObject<T>, so it supports iterator helper methods.

Example methods:
https://github.com/capricorn86/happy-dom/blob/v18.0.1/packages/happy-dom/src/fetch/Headers.ts#L125-L167

Change return types from IterableIterator to ArrayIterator

**Describe alternatives you've considered**
There are FormDataIterator and HeadersIterator declared in lib.dom.iterable.d.ts, so we can copy them here as well.
  - Currently they are structurally equivalent to ArrayIterator, so we do not need this.

**Additional context**
None

ikeyan:
Fixes #1856


capricorn86 (approved):
Thank you for your contribution @ikeyan! ⭐

Issue #1859, opened 2025-07-26
Buttons that do not trigger a form submit should not be added to form data


maxmil:
**Describe the bug**
If a form contains multiple buttons that define name and value (vanilla buttons or inputs of type submit, button or reset) when the form is submitted all their name value pairs are added to form data irrespecitve of whether they are the submitter.

In the HTML Living Spec in [4.10.22.4 Constructing the entry list](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set) it states that

>For each element field in controls, in tree order:
>
>If any of the following are true:
>   ...
>   field is a button but it is not submitter;
>   ....
>then continue.

**To Reproduce**
Create a form like so
```
<form method="post" action="http://example.com">
    <input type="submit" name="action" value="submitted">
    <input type="submit" name="action" value="notSubmitted">
    <input type="reset" name="action" value="reset">
    <input type="button" name="action" value="inputButton">
    <button name="action" value="vanillaButton"></button>
</form>
```

Click on the first button to submit the form

**Expected behavior**
The form data should contain only a single value `submitted` for `action`

PR #1860, opened 2025-07-26
fix: [#1859] Only add the value of buttons to the form data if they are the submitter

maxmil:


maxmil:
Ah. I've just noticed https://github.com/capricorn86/happy-dom/pull/1759 

Appologies for the duplication. This is another take on the same prolem.

capricorn86 (approved):
Thank you for your contribution @maxmil! :star: 

I have added some logic for error handling and tests as well.

capricorn86:
> Ah. I've just noticed #1759
> 
> Appologies for the duplication. This is another take on the same prolem.

No worries! Looking into it, this solution seemed simpler, so I chose to go forward with it, but I realized that the other solution would probably have worked as well.

maxmil:
Glad to see that this was accepted even if it took a while.

I'd also be interested including formAssociated custom elements when forms are submitted. Is that something that you would be interested in @capricorn86  if I put together a PR?

Issue #1858, opened 2025-07-22
forEach should accept `callback`'s `this` value and pass `this` to the third argument of `callback`


ikeyan:
**Describe the bug**
These are forEach from DOMTokenList and NodeList:
- `thisArg` should be of type `any`.
- the third argument of `callback` should be `this`, not an internal array.

https://github.com/capricorn86/happy-dom/blob/v18.0.1/packages/happy-dom/src/dom/DOMTokenList.ts#L226-L231
https://github.com/capricorn86/happy-dom/blob/v18.0.1/packages/happy-dom/src/nodes/node/NodeList.ts#L200C9-L205

These are forEach from Headers and FormData:
- should accept `thisArg?: any`

https://github.com/capricorn86/happy-dom/blob/v18.0.1/packages/happy-dom/src/fetch/Headers.ts#L119-L123
https://github.com/capricorn86/happy-dom/blob/v18.0.1/packages/happy-dom/src/form-data/FormData.ts#L93-L97

**To Reproduce**
```
// node
const {FormData} = require('happy-dom')
var d = new FormData()
d.append('a', 'b')
const {strict: assert} = require('assert')
d.forEach(function(v, k, t) { console.log(this, v, k, t) }, {foo: 'foo'})
// FormData {} b a FormData {}
d.forEach(function(v, k, t) { assert.equal(this.foo, 'foo'); assert.equal(v, 'b'); assert.equal(k, 'a'); assert.equal(t, d); }, {foo: 'foo'})
// Uncaught AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
//   undefined !== 'foo'
```

**Expected behavior**
`forEach` should work in the same way as `Array.prototype.forEach` or `Map.prototype.forEach`.

**Screenshots**
Not applicable.

**Device:**
Device independent

**Additional context**
References to Web IDL spec. There are 3 definitions of forEach depending on the kind of interface but they are essentially the same

- https://webidl.spec.whatwg.org/#js-iterable
- https://webidl.spec.whatwg.org/#js-map-forEach
- https://webidl.spec.whatwg.org/#js-set-forEach

PR #1861, opened 2025-07-30
fix: [#1858] `forEach` should accept `callback`'s `this` value and pass `this` to the third argument of `callback`

ikeyan:
Tests are also added


capricorn86 (dismissed):
Thank you for your contribution @ikeyan! ⭐

I'm sorry that you had to wait this long.

I made a couple of fixes:
 - Improved the performance of the for loops
 - Default to `Window` for `thisArg`
 - A couple of more unit tests

Issue #1846, opened 2025-06-16
Support `popstate` events on history navigation


taraldr:
**Is your feature request related to a problem? Please describe.**
When navigating a browser history session`popstate` events are not currently emitted. This makes it harder to test behaviors which depend upon history, such as testing how a `history.back()` is handled in my application.

**Describe the solution you'd like**
I would like `popstate` events to be emitted according to the browser [spec](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event) during navigation.

**Describe alternatives you've considered**
I've attempted dispatching `popstate` events manually in my tests, but I'd much prefer for this to be a part of happy-dom

kayahr:
Ran into the same problem while testing a new web framework with happy-dom.

This code triggers the popstate event handler in a real browser, but not in happy-dom:

```js
window.addEventListener("popstate", () => console.log("Navigated to:", window.location.href));
window.location.href = "#/test";
```

Using this as a workaround for now:

```js
const origHref = Object.getOwnPropertyDescriptor(Object.getPrototypeOf(window.location), "href");
Object.defineProperty(window.location, "href", {
    get() {
        return origHref.get.call(this);
    },
    set(value) {
        if (value !== origHref.get.call(this)) {
            origHref.set.call(this, value);
            window.dispatchEvent(new window.PopStateEvent("popstate"));
        }
    }
});
```

Not sure if the PopStateEvent is created correctly, in my tests the event content is not used.

capricorn86:
Thank you for your reporting @taraldr and @kayahr! 🙂 

Sorry that you had to wait for such a long time before I had time to look into this.

I believe that this has already been implemented in [v19.0.0](https://github.com/capricorn86/happy-dom/releases/tag/v19.0.0). The "popstate" event should only fire for history entities that where navigated to using `pushState()`.

I will close this issue now. Please let me know if I missed something.

PR #1873, opened 2025-08-16
feat: [#1846] Dispatch popstate event on supported History actions

ahme-dev:
This will implement #1846. 

I also have a [project](https://github.com/ahme-dev/htswap) that acts on popstate events for browser back/forward actions, and its tests fail as something like `window.history.back()` doesn't trigger an event.

### DIFF
--- a/packages/happy-dom/src/history/History.ts
+++ b/packages/happy-dom/src/history/History.ts
@@ -90,11 +90,29 @@ export default class History {
 	}
 
 	/**
+	 * Dispatches a popstate event.
+	 *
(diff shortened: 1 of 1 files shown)


capricorn86:
Thank you for your contribution @ahme-dev! :star: 

Sorry that you had to wait for such a long time before I had time to look into this.

I believe that this has already been implemented in [v19.0.0](https://github.com/capricorn86/happy-dom/releases/tag/v19.0.0). The "popstate" event should only fire for history entities that where navigated to using `pushState()`.

I will close this PR now.

Issue #1541, opened 2024-09-17
Solving for flaky tests both in CI and locally


mikedidomizio:
**Is your feature request related to a problem? Please describe.**

There seems to be a number of flaky tests in this project [as you're aware](https://github.com/capricorn86/happy-dom/pull/1533#issuecomment-2344371198). I think stability is important here, both for the author but also as a contributor.

**Describe the solution you'd like**

- GitHub 30 seconds - [Link](https://github.com/capricorn86/happy-dom/actions/runs/10833610399/job/30060815178#step:11:101) - 5 days ago
- GitHub 30 seconds - [Link](https://github.com/capricorn86/happy-dom/actions/runs/10800944742/job/29960034001#step:11:101) - last week
- GitHub 30 seconds - [Link](https://github.com/capricorn86/happy-dom/actions/runs/10645757890/job/29511877480#step:11:97) - 2 weeks ago
- GitHub 30 seconds - [Link](https://github.com/capricorn86/happy-dom/actions/runs/10641553101/job/29502850004#step:11:98) - 2 weeks ago
- This one it timed out I guess not due to the speed of loading, but maybe an issue on what GitHub loaded - [Link](https://github.com/capricorn86/happy-dom/actions/runs/10619149426/job/29436059185#step:11:288) - 3 weeks ago

## Navigates the "_top" frame of a detached browser

- [Link](https://github.com/capricorn86/happy-dom/actions/runs/10832428921/job/30056757103?pr=1536) - 5 days ago
- [Link](https://github.com/capricorn86/happy-dom/actions/runs/10880195266/job/30186454900#step:11:302) - yesterday

## `Returns the ISO week number` and other date tests.

Date tests fail locally for me.

![CleanShot 2024-09-17 at 06 12 49@2x](https://github.com/user-attachments/assets/8297b779-1338-49b7-bb29-c03371438b8f)

The reason behind this is because I'm in Eastern Canada with a Timezone of GMT-2.
I think you're in Sweden, which is GMT+2.
GitHub action runners are unknown... I can't seem to get a clear answer. I think they're UTC+0.

If you add:

```diff
+    beforeAll(() => {
+	    process.env.TZ = 'America/Toronto'
+    })

describe('HTMLInputElementDateUtility', () => {
```

You'll probably witness that these tests fail now.

If I go with

```diff
+    beforeAll(() => {
+	    process.env.TZ = 'Etc/GMT-2'
+    })

[...]

PR #1879, opened 2025-08-24
chore: [#1541] Set time zone when running unit tests

TrevorBurnham:
I pulled down the repo to contribute a patch and noticed that two of the tests immediately failed on my device here in the US. I found #1541, which described the same issue and suggested this fix. Setting `TZ` when you have tests involving dates is, I think, fairly standard.

✅ With this change, all tests pass on my machine.


TrevorBurnham:
@capricorn86 I've updated this branch and all checks are now passing.

capricorn86 (approved):

Sorry that you had to wait for such a long time before I had time to look into this.

PR #1881, opened 2025-08-24
fix: [#1880] Replace setTimeout with queueMicrotask where appropriate

### LINKED ISSUE #1880: Replace setTimeout with queueMicrotask where possible
There are a few places where `setTimeout` is used for an arbitrarily short delay. These could be converted to `queueMicrotask` to reduce idle time.

In the case of the [`setTimeout` in `MutationObserverListener`](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/mutation-observer/MutationObserverListener.ts#L66), this change would also make the behavior more consistent with the DOM spec, which states that mutation observer callbacks are executed as microtasks.

TrevorBurnham:
This commit replaces several instances of `setTimeout(..., 0)` with `queueMicrotask` to improve performance and align more closely with browser behavior for microtask scheduling.

The changes are:

- In `MutationObserverListener`, mutation observer callbacks are now scheduled with `queueMicrotask`, as specified by the DOM standard.
- In `Fetch`, the creation of a `ReadableStream` from a buffer now uses `queueMicrotask` to make the operation asynchronous more efficiently.

I've ignored the `setTimeout` in `@packages/happy-dom/src/window/BrowserWindow.ts` where it's being used to simulate smooth scrolling, since in that case the macrotask scheduling is necessary to accurately reflect DOM behavior.

Resolves #1880

### DIFF
--- a/integration-test/package.json
+++ b/integration-test/package.json
@@ -15,7 +15,7 @@
 	},
 	"devDependencies": {
 		"happy-dom": "^0.0.0",
-		"@types/node": "^20.0.0",
+		"@types/node": ">=20.0.0",
 		"express": "^4.20.0"
 	},
 	"engines": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
 			],
 			"devDependencies": {
 				"@types/he": "^1.1.2",
-				"@types/node": "^20.0.0",
+				"@types/node": ">=20.0.0",
 				"@typescript-eslint/eslint-plugin": "^6.19.0",
 				"@typescript-eslint/parser": "^6.19.0",
 				"chalk": "^4.1.0",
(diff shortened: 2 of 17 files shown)


TrevorBurnham:
@capricorn86 I've updated this branch. All checks are passing.

capricorn86:
@TrevorBurnham what I'm a bit worried about is the functionality in `AsyncTaskManager` which supports aborting all ongoing tasks. It's not possible to abort micro-tasks, so in order to know that the Window has been fully closed, it will have to wait for the micro-tasks to finish. I'm also not sure that it will be much of a performance improvement.

This might not be a problem and the tests seem to be passing. Another option would be to use `setImmediate()`, which can be used by calling `requestAnimationFrame()`.

What's your thoughts on this?

TrevorBurnham:
@capricorn86 I looked over all of the changes here. Regarding `postMessage`, I had it wrong: The HTML spec says that `postMessage` should "queue a global task" (macrotask), not a microtask. So I've reverted it back to `setTimeout` to stay spec-compliant.

But as far as the others, they're true microtasks, so the behavior here better matches the spec. I don't believe there's any need to be able to have abortable handlers here. The `MutationObserver` callback checks the `#destroyed` flag, so it won't run if the observer has been cleaned up. And the `Fetch` microtask just enqueues a buffer and closes immediately; it's not a long-running operation.

I've updated the PR description and rebased the branch.

capricorn86 (dismissed):

PR #1885, opened 2025-08-24
feat: [#1884] Use classList.contains instead of splitting className

### LINKED ISSUE #1884: classList is parsed with regex when .contains() could be used
In the `matchClass` function in `SelectorItem`, a [regex is used to split an element's `className` attribute](https://github.com/capricorn86/happy-dom/blob/cf74f5f63ca562e075c9c14b77ecfbb8fbc43dea/packages/happy-dom/src/query-selector/SelectorItem.ts#L433) in order to check whether that element has a particular class. It should be more efficient to just use `classList.contains`.

TrevorBurnham:
Resolves #1884

This change should yield a small performance improvement.

Note that the splitting logic was introduced by https://github.com/capricorn86/happy-dom/pull/386, which was a bugfix. However, the tests introduced by that PR (which should cover the issue it describes) still pass with `contains`.


capricorn86 (approved):

Issue #1886, opened 2025-08-24
StringUtility uses a loop when it could use a RegExp


TrevorBurnham:
`StringUtility` [loops over all characters in a string](https://github.com/capricorn86/happy-dom/blob/33a72ca38277d2aab5341b4b66b0ca6380285c0b/packages/happy-dom/src/StringUtility.ts#L21-L28) to convert ASCII letters to the desired casing. A RegExp should be more efficient and succinct.

PR #1887, opened 2025-08-24
feat: [#1886] Use RegExp to convert ASCII character casing

TrevorBurnham:
This PR converts the loops used to convert ASCII character casing in `StringUtils` to regular expressions, which should be more efficient. I've also added unit tests to prevent regressions.

As an aside, it would be interesting to do some benchmarks to see how much benefit the caches have here. I'm sure a cache lookup is a bit faster than the regex, but the tradeoff is that the caches could potentially consume quite a bit of extra memory in a DOM with lots of custom tags and attributes.

### DIFF
--- a/packages/happy-dom/src/utilities/StringUtility.ts
+++ b/packages/happy-dom/src/utilities/StringUtility.ts
@@ -1,5 +1,7 @@
 const ASCII_LOWER_CASE_CACHE: Map<string, string> = new Map();
 const ASCII_UPPER_CASE_CACHE: Map<string, string> = new Map();
+const ASCII_UPPER_CASE_REGEXP = /[A-Z]/g;
+const ASCII_LOWER_CASE_REGEXP = /[a-z]/g;
 
 /**
  * String utility.
@@ -17,15 +19,9 @@ export default class StringUtility {
 		if (cached) {
 			return cached;
 		}
-		let newText = '';
-		for (const char of text) {
-			const value = char.charCodeAt(0);
-			if (value >= 65 && value <= 90) {
-				newText += String.fromCharCode(value + 32);
-			} else {
-				newText += char;
-			}
-		}
+		const newText = text.replace(ASCII_UPPER_CASE_REGEXP, (char) =>
+			String.fromCharCode(char.charCodeAt(0) + 32)
+		);
 		ASCII_LOWER_CASE_CACHE.set(text, newText);
 		return newText;
 	}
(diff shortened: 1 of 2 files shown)


Satya900:
<html>
<body>
<html><head></head><body><p>Here’s the same explanation in clean <strong>Markdown (MD)</strong> format, ready for a GitHub issue, PR comment, or build report:</p>
<hr>
<h2>🧩 Build Failure Summary</h2>
<h3>❌ Failed Tests</h3>
<p>Two tests in <code inline="">test/utilities/StringUtility.test.ts</code> failed:</p>
<ol>
<li>
<p><strong><code inline="">asciiLowerCase()</code> → leaves non-ASCII characters unchanged</strong></p>
</li>
<li>
<p><strong><code inline="">asciiUpperCase()</code> → leaves non-ASCII characters unchanged</strong></p>
</li>
</ol>
<hr>
<h3>🔍 What Happened</h3>
<p>The tests expect <code inline="">asciiLowerCase()</code> and <code inline="">asciiUpperCase()</code> to modify <strong>only ASCII letters (A–Z, a–z)</strong> and leave <strong>non-ASCII characters</strong> (like accented letters or Unicode symbols) unchanged.</p>
<p>However, the current implementation is incorrectly altering or skipping some non-ASCII characters.</p>

Function | Input | Expected Output | Actual Output
-- | -- | -- | --
asciiLowerCase() | "HéLLÖ" | "héllö" | "héllÖ"
asciiUpperCase() | "hélLö" | "HéLLÖ" | "HéLLö"


<p>The mismatch shows that:</p>
<ul>
<li>
<p>In the lowercase test, the final <code inline="">Ö</code> wasn’t converted as expected.</p>
</li>
<li>
<p>In the uppercase test, the final <code inline="">ö</code> wasn’t converted as expected.</p>
</li>
</ul>
<hr>
<h3>⚙️ Why It Matters</h3>
<p>These tests ensure that ASCII-specific utilities do <strong>not modify extended Unicode characters</strong>.<br>
Accidental changes here could cause incorrect behavior in multi-language environments or with symbols beyond ASCII.</p>
<hr>
<h3>🧠 Likely Cause</h3>
<p>The functions may be using JavaScript’s built-in:</p>
<pre><code class="language-ts">String.toLowerCase()
String.toUpperCase()
</code></pre>
<p>These methods are <strong>Unicode-aware</strong>, not ASCII-only — they transform all characters, including non-ASCII ones, which violates the test expectation.</p>
<hr>
<h3>🛠️ Suggested Fix</h3>
<p>Restrict transformations strictly to the ASCII character range.<br>
Here’s an example implementation:</p>
<pre><code class="language-ts">static asciiLowerCase(str: string): string {
  return str.replace(/[A-Z]/g, c =&gt; String.fromCharCode(c.charCodeAt(0) + 32));
}

static asciiUpperCase(str: string): string {
  return str.replace(/[a-z]/g, c =&gt; String.fromCharCode(c.charCodeAt(0) - 32));
}
</code></pre>
<p>This ensures:</p>
<ul>
<li>
<p>Only ASCII letters are modified.</p>
</li>
<li>
<p>Non-ASCII characters remain untouched.</p>
</li>
</ul>
<hr>
<h3>✅ Summary</h3>
<ul>
<li>
<p><strong>Issue:</strong> Non-ASCII characters were altered or left inconsistent.</p>
</li>
<li>
<p><strong>Fix:</strong> Limit transformations to ASCII character codes only.</p>
</li>
<li>
<p><strong>Result:</strong> Both <code inline="">asciiLowerCase()</code> and <code inline="">asciiUpperCase()</code> tests should pass once corrected.</p>
</li>
</ul>
<hr></body></html>
</body>
</html>

PR #1889, opened 2025-08-25
fix: [#1888] Replaces inefficient loops for clearing children

### LINKED ISSUE #1888: Child nodes are cleared using an O(n^2) loop
In several different parts of the codebase, all children are removed from a parent using a loop like this:

```js
while (childNodes.length) removeChild(childNodes[0])
```

Since each removal invokes `splice(0, 1)` on the child array, you pay the price of shifting every remaining child on each loop iteration. Hence, the number of array-shifting operations is `O(n^2)` with respect to the number of children.

TrevorBurnham:
Fixes #1888

This PR introduces a fast path for clearing all children from a parent node in `O(n)` time and wires it into hot code paths (`innerHTML`, `textContent`, `replaceChildren`, `ShadowRoot.setHTMLUnsafe`, and `Document.open`). It also optimizes cleanup in `Node#[PropertySymbol.cloneNode]` using the same technique. The change improves spec compliance for `MutationObserver` semantics while significantly reducing time and allocations for large subtree clears.

## Motivation

Several core DOM APIs in happy-dom clear a node by repeatedly calling `removeChild(childNodes[0])` in a loop. Each removal from index 0 shifts the array, leading to `O(n²)` behavior and extra GC pressure for large trees. This PR replaces those loops with a single-pass algorithm.

### New utility: `ParentNodeUtility.clearChildren(parent)`

- Detaches and cleans all children in a single sweep.
- Preserves `NodeList` reference identity by truncating arrays (`length = 0`) rather than reassigning them.
- Clears `elementArray` in one step.
- Ensures correct lifecycle transitions (connected/disconnected) and cache invalidations.
- Unobserves subtree mutation observers from removed subtrees as before.

### `Node#[PropertySymbol.cloneNode]` cleanup

Replaces the previous `while (childNodes.length) removeChild(childNodes[0])` pattern with a local fast-detach that snapshots children, truncates arrays (`length = 0`), and then clears parent links/cache and dispatches disconnects per child.

## MutationObserver behavior change (spec compliance)

This PR changes `MutationObserver` behavior to match the DOM specification's ["replace all" algorithm](https://dom.spec.whatwg.org/#concept-node-replace-all).

**Before:** When clearing children (e.g., setting `innerHTML = ''`), happy-dom emitted one `MutationRecord` per removed child node:

```js
// 3 children removed → 3 separate MutationRecords
[
  { removedNodes: [a], nextSibling: b },
  { removedNodes: [b], nextSibling: c },
  { removedNodes: [c], nextSibling: null }
]
```

**After:** A single `MutationRecord` is emitted containing all removed nodes, matching browser behavior per the DOM spec:

```js
// 3 children removed → 1 MutationRecord
[
  { removedNodes: [a, b, c], previousSibling: null, nextSibling: null }
]
```

This is both more spec-compliant and more efficient (one record allocation instead of N).

## Other improvements

- **Type safety:** Reduced `<any>` casts by using proper TypeScript types for internal properties (`IMutationListener`, `Node` properties)
- **Edge case handling:** `clearChildren` returns early with no mutation emitted when the parent has no children
- **Operation order fix:** Ensured `nodeArray` is cleared before calling `disconnectedFromNode()` on removed nodes, matching the order of operations in `removeChild`. This fixes a timing issue that could cause test failures on Node.js 20.

### DIFF
--- a/packages/happy-dom/src/nodes/document-fragment/DocumentFragment.ts
+++ b/packages/happy-dom/src/nodes/document-fragment/DocumentFragment.ts
@@ -84,10 +84,7 @@ export default class DocumentFragment extends Node {
 	 * @param textContent Text content.
 	 */
 	public set textContent(textContent: string) {
-		const childNodes = this[PropertySymbol.nodeArray];
-		while (childNodes.length) {
-			this.removeChild(childNodes[0]);
-		}
+		ParentNodeUtility.clearChildren(this);
 		if (textContent) {
 			this.appendChild(this[PropertySymbol.ownerDocument].createTextNode(textContent));
 		}
--- a/packages/happy-dom/src/nodes/document/Document.ts
+++ b/packages/happy-dom/src/nodes/document/Document.ts
@@ -1723,10 +1723,7 @@ export default class Document extends Node {
 			}
 		}
 
-		const childNodes = this[PropertySymbol.nodeArray];
-		while (childNodes.length) {
-			this.removeChild(childNodes[0]);
-		}
+		ParentNodeUtility.clearChildren(this);
 
 		// Default document elements
 		const doctype = this[PropertySymbol.implementation].createDocumentType('html', '', '');
--- a/packages/happy-dom/src/nodes/element/Element.ts
+++ b/packages/happy-dom/src/nodes/element/Element.ts
@@ -401,10 +401,7 @@ export default class Element
 	 * @param textContent Text content.
 	 */
 	public set textContent(textContent: string) {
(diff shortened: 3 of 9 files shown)


TrevorBurnham:
@capricorn86 I've rebased this PR. It's ready for review.

capricorn86:
I would like to look into this PR a bit more and see if using a `[PropertySymbol.clearChildren]()` method that is overriden for each sub-class would make it easier to maintain. I can try it locally and see when I have a bit more time.

PR #1891, opened 2025-09-03
fix: [#1890] Fixes XMLHttpRequest `responseText` on partial response

### LINKED ISSUE #1890: XMLHttpRequest `responseText` not available until request complete
**Describe the bug**
While a request to a XMLHttpRequest is being received, the `responseText` property can be read. This is a common approach to implement streaming responses. This does _not_ however work in happy-dom - it will simply return an empty string until the whole response is received.

**To Reproduce**
Add the following (failing) test to the `XMLHttpRequest` test suite:

```ts
it('Should return responseText with correct length during streaming in progress event', async () => {
  await new Promise((resolve, reject) => {
    const responseText = 'This is a streaming response';

    vi.spyOn(Fetch.prototype, 'send').mockImplementation(
      async () =>
        <Response>{
          headers: <Headers>new Headers({
            'Content-Length': String(responseText.length)
          }),
          body: new ReadableStream({
            start(controller) {
              controller.enqueue(responseText);
              controller.close();
            }
          })
        }
    );

    request.open('GET', REQUEST_URL, true);

    let progressTriggered = false;

    request.addEventListener('progress', (event) => {
      progressTriggered = true;
      const progressEvent = <ProgressEvent>event;
      try {
        expect(request.responseText.length).toBe(progressEvent.loaded);
        expect(request.responseText).toBe(responseText);
        resolve(null);
      } catch (error) {
        reject(error);
      }
    });

    request.addEventListener('load', () => {
      if (!progressTriggered) {
        reject(new Error('Progress event was never triggered'));
      }
    });

    request.send();
  });
});
```

**Expected behavior**
Once a `progress` event has been received, reading `responseText` should give the currently received response body text.

**Additional context**

rexxars:
Fixes #1890.

Instead of accumulating data to a local variable within `sendAsync`, we move it to a private class property that can be read from `responseText`. Once the response is complete, we reuse and return the complete value - but while streaming the response, we can access the partial response by concatenating the received buffers.

### DIFF
--- a/packages/happy-dom/src/xml-http-request/XMLHttpRequest.ts
+++ b/packages/happy-dom/src/xml-http-request/XMLHttpRequest.ts
@@ -49,6 +49,7 @@ export default class XMLHttpRequest extends XMLHttpRequestEventTarget {
 	#response: Response | ISyncResponse | null = null;
 	#responseType: XMLHttpResponseTypeEnum | '' = '';
 	#responseBody: ArrayBuffer | Blob | Document | object | string | null = null;
+	#accumulatedData: Buffer = Buffer.from([]);
 	#readyState: XMLHttpRequestReadyStateEnum = XMLHttpRequestReadyStateEnum.unsent;
 	#overriddenMimeType: string | null = null;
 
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Thank you for your contribution @rexxars! :star:

Issue #1928, opened 2025-10-08
Support `<details>` click firing the `onToggle` event handler


matthew-gladman-oua:
**Is your feature request related to a problem? Please describe.**
We use [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details) quite a bit in our web app. Currently `happy-dom` does not appear to trigger the `onToggle` when the element is clicked. This works fine in `jsdom`

**Describe the solution you'd like**
When clicking a `<details>` element, the `onToggle` event handler should fire

**Describe alternatives you've considered**
Current alternative is to run tests that require this functionality to use `jsdom`. Example for vitest:

```ts
// @vitest-environment jsdom

// ... imports and test code
```

EDIT: I can see this unit test, which indicates it **should** work: 
- https://github.com/capricorn86/happy-dom/blob/c80a08f30ad97b04fbb251ab11b87cb9d5706207/packages/%40happy-dom/jest-environment/test/testing-library/TestingLibrary.test.tsx#L82

----

EDIT 2: I think I found the problem space:

```ts
// ❌ doesn't work
it('should work with details and summary in span', () => {
  const onToggle = vi.fn();
  render(
    <details onToggle={onToggle} open={true}>
      <summary>
        <span>test</span>
      </summary>
    </details>,
  );
  expect(screen.getByRole('group')).toHaveAttribute('open');
  fireEvent.click(screen.getByText('test'));

  expect(screen.getByRole('group')).not.toHaveAttribute('open');
});

// ✅ works, but is invalid html
it('should work with details and raw summary', () => {
  const onToggle = vi.fn();
  render(
    <details onToggle={onToggle} open={true}>
      <summary>
        test
      </summary>
    </details>,
  );
  expect(screen.getByRole('group')).toHaveAttribute('open');
  fireEvent.click(screen.getByText('test'));

  expect(screen.getByRole('group')).not.toHaveAttribute('open');
});

// ✅ also works, but is not quite correct behavior
it('should work with details grab parent node', () => {
  const onToggle = vi.fn();
  render(
    <details onToggle={onToggle} open={true}>
      <summary>
        <span>test</span>
      </summary>
    </details>,
  );

  screen.debug(screen.getByRole('group', { hidden: true }));
  expect(screen.getByRole('group')).toHaveAttribute('open');
  fireEvent.click(screen.getByText('test').parentNode!); // clicking the <summary> instead of the child element

[...]

PR #1929, opened 2025-10-09
fix: [#1928] Support details click firing the onToggle event handler

Nxooah:
- Fixed HTMLDetailsElement.dispatchEvent() to handle clicks on descendant elements of summary
- Now uses Element.closest() to find the nearest summary element in the DOM hierarchy
- Added comprehensive tests to verify the fix works for nested elements
- Resolves issue where clicking on child elements (like spans) inside summary didn't toggle details


capricorn86 (approved):
Thank you for your contribution @Nxooah! :star: 

I made a fix to improve performance for checking for the summary element.

PR #1944, opened 2025-10-28
feat: [#0] Adds support for rendering HTML strings

capricorn86:
- Adds support for rendering HTML strings to `ServerRenderer` in the `@happy-dom/server-renderer` package
   - It previously only had support for rendering URLs
- Multiple fixes that was related to adding support adding server-rendering hooks to frameworks


Spixmaster:
Can you explain what you mean with rendering HTML strings? Where is it needed?

capricorn86:
> Can you explain what you mean with rendering HTML strings? Where is it needed?

I have updated the description of the release notes and the PR now. It was a bit unclear what it meant by rendering strings. You can see an example for this under "Server-Side Rendering of Web Components in SvelteKit" under [Server-Renderer](https://github.com/capricorn86/happy-dom/wiki/Server-Renderer) in the Wiki.

Issue #1947, opened 2025-10-29
XMLEncodeUtility.decodeHTMLAttributeValue not spec compliant


cgoetz-inovex:
**Describe the bug**
https://github.com/capricorn86/happy-dom/blob/2cc044385bf53c8afc83aa9d1d1fb9735555680b/packages/happy-dom/src/utilities/XMLEncodeUtility.ts#L65 only replaces named character references `&quot;` and `&amp;`. The [spec for attribute values](https://html.spec.whatwg.org/multipage/syntax.html#syntax-attribute-value) specifies a mixture of text and [character references](https://html.spec.whatwg.org/multipage/syntax.html#syntax-charref), which can also contain (hexa)decimal numeric character references like `&#34;`.

**To Reproduce**
apply patch and execute new test:
```
Subject: [PATCH] add test for numeric character reference
---
Index: packages/happy-dom/test/html-parser/HTMLParser.test.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/packages/happy-dom/test/html-parser/HTMLParser.test.ts b/packages/happy-dom/test/html-parser/HTMLParser.test.ts
--- a/packages/happy-dom/test/html-parser/HTMLParser.test.ts	(revision 2cc044385bf53c8afc83aa9d1d1fb9735555680b)
+++ b/packages/happy-dom/test/html-parser/HTMLParser.test.ts	(date 1761741688049)
@@ -2208,5 +2208,14 @@
                 </tr>
             `);
 		});
+
+        it('Handles numeric character references for #tbd', () => {
+            const result = new HTMLParser(window).parse(
+                `<div data-foo="&#34;"></div>`
+            );
+            expect(new HTMLSerializer().serializeToString(result)).toBe(
+                `<div data-foo="&quot;"></div>`
+            );
+        });
 	});
 });
```
test fails with:
```
AssertionError: expected '<div data-foo="&amp;#34;"></div>' to be '<div data-foo="&quot;"></div>' // Object.is equality
Expected :<div data-foo="&quot;"></div>
Actual   :<div data-foo="&amp;#34;"></div>
```

**Expected behavior**
Test should succeed.

**Device:**
 - OS: macOS Tahoe 26.0.1
 - Bun version 1.2.19
 - 2cc044385bf53c8afc83aa9d1d1fb9735555680b

**Additional context**
[...]

PR #1948, opened 2025-10-30
fix: [#1947] Decode &apos; / numeric character references in attr values

cgoetz-inovex:
- attribute values can contain named and numeric character references per https://html.spec.whatwg.org/multipage/syntax.html#syntax-attribute-value
- add decoding for all numeric character references
- add decoding for &apos;, this could probably appear in single-quoted attribute values
- ignore other named character references, the list is huge


cgoetz-inovex:
Regarding the failed build: works locally, would appreciate some support/guidance.

capricorn86 (approved):

capricorn86:
Thank you for contributing @cgoetz-inovex! :star: 

I believe that #2016 solved the issue, so I will only merge the unit test.

PR #1960, opened 2025-12-01
fix: [#1955] Fixes stepUp and stepDown on HTMLInputElement according to spec

### LINKED ISSUE #1955: HTMLInputElement "number" type does not respect min and max settings.
**Describe the bug**
Looking here, there is no way to limit value stepping based on min and max values:
https://github.com/capricorn86/happy-dom/blob/2cc044385bf53c8afc83aa9d1d1fb9735555680b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElementValueStepping.ts#L16-L26

These values are also not limited in the `stepUp`/`stepDown` functions here:
https://github.com/capricorn86/happy-dom/blob/2cc044385bf53c8afc83aa9d1d1fb9735555680b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts#L1367-L1384

**To Reproduce**
Attempt `stepUp` or `stepDown` on an input with this markup: `<input type="number" min="5" max="5" value="5" />`

**Expected behavior**
The input's value should not change because it is limited by its min and max attributes, but it does change.

stevematney:
Fixes #1955

Referencing spec from: https://html.spec.whatwg.org/multipage/input.html#common-input-element-apis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1488,9 +1488,9 @@
 			}
 		},
 		"node_modules/@istanbuljs/load-nyc-config/node_modules/js-yaml": {
-			"version": "3.14.2",
-			"resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.14.2.tgz",
-			"integrity": "sha512-PMSmkqxr106Xa156c2M265Z+FTrPl+oxd/rgOQy2tijQeK5TxQ43psO1ZCwhVOSdnn+RzkzlRz/eY4BgJBYVpg==",
+			"version": "3.14.1",
+			"resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.14.1.tgz",
+			"integrity": "sha512-okMH7OXXJ7YrN9Ok3/SXrnu4iX9yOk+25nqX4imS2npuvTYDmo/QEZoqwZkYaIDk3jVvBOTOIEgEhaLOynBS9g==",
 			"dev": true,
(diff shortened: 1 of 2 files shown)


stevematney:
@capricorn86 just thinking about this, and wondering if I should rename the commit for this with `feat` rather than `fix`. It's technically adding new functionality, even though it is addressing the issue I reported.

capricorn86:
> @capricorn86 just thinking about this, and wondering if I should rename the commit for this with `feat` rather than `fix`. It's technically adding new functionality, even though it is addressing the issue I reported.

The merge is done with squash nowadays, so it's based on that commit message. I think this could be considered a "fix" as there was an implementation, but it was incorrect or too simple.

stevematney:
> I have fixed the implementation, so that it works according to spec. It should step based on the next valid step value. The valid step is calculated from the "min" value (see example below).

Oh, that makes sense! That was an oversight on my part. Thanks for fixing it! I appreciate you getting this done and merged in!

capricorn86:
> > I have fixed the implementation, so that it works according to spec. It should step based on the next valid step value. The valid step is calculated from the "min" value (see example below).
> 
> Oh, that makes sense! That was an oversight on my part. Thanks for fixing it! I appreciate you getting this done and merged in!

Easy to miss! It took some time for me to understand how it where suppose to work and the spec was really hard to interpret :sweat_smile:

PR #1969, opened 2025-12-31
fix: Node.replaceWith does not throw w/o parent

### LINKED ISSUE #1915: Element#replaceWith throws when parent element doesn't exist
**Describe the bug**
When I call `elementA.replaceWith(elementB)` and `elementA` doesn't have a parent element then this call throws. If I read the [DOM Standard](https://dom.spec.whatwg.org/#ref-for-dom-childnode-replacewith%E2%91%A0) correctly then in this situation step 2 ("If _parent_ is null, then return.") should cause the function to return early without throwing.

The error thrown:
```
> document.createElement('div').replaceWith(document.createElement('div'))
Uncaught:
DOMException: This element has no parent node.
    at ChildNodeUtility.replaceWith (file:///C:/dev/foo/node_modules/happy-dom/lib/nodes/child-node/ChildNodeUtility.js:27:19)
    at HTMLDivElement.replaceWith (file:///C:/dev/foo/node_modules/happy-dom/lib/nodes/element/Element.js:477:26)
```

**To Reproduce**
```ts
import {Window} from 'happy-dom'

const {document} = new Window({url: 'https://dummy'})
document.createElement('div').replaceWith(document.createElement('div'))
```

**Expected behavior**
No error

**Screenshots**
n/a

**Device:**
n/a

**Additional context**
Reproduced with happy-dom@19.0.2

lukeed:
Removes invalid `DOMException` -- browsers do not throw on missing parent

```js
let div = document.createElement("div");
let span = document.createElement("span");

div.parentNode; // null
span.parentNode; // null

div.replaceWith(span); // ok, noop
```

Other methods in this file also early-return

> Closes #1915

### DIFF
--- a/packages/happy-dom/test/nodes/element/Element.test.ts
+++ b/packages/happy-dom/test/nodes/element/Element.test.ts
@@ -1203,6 +1203,10 @@ describe('Element', () => {
 			element.replaceWith(node1, node2);
 			expect(isCalled).toBe(true);
 		});
+
+		it('Should not throw when there is no parent node.', () => {
+			document.createElement('div').replaceWith(document.createElement('div'));
+		});
 	});
 
 	describe('before()', () => {


lukeed:
I believe the failing test is unrelated @capricorn86 do you want me to do anything else here?

capricorn86 (approved):
Thank you for your contribution @lukeed! ⭐ 

I added a unit test.

PR #1970, opened 2026-01-01
chore: upgrade vitest

btea:
Upgrade Vitest version to 4


capricorn86 (approved):

Issue #1978, opened 2026-01-16
Bug: HTML entities for emoji (U+1F000-U+1FFFF) are incorrectly parsed, stripping the leading "1"


ke-at-microsoft:
## Summary

happy-dom v20.0.7 incorrectly parses HTML entities for emoji and other characters in the Unicode supplementary plane (code points U+1F000 to U+1FFFF). The parser strips the leading "1" from the hex representation, resulting in incorrect characters.

## Environment

- **happy-dom version**: 20.0.7 (also tested with v20.3.1 - bug still exists)
- **Node.js version**: Node.js v18+
- **Operating System**: Windows/Linux/macOS (platform-independent)

## Expected Behavior

HTML numeric character references for emoji should decode to the correct Unicode code points:

| Input (Decimal) | Input (Hex) | Expected Output | Expected Code Point |
|----------------|-------------|-----------------|-------------------|
| `&#128202;` | `&#x1F4CA;` | 📊 | U+1F4CA |
| `&#128187;` | `&#x1F4BB;` | 💻 | U+1F4BB |
| `&#127873;` | `&#x1F381;` | 🎁 | U+1F381 |
| `&#128512;` | `&#x1F600;` | 😀 | U+1F600 |

## Actual Behavior

The parser strips the leading "1" from code points in the 0x1F000-0x1FFFF range:

| Input (Decimal) | Input (Hex) | Actual Output | Actual Code Point | Error |
|----------------|-------------|---------------|------------------|-------|
| `&#128202;` | `&#x1F4CA;` |  | U+F4CA | ❌ Missing "1" |
| `&#128187;` | `&#x1F4BB;` |  | U+F4BB | ❌ Missing "1" |
| `&#127873;` | `&#x1F381;` |  | U+F381 | ❌ Missing "1" |
| `&#128512;` | `&#x1F600;` |  | U+F600 | ❌ Missing "1" |

This results in displaying incorrect characters from the Private Use Area (U+F000-U+FFFF) instead of the intended emoji.

### Test Script

```javascript
const { Window } = require('happy-dom');

const window = new Window();
const parser = new window.DOMParser();

// Test emoji HTML entity parsing
const testCases = [
  { decimal: '&#128202;', hex: '&#x1F4CA;', expected: 0x1F4CA, name: '📊 Chart' },
  { decimal: '&#128187;', hex: '&#x1F4BB;', expected: 0x1F4BB, name: '💻 Laptop' },
[...]

PR #1979, opened 2026-01-16
fix: [#1978] Use String.fromCodePoint() to decode numeric HTML/XML character entities.

ke-at-microsoft:
## Root Cause ## 

The bug was in XMLEncodeUtility.ts, specifically in the decodeHTMLEntities() and decodeXMLEntities() methods. They were using String.fromCharCode() instead of String.fromCodePoint() to decode numeric HTML/XML character entities.

### The Problem ###

String.fromCharCode() only handles code points 0-65535 (Basic Multilingual Plane)
For emoji and other supplementary plane characters (U+10000+), it only uses the lower 16 bits
Example: &#x1F4CA; (📊) was incorrectly decoded because 0x1F4CA was treated as 0xF4CA

### Fix ### 

In XMLEncodeUtility.ts, changed String.fromCharCode to String.fromCodePoint in decodeHTMLEntities(), and changed String.fromCharCode to String.fromCodePoint in decodeXMLEntities().

### DIFF
--- a/packages/happy-dom/test/html-parser/HTMLParser.test.ts
+++ b/packages/happy-dom/test/html-parser/HTMLParser.test.ts
@@ -1975,9 +1975,7 @@ describe('HTMLParser', () => {
 
 		it('Decodes emoji HTML entities correctly for #1978', () => {
 			// Test hexadecimal emoji entities (supplementary plane characters U+1F000-U+1FFFF)
-			const result = new HTMLParser(window).parse(
-				'<div>&#x1F4CA; &#x1F600; &#x1F680;</div>'
-			);
+			const result = new HTMLParser(window).parse('<div>&#x1F4CA; &#x1F600; &#x1F680;</div>');
 			const div = result.childNodes[0];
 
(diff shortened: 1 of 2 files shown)


capricorn86:
Thank you for your contribution @ke-at-microsoft! :star: 

I believe that this was fixed in #1999. The fixes uses the "entities" package, which should do this correctly. Hopefully it will still be performant.

I will close this PR now. Let me know if I missed something.

ke-at-microsoft:
@capricorn86 using the "entities" package sounds good to me, it uses a polyfill of String.fromCodePoint for older runtime (https://github.com/fb55/entities/blob/844902b534ead1fbceb1b7be326d4be9431c43ba/src/decode-codepoint.ts#L38). I am not sure how many old runtimes we need to worry about (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint), most browsers support it. "entities" is a relatively small helper, probably won't cause much performance issue. thanks!

PR #1982, opened 2026-01-17
fix: [#1981] Remove redundant RegExp flags in HTMLParser

### LINKED ISSUE #1981: Performance: Remove redundant RegExp flags in HTMLParser
In HTMLParser.ts, the `MARKUP_REGEXP` and `ATTRIBUTE_REGEXP` constants are defined with the gm flags as part of the literal:

```ts
const MARKUP_REGEXP = /<([^\s/!>?]+)|...>/gm;
const ATTRIBUTE_REGEXP = /\s*([a-zA-Z0-9-_:.$@?\\<\[\]]+)...>/gm;
```

Later code then creates new RegExp instances from those constants with the same flags:

```ts
this.markupRegExp = new RegExp(MARKUP_REGEXP, 'gm');
const attributeRegexp = new RegExp(ATTRIBUTE_REGEXP, 'gm');
```

That needlessly duplicates the RegExp. We should be able to reuse the constants instead, slightly improving performance.

TrevorBurnham:
Fixes #1981

## Description

When creating a new RegExp from an existing RegExp literal, passing flags as a second argument is redundant since the literal already includes the flags. This PR removes the unnecessary `'gm'` flags parameter from the `new RegExp()` calls in `HTMLParser.ts`, allowing the regexp to be copied more efficiently.

It's still necessary to use `new RegExp()` to copy the RegExp constants here to avoid conflicting state.

### Before
```typescript
this.markupRegExp = new RegExp(MARKUP_REGEXP, 'gm');
const attributeRegexp = new RegExp(ATTRIBUTE_REGEXP, 'gm');
```

### After
```typescript
this.markupRegExp = new RegExp(MARKUP_REGEXP);
const attributeRegexp = new RegExp(ATTRIBUTE_REGEXP);
```

## Why this works

The `MARKUP_REGEXP` and `ATTRIBUTE_REGEXP` constants are already defined with the `gm` flags:

```typescript
const MARKUP_REGEXP = /<([^\s/!>?]+)|...>/gm;
const ATTRIBUTE_REGEXP = /\s*([a-zA-Z0-9-_:.$@?\\<\[\]]+)...>/gm;
```

When `new RegExp(existingRegExp, flags)` is called with a flags argument, JavaScript must parse the flags string and apply them. When called without flags (`new RegExp(existingRegExp)`), it simply copies the existing RegExp including its flags, which is slightly more efficient.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5542,7 +5542,6 @@
 			"version": "4.5.0",
 			"resolved": "https://registry.npmjs.org/entities/-/entities-4.5.0.tgz",
 			"integrity": "sha512-V0hjH4dGPh9Ao5p0MoRY6BVqtwCjhz6vI5LT8AJ55H+4g9/4vbHx1I54fS0XuclLhDHArPQCiMjDxjaL8fPxhw==",
-			"dev": true,
 			"license": "BSD-2-Clause",
 			"engines": {
 				"node": ">=0.12"
@@ -13877,6 +13876,7 @@
 				"@types/node": "^20.0.0",
 				"@types/whatwg-mimetype": "^3.0.2",
 				"@types/ws": "^8.18.1",
+				"entities": "^4.5.0",
 				"whatwg-mimetype": "^3.0.0",
 				"ws": "^8.18.3"
 			},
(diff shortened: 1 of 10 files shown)


TrevorBurnham:
The flaky test that's failing in CI should be addressed by #1986.

capricorn86 (approved):

PR #1986, opened 2026-01-17
fix: [#1985] Use parseFloat for Cache-Control max-age parsing

### LINKED ISSUE #1985: Flaky test: SyncFetch cache revalidation test fails intermittently in CI
I noticed that the tests failed in CI for #1982, but the test failure appears to be unrelated to the change. I believe the test in question, `Revalidates cache with a "If-Modified-Since" request for a GET response with "Cache-Control" set to a "max-age"`, is flaky, causing random failures like this:

```
AssertionError: expected { 'content-type': 'text/html', …(3) } to deeply equal { 'content-type': 'text/html', …(3) }

- Expected
+ Received

{
-   "Cache-Control": "max-age=1",
-   "Last-Modified": "Mon, 11 Dec 2023 02:00:00 GMT",
+   "cache-control": "max-age=0.0001",
    "content-length": "9",
    "content-type": "text/html",
+   "last-modified": "Mon, 11 Dec 2023 01:00:00 GMT",
}
```

The test sets max-age=0.0001 (0.1 milliseconds) and waits 100ms before making a second request, expecting the cache to be expired and revalidated. However, the second request returns the cached response instead of triggering a revalidation request with `If-Modified-Since`. This suggests a timing issue where the cache expiration check may not be working reliably in slower CI environments.

TrevorBurnham:
This PR fixes a bug in the response cache implementation where `parseInt()` was used to parse the `max-age` value from `Cache-Control` headers. This caused decimal values like `0.0001` to be truncated to `0`, leading to flaky test behavior.

Fixes #1985

## Problem

The cache implementation in `ResponseCache.ts` used `parseInt()` to parse max-age values:

```typescript
cachedResponse.expires = Date.now() + parseInt(value) * 1000 - (age ? parseInt(age) * 1000 : 0);
```

When `max-age=0.0001` was set, `parseInt('0.0001')` returned `0`, making the cache expire immediately at `Date.now()`. Due to timing variations, the strict less-than check (`entry.expires < Date.now()`) would sometimes fail when both timestamps were equal, causing the cache to return stale responses instead of triggering revalidation.

## Solution

Changed `parseInt()` to `parseFloat()` to properly handle decimal max-age values:

```typescript
cachedResponse.expires = Date.now() + parseFloat(value) * 1000 - (age ? parseFloat(age) * 1000 : 0);
```

This correctly calculates sub-second expiration times, which is valid per HTTP caching specifications.

### DIFF
--- a/packages/happy-dom/src/fetch/cache/response/ResponseCache.ts
+++ b/packages/happy-dom/src/fetch/cache/response/ResponseCache.ts
@@ -154,7 +154,7 @@ export default class ResponseCache implements IResponseCache {
 				switch (key) {
 					case 'max-age':
 						cachedResponse.expires =
-							Date.now() + parseInt(value) * 1000 - (age ? parseInt(age) * 1000 : 0);
+							Date.now() + parseFloat(value) * 1000 - (age ? parseFloat(age) * 1000 : 0);
 						break;
 					case 'no-cache':
 					case 'no-store':


capricorn86 (approved):

PR #1990, opened 2026-01-17
fix: [#1989] Optimize DOMTokenList add/remove with Set for O(1) lookups

### LINKED ISSUE #1989: Performance: Optimize DOMTokenList add/remove methods with Set for O(1) lookups
The `add()` and `remove()` methods in `DOMTokenList` currently use `Array.indexOf()` to check for token existence, which is `O(n)` per lookup. When adding or removing multiple tokens, this results in `O(n*m)` complexity where `n` is the list size and `m` is the number of tokens being added/removed.

TrevorBurnham:
Fixes #1989

## Summary

This PR optimizes the `add()` and `remove()` methods in `DOMTokenList` by using `Set` for O(1) token lookups instead of `Array.indexOf()` which is O(n).

### `add()` method

**Before:**
```typescript
for (const token of tokens) {
  const index = list.indexOf(token);  // O(n) per token
  if (index === -1) {
    list.push(token);
  }
}
```

**After:**
```typescript
const existingTokens = new Set(list);
for (const token of tokens) {
  if (!existingTokens.has(token)) {   // O(1)
    existingTokens.add(token);
    list.push(token);
  }
}
```

### `remove()` method

**Before:**
```typescript
for (const token of tokens) {
  const index = list.indexOf(token);  // O(n) per token
  if (index !== -1) {
    list.splice(index, 1);            // O(n) shift
  }
}
```

**After:**
```typescript
const tokensToRemove = new Set(tokens);
const newList = list.filter((item) => !tokensToRemove.has(item));  // O(n) total
```

## Complexity improvement

| Method | Before | After |
|--------|--------|-------|
| `add()` | O(n × m) | O(n + m) |
| `remove()` | O(n × m) | O(n + m) |

Where n = existing tokens, m = tokens being added/removed.

### DIFF
--- a/packages/happy-dom/src/dom/DOMTokenList.ts
+++ b/packages/happy-dom/src/dom/DOMTokenList.ts
@@ -254,10 +254,11 @@ export default class DOMTokenList {
 	 */
 	public add(...tokens: string[]): void {
 		const list = this[PropertySymbol.getTokenList]().slice();
+		const existingTokens = new Set(list);
 
 		for (const token of tokens) {
-			const index = list.indexOf(token);
-			if (index === -1) {
+			if (!existingTokens.has(token)) {
+				existingTokens.add(token);
 				list.push(token);
 			}
 		}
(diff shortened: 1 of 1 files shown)


capricorn86 (approved):

PR #1991, opened 2026-01-17
fix: [#1876] Use proxy in Node.contains() for HTMLFormElement/HTMLSelectElement

### LINKED ISSUE #1876: Node.contains is not functional with form/select elements
**Describe the bug**

I noticed that `<form>` and `<select>` elements do not properly identify child nodes using the `contains` API.

**To Reproduce**

I wrote a small script to demonstrate:

```js
const { Window } = require('happy-dom')

async function main() {
  const window = new Window()

  document.write(`
    <html>
      <body>
        <div id="works-with-div">
          <div class="parent">
            <input class="child" />
          </div>
        </div>

        <div id="does-not-work-with-form">
          <form class="parent">
            <input class="child" />
          </form>
        </div>

        <div id="does-not-work-with-select">
          <select class="parent">
            <option class="child" />
          </select>
        </div>
      </body>
    </html>
  `)

  for(const scenario of ['works-with-div', 'does-not-work-with-form', 'does-not-work-with-select']) {
    console.log(`\nTesting ${scenario}`)

    const parentElement = document.querySelector(`#${scenario} .parent`)
    const childElement = document.querySelector(`#${scenario} .child`)

    console.log(`Parent element ${parentElement.nodeName} exists`, !!parentElement)
    console.log(`Child element ${childElement.nodeName} exists`, !!childElement)
    console.log(`Parent element contains child element`, parentElement?.contains(childElement), '\n')
  }
}

main().catch(console.error)
```

This produces:

```txt
Testing works-with-div
Parent element DIV exists true
Child element INPUT exists true
Parent element contains child element true 


Testing does-not-work-with-form
Parent element FORM exists true
Child element INPUT exists true
Parent element contains child element false 


Testing does-not-work-with-select
Parent element SELECT exists true
Child element OPTION exists true
Parent element contains child element false 
```

**Expected behavior**

Child elements within form/select elements should be able to be identified using the `contains` API.

**Additional context**

I believe this relates to the proxies that are instantiated for HTMLFormElement and HTMLSelectElement. Within NodeUtility, the [isInclusiveAncestor](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/node/NodeUtility.ts#L73-L78) check loops through the parentNodes until the given `ancestorNode` is identified; however, while the `ancestorNode` is the underlying target of the proxy (the raw element), the parentNode accessors on the `referenceNode` return the proxies for these elements, so there is never a direct equality detected against the ancestor. Not sure how best to detangle that; or, if there are other workarounds please let me know.

Thanks for the great work with happy-dom!

TrevorBurnham:
Fixes #1876

## Problem

`Node.contains()` returns `false` for child elements of `HTMLFormElement` and `HTMLSelectElement`, even when the child is actually contained within the parent.

```javascript
form.appendChild(input);

form.contains(input); // Returns false, should be true
```

## Root Cause

`HTMLFormElement` and `HTMLSelectElement` return a Proxy from their constructors to support indexed access to form controls. When `appendChild()` is called, it correctly sets the child's `parentNode` to the proxy (via `self = this[PropertySymbol.proxy] || this`).

However, `contains()` was passing `this` directly to `isInclusiveAncestor()`. Inside the method, `this` refers to the underlying target object, not the proxy. When `isInclusiveAncestor()` traverses up the DOM tree comparing `ancestorNode === parent`, the comparison fails because one is the target and the other is the proxy.

## Solution

Modified `contains()` to use the proxy when available:

```typescript
public contains(otherNode: Node): boolean {
    if (otherNode === undefined) {
        return false;
    }
    const self = this[PropertySymbol.proxy] || this;
    return NodeUtility.isInclusiveAncestor(self, otherNode);
}
```

This follows the same pattern already used in `appendChild()`, `insertBefore()`, and `compareDocumentPosition()`.

## Testing

Added test cases for:
- `HTMLFormElement` containing direct child elements
- `HTMLSelectElement` containing option elements
- Nested elements within `HTMLFormElement`

### DIFF
--- a/packages/happy-dom/src/nodes/node/Node.ts
+++ b/packages/happy-dom/src/nodes/node/Node.ts
@@ -336,7 +336,10 @@ export default class Node extends EventTarget {
 		if (otherNode === undefined) {
 			return false;
 		}
-		return NodeUtility.isInclusiveAncestor(this, otherNode);
+		// HTMLFormElement and HTMLSelectElement return a Proxy from their constructor.
+		// We need to use the proxy for comparison to ensure correct behavior.
+		const self = this[PropertySymbol.proxy] || this;
+		return NodeUtility.isInclusiveAncestor(self, otherNode);
 	}
 
 	/**
--- a/packages/happy-dom/test/nodes/node/Node.test.ts
+++ b/packages/happy-dom/test/nodes/node/Node.test.ts
@@ -355,6 +355,39 @@ describe('Node', () => {
 
 			expect(div.contains(<Node>(<unknown>undefined))).toBe(false);
 		});
+
+		it('Returns "true" for HTMLFormElement containing child elements (issue #1876).', () => {
+			const form = document.createElement('form');
+			const input = document.createElement('input');
+
+			form.appendChild(input);
+
+			expect(form.contains(input)).toBe(true);
+			expect(form.contains(form)).toBe(true);
+		});
+
+		it('Returns "true" for HTMLSelectElement containing child elements (issue #1876).', () => {
+			const select = document.createElement('select');
+			const option = document.createElement('option');
+
+			select.appendChild(option);
+
+			expect(select.contains(option)).toBe(true);
+			expect(select.contains(select)).toBe(true);
+		});
+
+		it('Returns "true" for nested elements within HTMLFormElement (issue #1876).', () => {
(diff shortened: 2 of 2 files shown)


capricorn86 (approved):

PR #1993, opened 2026-01-17
fix: [#1895] EventTarget should not call arbitrary on* properties

### LINKED ISSUE #1895: EventTarget calls event functions that start with on
**Describe the bug**
If an instance of the `EventTarget` has an event that starts with `on` that event will be called while dispatching

**To Reproduce**
```ts
const t = new EventTarget();

t.onopen = () => console.log("onopen function");
t.addEventListener('open', () => console.log("onopen event"));

t.dispatchEvent(new Event('open'));
```

**Expected behavior**
It prints `onopen event`

**Actual behavior**
It prints `onopen function` and `onopen event`

**Additional context**
For exmaple, it breaks the [WebSocketOverride](https://github.com/mswjs/interceptors/blob/main/src/interceptors/WebSocket/WebSocketOverride.ts) in the `msw` that is extended by `EventTarget`

TrevorBurnham:
Fixes #1895

## Problem

Setting `on*` properties on a plain `EventTarget` instance incorrectly causes those functions to be called when dispatching events:

```javascript
const t = new EventTarget();
t.onopen = () => console.log("onopen function");
t.addEventListener('open', () => console.log("onopen event"));
t.dispatchEvent(new Event('open'));

// Expected: "onopen event"
// Actual: "onopen function" AND "onopen event"
```

This breaks libraries like [msw](https://github.com/mswjs/interceptors) that extend `EventTarget`.

## Root Cause

In browsers, `on*` event handler IDL attributes are only defined on specific interfaces (`HTMLElement`, `Document`, `Window`, etc. via the `GlobalEventHandlers` mixin), not on the base `EventTarget` class.

Happy-dom was checking for any `on*` property on the object and calling it as an event handler, regardless of whether the object was a DOM element with defined event handler attributes.

## Solution

Modified `EventTarget.#callDispatchEventListeners` to only call `on*` handlers if the object has `propertyEventListeners` (which is only defined on DOM elements, documents, shadow roots, etc.).

Plain `EventTarget` instances no longer have arbitrary `on*` properties called as event handlers.

## Testing

- Updated existing test to use a DOM element instead of plain EventTarget
- Added tests verifying that plain EventTarget does not call `on*` properties

### DIFF
--- a/integration-test/package.json
+++ b/integration-test/package.json
@@ -15,7 +15,7 @@
 	},
 	"devDependencies": {
 		"happy-dom": "^0.0.0",
-		"@types/node": "^20.0.0",
+		"@types/node": ">=20.0.0",
 		"express": "^4.20.0"
 	},
 	"engines": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
 			],
 			"devDependencies": {
 				"@types/he": "^1.1.2",
-				"@types/node": "^20.0.0",
+				"@types/node": ">=20.0.0",
 				"@typescript-eslint/eslint-plugin": "^6.19.0",
 				"@typescript-eslint/parser": "^6.19.0",
 				"chalk": "^4.1.0",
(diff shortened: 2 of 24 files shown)


TrevorBurnham:
CI is failing due to the flaky test described at #1985, which should be fixed by #1986.

capricorn86 (changes requested):
This may break some classes (such as `MediaStream` and `MediaQueryList`) that are not using `[PropertySymbol.propertyEventListeners]` yet.

In order to solve this we will probably have to find all classes that extends `EventTarget` and make sure that they also use `[PropertySymbol.propertyEventListeners]`.

TrevorBurnham:
@capricorn86 Good catch! I've opened up a new PR with a different approach: #2004

PR #1995, opened 2026-01-18
fix: [#1963] Use PropertySymbol.body to set body in cloneBodyStream

### LINKED ISSUE #1963: Crash when trying to mutate request body in `FetchBodyUtility.cloneBodyStream`
**Describe the bug**
I wish I could share a detailed explanation of what I'm seeing... but I lack the time and understanding to go deep into what's going on.

I'm setting up msw in my project that runs integration tests with vitests and sets happy-dom as the running environment.

When mocking a POST request, I get the following error

```
TypeError: Cannot set property body of #<Request> which has only a getter
    at Function.cloneBodyStream (file:///Users/sebastianbalay/projects/stilla-server/node_modules/happy-dom/lib/fetch/utilities/FetchBodyUtility.js:130:32)
```

**To Reproduce**
I don't know exactly what's different from my test suite that causes this crash, but all I'm doing is mocking a POST request with MSW.

Feel free to ask follow up question and I'll do my best to answer.

**Expected behavior**
Not crash

## Suggested solution

Looking around the happy-dom codebase, it seems that request objects use symbol to access read-write versions of its attributes, so the fix for this issue is maybe changing https://github.com/capricorn86/happy-dom/blob/v20.0.11/packages/happy-dom/src/fetch/utilities/FetchBodyUtility.ts#L149 with:

```
// use symbol accessors instead of requestOrResponse.body = stream1;
requestOrResponse[PropertySymbol.body] = stream1;
```

I applied that patch locally and things seem to be working.

TrevorBurnham:
Fixes #1963

## Problem

When using MSW or other libraries that clone request bodies, `FetchBodyUtility.cloneBodyStream()` throws:

```
TypeError: Cannot set property body of #<Request> which has only a getter
```

This occurs because the code was trying to set `requestOrResponse.body = stream` directly, but `Request.body` is a getter-only property.

## Solution

The fix checks whether the object is a `Request` or `Response` and sets the appropriate property:

- For `Request`: sets `[PropertySymbol.body]` (which the getter returns)
- For `Response`: sets the public `body` property directly

The detection uses `PropertySymbol.body in requestOrResponse` since `Request` defines body via a symbol property with a getter, while `Response` uses a public readonly property.

## Changes

- `packages/happy-dom/src/fetch/utilities/FetchBodyUtility.ts` - Updated `cloneBodyStream()` to handle both Request and Response objects correctly
- `packages/happy-dom/test/fetch/Request.test.ts` - Added test case that reproduces the issue

## Testing

I've added a new test case that verifies the fix.

### DIFF
--- a/packages/happy-dom/src/fetch/utilities/FetchBodyUtility.ts
+++ b/packages/happy-dom/src/fetch/utilities/FetchBodyUtility.ts
@@ -135,7 +135,15 @@ export default class FetchBodyUtility {
 			(<any>requestOrResponse.body)[PropertySymbol.nodeStream].pipe(stream1);
 			(<any>requestOrResponse.body)[PropertySymbol.nodeStream].pipe(stream2);
 			// Sets the body of the cloned request/response to the first pass through stream.
-			requestOrResponse.body = this.nodeToWebStream(stream1);
+			// Request uses [PropertySymbol.body] with a getter, Response uses public readonly body.
+			const newStream = this.nodeToWebStream(stream1);
+			if (PropertySymbol.body in requestOrResponse) {
+				// Request object - set the symbol property (getter will return it)
+				(<any>requestOrResponse)[PropertySymbol.body] = newStream;
+			} else {
+				// Response object - set the public property directly
+				(<ReadableStream | null>(<any>requestOrResponse).body) = newStream;
+			}
 			// Returns the clone.
 			return this.nodeToWebStream(stream2);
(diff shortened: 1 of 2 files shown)


TrevorBurnham:
The test failure in CI appears to be unrelated to this change, though I can't explain it.

TrevorBurnham:
I've opened a separate PR to address the flaky test: #2001

PR #1996, opened 2026-01-18
fix: [#1975] Allow paths in WebSocket URLs

### LINKED ISSUE #1975: shouldn't path be valid in WebSocket Url
Since 20.1.0, the lib support WebSocket, but I get an error when I try to connect  with path in url (ex: `wss://my-test.server.com/ws` or `wss://my-test.server.com/chat` should be valid)

Not sure why it throw an error if path is present https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/web-socket/WebSocket.ts#L64

TrevorBurnham:
Fixes #1975

## Problem

Since version 20.1.0, WebSocket URLs with paths (e.g., `wss://my-test.server.com/ws` or `wss://my-test.server.com/chat`) were incorrectly rejected with the error:

```
The URL contains a path name ('/ws'). Paths are not allowed in WebSocket URLs.
```

This is incorrect behavior - WebSocket URLs absolutely can have paths according to the WebSocket specification.

## Solution

- Removed the incorrect path validation that rejected URLs with `pathname.length > 1`
- Added fragment (hash) removal per the WebSocket spec - fragments should be stripped from WebSocket URLs

## Changes

- `packages/happy-dom/src/web-socket/WebSocket.ts` - Removed path validation, added fragment removal
- `packages/happy-dom/test/web-socket/WebSocket.test.ts` - Added tests for URLs with paths and fragment removal

## Testing

I've added 2 new test cases to cover this behavior.

### DIFF
--- a/packages/happy-dom/src/web-socket/WebSocket.ts
+++ b/packages/happy-dom/src/web-socket/WebSocket.ts
@@ -59,11 +59,11 @@ export default class WebSocket extends EventTarget {
 			);
 		}
 
-		if (parsedURL.pathname.length > 1) {
-			throw new window.DOMException(
-				`The URL contains a path name ('${parsedURL.pathname}'). Paths are not allowed in WebSocket URLs.`,
-				DOMExceptionNameEnum.syntaxError
-			);
+		// Note: WebSocket URLs can have paths (e.g., wss://example.com/chat/room1)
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):

PR #2001, opened 2026-01-18
fix: [#2000] Use waitUntilComplete() for reliable JavaScript URL test timing

### LINKED ISSUE #2000: Flaky test: "Opens a URL with Javascript" fails intermittently in CI
The test `BrowserWindow > open() > Opens a URL with Javascript` fails intermittently in CI ([example](https://github.com/capricorn86/happy-dom/actions/runs/21105560320/job/60696303456?pr=1995)).

The test uses `setTimeout(resolve, 1)` to wait for JavaScript URL execution to complete:

```js
const newWindow = window.open(`javascript:document.write('Test');`);
// ...
await new Promise((resolve) => setTimeout(resolve, 1));
expect(newWindow.document.body.innerHTML).toBe('Test');
```

However, JavaScript URL execution in `BrowserFrameNavigator.navigate()` uses `requestAnimationFrame()` followed by `setImmediate()`. Since these are scheduled via different mechanisms than setTimeout, there's no guaranteed ordering — especially under CI load where event loop timing can vary. A fixed timeout is inherently unreliable here

TrevorBurnham:
Fixes #2000

## Problem

The test `BrowserWindow > open() > Opens a URL with Javascript` was failing intermittently in CI:

```
AssertionError: expected '' to be 'Test' // Object.is equality
```

The test used a short `setTimeout` to wait for JavaScript URL execution to complete, but this is unreliable because `BrowserFrameNavigator.navigate()` executes `javascript:` URLs using `requestAnimationFrame()` + `setImmediate()`. These scheduling mechanisms have no guaranteed ordering relative to `setTimeout`, especially under CI load.

## Solution

Replace the arbitrary timeout with `browser.waitUntilComplete()`, which properly waits for all async tasks to finish. This pattern is already used elsewhere in the test file.

## Changes

- `packages/happy-dom/test/window/BrowserWindow.test.ts`: Updated two tests to use `browser.waitUntilComplete()` instead of `setTimeout`

### DIFF
--- a/packages/happy-dom/test/window/BrowserWindow.test.ts
+++ b/packages/happy-dom/test/window/BrowserWindow.test.ts
@@ -2240,7 +2240,7 @@ describe('BrowserWindow', () => {
 			const newWindow = <Window>window.open(`javascript:document.write('Test');`);
 			expect(newWindow).toBeInstanceOf(BrowserWindow);
 			expect(newWindow.location.href).toBe('about:blank');
-			await new Promise((resolve) => setTimeout(resolve, 1));
+			await browser.waitUntilComplete();
 			expect(newWindow.document.body.innerHTML).toBe('Test');
 		});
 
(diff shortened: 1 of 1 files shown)


capricorn86 (approved):

PR #2003, opened 2026-01-18
fix: [#1941] Node.getRootNode returns correct root for detached shadow trees

### LINKED ISSUE #1941: `Node.getRootNode` does not work correctly unless it attached to the `document`
**Describe the bug**
`Node.getRootNode({ composed: false })` retunrs the `Node` itself instead of the `ShadowRoot` if we do not add the `ShadowRoot` into the `document`.

**To Reproduce**
```ts
import { expect, test } from "vitest";

test("ShadowRoot", () => {
	const host = document.createElement("div");
	const child = document.createElement("span");

	const shadowRoot = host.attachShadow({ mode: "open" });

	shadowRoot.append(child);

	// It works if we uncomment this line
	// document.body.append(host);

	// It becomes a span if the `host` is not added to the `document`
	const root = child.getRootNode({ composed: false }); // Supposed to return the ShadowRoot

	expect(root).toBe(shadowRoot);
});
```

**Expected behavior**
Test should be passed

TrevorBurnham:
Fixes #1941

`Node.getRootNode({ composed: false })` was returning the node itself instead of the `ShadowRoot` when the shadow root was not attached to the document.

## Problem

When a node is inside a `ShadowRoot` that hasn't been appended to the document, calling `getRootNode()` incorrectly returned the node itself rather than the `ShadowRoot`.

```javascript
const host = document.createElement('div');

const shadowRoot = host.attachShadow({ mode: 'open' });
shadowRoot.append(child);

// Host is NOT attached to document
child.getRootNode({ composed: false }); // ❌ Returned: child (the span)
                                         // ✅ Expected: shadowRoot
```

## Solution

Modified `getRootNode()` to traverse up the parent chain when the node is not connected to the document, finding the actual root of the detached tree structure.

## Changes

- `packages/happy-dom/src/nodes/node/Node.ts` - Updated `getRootNode()` to traverse parent chain for detached nodes
- `packages/happy-dom/test/nodes/node/Node.test.ts` - Added test cases for detached ShadowRoot scenarios

## Testing

I've added two new test cases:
- Returns ShadowRoot when element is inside a detached ShadowRoot
- Returns ShadowRoot when composed is true and ShadowRoot is detached

### DIFF
--- a/integration-test/package.json
+++ b/integration-test/package.json
@@ -15,7 +15,7 @@
 	},
 	"devDependencies": {
 		"happy-dom": "^0.0.0",
-		"@types/node": "^20.0.0",
+		"@types/node": ">=20.0.0",
 		"express": "^4.20.0"
 	},
 	"engines": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
 			],
 			"devDependencies": {
 				"@types/he": "^1.1.2",
-				"@types/node": "^20.0.0",
+				"@types/node": ">=20.0.0",
 				"@typescript-eslint/eslint-plugin": "^6.19.0",
 				"@typescript-eslint/parser": "^6.19.0",
 				"chalk": "^4.1.0",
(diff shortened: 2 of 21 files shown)


TrevorBurnham:
CI test failure is due to #1985, which should be fixed by #1986.

PR #2004, opened 2026-01-19
fix: [#1895] EventTarget should not call arbitrary on* properties (take two)

### LINKED ISSUE #1895: EventTarget calls event functions that start with on

TrevorBurnham:
Fixes #1895

## Problem

Setting `on*` properties on a plain `EventTarget` instance incorrectly causes those functions to be called when dispatching events:

```javascript
const t = new EventTarget();
t.onopen = () => console.log("onopen function");
t.addEventListener('open', () => console.log("onopen event"));
t.dispatchEvent(new Event('open'));

// Expected: "onopen event"
// Actual: "onopen function" AND "onopen event"
```

This breaks libraries like [msw](https://github.com/mswjs/interceptors) that extend `EventTarget` and use `on*` properties for other purposes.

## Root Cause

In browsers, `on*` event handler IDL attributes are only defined on specific interfaces (`HTMLElement`, `Document`, `Window`, etc. via the `GlobalEventHandlers` mixin), not on the base `EventTarget` class.

Happy-dom was checking for any `on*` property on the object and calling it as an event handler, regardless of whether the object was a DOM element with defined event handler attributes.

## Solution

Modified `EventTarget.#callDispatchEventListeners` to distinguish between three cases:

1. **Classes with `propertyEventListeners`** (Element, Document, ShadowRoot, XMLHttpRequestEventTarget) - always check for `on*` handlers since the getters handle lazy evaluation of attribute-based event listeners like `onclick="..."`

2. **Subclasses that define `on*` as class properties** (MediaQueryList, AbortSignal, MediaStream, PermissionStatus, etc.) - call handlers if the property exists as an own property and the class is not a plain EventTarget

3. **Plain EventTarget instances** - do not call arbitrary `on*` properties

The key check is `this.constructor.name !== 'EventTarget'`, which identifies plain EventTarget instances (including the window-extended version) while allowing all subclasses that define `on*` handlers to work correctly.

## Testing

- Updated existing test to use a DOM element instead of plain EventTarget
- Added tests verifying that plain EventTarget does not call `on*` properties
- Added tests verifying that subclasses (MediaQueryList, MediaStream, AbortSignal) still have their `on*` handlers called correctly

## Changes

- `packages/happy-dom/src/event/EventTarget.ts` - Modified `#callDispatchEventListeners` to check class type before calling `on*` handlers
- `packages/happy-dom/src/xml-http-request/XMLHttpRequestEventTarget.ts` - Converted to use `propertyEventListeners` Map with getters/setters (consistent with Element pattern)
- `packages/happy-dom/test/event/EventTarget.test.ts` - Added comprehensive tests for the fix

### DIFF
--- a/packages/happy-dom/src/event/EventTarget.ts
+++ b/packages/happy-dom/src/event/EventTarget.ts
@@ -319,29 +319,48 @@ export default class EventTarget {
 		}
 
 		if (event.eventPhase !== EventPhaseEnum.capturing) {
+			// Only call on* handlers if they are defined as part of the class interface (via
+			// propertyEventListeners or as an own property on a non-EventTarget class), not arbitrary
+			// properties assigned at runtime on plain EventTarget instances. (issue #1895)
 			const onEventName = 'on' + event.type.toLowerCase();
-			const eventListener = (<any>this)[onEventName];
+			const propertyEventListeners = (<any>this)[PropertySymbol.propertyEventListeners];
 
-			if (typeof eventListener === 'function') {
-				// We can end up in a never ending loop if the listener for the error event on Window also throws an error.
-				if (
-					window &&
-					(this !== <EventTarget>window || event.type !== 'error') &&
-					!browserSettings?.disableErrorCapturing &&
-					browserSettings?.errorCapture === BrowserErrorCaptureEnum.tryAndCatch
-				) {
-					let result: any;
-					try {
-						result = eventListener(event);
-					} catch (error) {
-						window[PropertySymbol.dispatchError](<Error>error);
-					}
+			// Check if this on* handler is defined as part of the class interface:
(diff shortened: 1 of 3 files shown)


capricorn86 (approved):

PR #2005, opened 2026-01-19
chore: [#1978] Add tests for emoji numeric entity decoding

TrevorBurnham:
#1984 had the helpful side effect of fixing #1978. This PR adds test cases specifically for the bug described at #1978 to prevent future regressions.


capricorn86:
Nice! I missed that we should have merged the unit tests in #1979 before I closed it

PR #2006, opened 2026-01-19
fix: [#1942] Implement implicit form submission on Enter key in input elements

### LINKED ISSUE #1942: Form not submitted when pressing Enter key inside <input> element
**Describe the bug**
When pressing the `Enter` key inside a form input, the form is not submitted automatically. In a real browser, this action triggers the form’s `submit` event if the input is inside a `<form>` element. However, in `happy-dom`, no submit event is fired, even though the key event is dispatched correctly. This causes form tests using `vitest` + `@vue/test-utils` to fail.

**To Reproduce**
1. Create a simple Vue component:

```
<template>
  <form @submit.prevent="handleSubmit">
    <input v-model="name" placeholder="Your name" />
  </form>
</template>

<script setup>
import { ref } from 'vue'

const emit = defineEmits(['submit'])
const name = ref('')

function handleSubmit() {
  // emit the name as payload
  emit('submit', name.value)
}
</script>


```

2. Run it in Happy DOM and simulate key press
```
import { mount } from '@vue/test-utils'
import { describe, it, expect } from 'vitest'
import MyForm from './MyForm.vue'

describe('MyForm', () => {
  it('emits submit with input value when Enter pressed (happy-dom)', async () => {
    const wrapper = mount(MyForm)

    const input = wrapper.find('input')
    await input.setValue('John Doe')

    // simulate pressing Enter (happy-dom won't auto-submit)
    await input.trigger('keydown.enter')

   // await wrapper.find('form').trigger('submit'); if trigger form submit manually it works as expected

    // assertions
    expect(wrapper.emitted()).toHaveProperty('submit')
  })
})


```
4. Observe that no submit event is triggered.

**Expected behavior**
Pressing `Enter` in an input inside a form should trigger a `submit` event on the form, matching real browser behavior (e.g., Chrome, Firefox, Safari).

**Screenshots**

**Device:**
 - OS:  iOS
 - happy-dom: v20.0.8

TrevorBurnham:
Fixes #1942

## Summary

This PR implements the browser behavior where pressing the Enter key inside a form input element triggers implicit form submission. This is a standard browser behavior that was missing from happy-dom.

### `packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts`

- Added handling for `keydown` events with `key === 'Enter'` in the `dispatchEvent()` method
- Added `#handleEnterKeySubmission()` private method that implements the implicit submission logic per the HTML spec
- Added `#getDefaultSubmitButton()` helper to find the form's default submit button
- Added `#getImplicitSubmissionBlockingElements()` helper to determine if implicit submission should occur when no submit button exists

### `packages/happy-dom/test/nodes/html-input-element/HTMLInputElement.enterKeySubmit.test.ts`

Added comprehensive test coverage for the new functionality (11 tests).

## Implementation Details

The implementation follows the [HTML spec for implicit submission](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission):

- **Supported input types**: `text`, `search`, `url`, `tel`, `email`, `password`, `date`, `month`, `week`, `time`, `datetime-local`, `number`
- **Respects `preventDefault()`**: If a keydown listener calls `preventDefault()`, form submission is blocked
- **Disabled inputs**: Don't trigger submission
- **Default submit button**: Uses the form's first submit button as the submitter when available
- **Single field forms**: If no submit button exists, only submits when there's exactly one text-like input (per spec)
- **Validation**: Uses `requestSubmit()` which validates the form before submission

### DIFF
--- a/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
+++ b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
@@ -18,8 +18,10 @@ import Document from '../document/Document.js';
 import ShadowRoot from '../shadow-root/ShadowRoot.js';
 import { URL } from 'url';
 import MouseEvent from '../../event/events/MouseEvent.js';
+import KeyboardEvent from '../../event/events/KeyboardEvent.js';
 import NodeList from '../node/NodeList.js';
 import ElementEventAttributeUtility from '../element/ElementEventAttributeUtility.js';
+import HTMLButtonElement from '../html-button-element/HTMLButtonElement.js';
 
 /**
  * HTML Input Element.
@@ -1402,6 +1404,16 @@ export default class HTMLInputElement extends HTMLElement {
 	 * @override
 	 */
 	public override dispatchEvent(event: Event): boolean {
+		// Handle Enter key for implicit form submission
+		if (
+			event[PropertySymbol.type] === 'keydown' &&
+			event[PropertySymbol.eventPhase] === EventPhaseEnum.none &&
+			event instanceof KeyboardEvent &&
+			event.key === 'Enter'
+		) {
+			return this.#handleEnterKeySubmission(event);
+		}
+
 		if (
 			event[PropertySymbol.type] !== 'click' ||
 			event[PropertySymbol.eventPhase] !== EventPhaseEnum.none ||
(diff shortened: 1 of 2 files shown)


capricorn86 (changes requested):
I believe that this is a user triggered event. Meaning that it can't be simulated in the DOM by dispatching a keyboard event.

A tool like "testing-library" can be used to simulate user triggered events.

There is also a plan to add support for simulating user triggered events to the Happy DOM Browser sometime in the future.

Running this in the browser doesn't trigger the submit event:
```javascript
var form = document.createElement('form');
var input = document.createElement('input');
input.type = 'text';
input.name = 'testInput';
form.appendChild(input);

form.addEventListener('submit', (event) => {
    event.preventDefault();
    console.log('submitted');
});

var keydownEvent = new KeyboardEvent('keydown', {
    key: 'Enter',
    code: 'Enter',
    keyCode: 13,
    bubbles: true,
    cancelable: true
});

input.dispatchEvent(keydownEvent);
```

TrevorBurnham:
Ahh, you're right. 🤦‍♂️ No changes needed here!

Issue #1912, opened 2025-09-29
CSS attribute selector returns no elements


devmattrick:
**Describe the bug**
The CSS attribute selector (`[Name]`) doesn't seem to be working right now in happy-dom.

**To Reproduce**
See minimal example in https://github.com/devmattrick/happy-dom-issues

**Expected behavior**
Should return elements matching the attribute selector.

**Screenshots**
Browser
<img width="361" height="166" alt="Image" src="https://github.com/user-attachments/assets/a9acc372-9ee1-402c-bbf3-b9d0cffd56ef" />

jsdom
```
$ npm run test:jsdom

> happy-dom-scope@1.0.0 test:jsdom
> node -r global-jsdom/register test.js

Number of elements with Name attribute: 1
```

happy-dom
```
$ npm run test:happy-dom

> happy-dom-scope@1.0.0 test:happy-dom
> node -r ./happy-dom.js test.js

Number of elements with Name attribute: 0
```

**Device:**
 - OS: Windows 11
 - Browser: N/A
 - Version: 19.0.2

**Additional context**
N/A

codervipul775:
Hey @devmattrick . I want to work on this bug. Please assign it to me .

codervipul775:
@devmattrick , I ahve raised PR for fixing the bug CSS attribute selector returns no elements #1912.
Please review it and merge it.

hartej46:
Hey @devmattrick . I want to work . 
Please assign it to me .

capricorn86:
Thank you for reporting @devmattrick! 🙂 

I believe that the issue has been fixed now in v20.6.3.

devmattrick:
Awesome, thanks for your work on this @TrevorBurnham and of course @capricorn86 for the library!

capricorn86:
> Awesome, thanks for your work on this [@TrevorBurnham](https://github.com/TrevorBurnham) and of course [@capricorn86](https://github.com/capricorn86) for the library!

Thank you @devmattrick! I went a different route with the PR #2068 to keep a good performance and not introducing a new library. The fix included the unit tests from @TrevorBurnham PR.

PR #2010, opened 2026-01-20
fix: [#1910] Replace custom CSS selector parser with parsel-js

### LINKED ISSUE #1910: Error when parsing complex CSS has expression
**Describe the bug**
Crash when using a `screen.findByRole` command from testing-library/react.

```log
TypeError: Cannot read properties of null (reading '0')
at SelectorItem.matchPseudoItem (frontend/node_modules/happy-dom/src/query-selector/SelectorItem.ts:345:26)
at SelectorItem.matchPseudo (frontend/node_modules/happy-dom/src/query-selector/SelectorItem.ts:175:31)
at SelectorItem.match (frontend/node_modules/happy-dom/src/query-selector/SelectorItem.ts:107:24)
at SelectorItem.matchPseudoItem (frontend/node_modules/happy-dom/src/query-selector/SelectorItem.ts:252:23)
at SelectorItem.matchPseudo (frontend/node_modules/happy-dom/src/query-selector/SelectorItem.ts:175:31)
at SelectorItem.match (frontend/node_modules/happy-dom/src/query-selector/SelectorItem.ts:107:24)
at Function.matchSelector (frontend/node_modules/happy-dom/src/query-selector/QuerySelector.ts:383:31)
at Function.matches (frontend/node_modules/happy-dom/src/query-selector/QuerySelector.ts:353:24)
at CSSStyleDeclarationComputedStyle.parseCSSRules (frontend/node_modules/happy-dom/src/css/declaration/computed-style/CSSStyleDeclarationComputedStyle.ts:301:36)
at CSSStyleDeclarationComputedStyle.getComputedStyle 
```
**To Reproduce**
The crash happens with the following CSS string that is part of a reendered "table" element:
```css
.emXRrt>tbody>tr>td:nth-child(2),.emXRrt:not(:has(> tbody > tr:not([data-ignore-row])))>thead>tr>th:nth-child(2)
```

When calling `SelectorParser.getSelectorGroups`, the broken entry causing the error can be seen nested in the second array entry a bit nested:

```json
[
  {
    "name": "has",
    "arguments": null,
    "selectorItems": null,
    "nthFunction": null
  }
]
```

**Expected behavior**
No parsing error
**Screenshots**

**Additional context**
happy-dom 18.0.1, vitest 3.2.4

Thank you :)

TrevorBurnham:
Fixes #1910

## Problem

Complex CSS selectors with deeply nested pseudo-classes were being parsed incorrectly, causing wrong query results. For example:

```css
.emXRrt:not(:has(> tbody > tr:not([data-ignore-row])))>thead>tr>th:nth-child(2)
```

The issue was that the custom regex-based parser only handled one level of nested parentheses.

## Solution

Replaced the custom regex-based CSS selector parser with [parsel-js](https://parsel.verou.me/), a battle-tested CSS selector parser by Lea Verou.

Benefits:
- Handles arbitrarily nested parentheses correctly
- More spec-compliant CSS selector parsing  
- Well-maintained external library (~5KB minified)
- Reduces custom parsing code complexity
- Handles edge cases we may not have considered

## Implementation Details

The new parser:
1. Pre-processes selectors to handle escaped characters that parsel-js can't handle (e.g., `\:`, `\#`, `\&`, `\[`, `\]`)
2. Uses parsel-js to parse the selector into an AST
3. Converts the AST to the existing `SelectorItem` format
4. Restores escaped characters in the final output

Edge cases handled:
- Escaped characters in class names, IDs, and attribute names
- Empty attribute values (`[attr=""]`)
- Brackets in attribute values (`[attr="bracket[]bracket"]`)
- Space before combinators in `:has()` (e.g., `:has( +h2)`)

## Testing

Added two test cases for issue #1910:
1. The exact selector from the issue report with a table structure
2. A simpler test case demonstrating `:not(:has(> .child:not([attr])))` parsing

Updated tests for stricter CSS compliance:
- Invalid selectors (`.before:`, `.before&after`, `button:not([type]`) now throw errors instead of returning empty results
- Fixed test for escaped colon in attribute selector to use correct attribute name

## Changes

- `packages/happy-dom/package.json`: Added `parsel-js` dependency
- `packages/happy-dom/src/query-selector/SelectorParser.ts`: Rewrote to use parsel-js for parsing
- `packages/happy-dom/test/query-selector/QuerySelector.test.ts`: Added tests for issue #1910 and updated edge case tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9963,6 +9963,22 @@
 				"url": "https://github.com/sponsors/sindresorhus"
 			}
 		},
+		"node_modules/parsel-js": {
+			"version": "1.2.2",
+			"resolved": "https://registry.npmjs.org/parsel-js/-/parsel-js-1.2.2.tgz",
+			"integrity": "sha512-AVJMlwQ4bL2Y0VvYJGk+Fp7eX4SCH2uFoNApmn4yKWACUewZ+alwW3tyoe1r5Z3aLYQTuAuPZIyGghMfO/Tlxw==",
+			"funding": [
+				{
+					"type": "individual",
+					"url": "https://github.com/sponsors/LeaVerou"
+				},
+				{
+					"type": "opencollective",
+					"url": "https://opencollective.com/leaverou"
+				}
+			],
+			"license": "MIT"
+		},
 		"node_modules/parseurl": {
 			"version": "1.3.3",
 			"resolved": "https://registry.npmjs.org/parseurl/-/parseurl-1.3.3.tgz",
@@ -13855,6 +13871,7 @@
 				"@types/whatwg-mimetype": "^3.0.2",
 				"@types/ws": "^8.18.1",
 				"entities": "^4.5.0",
+				"parsel-js": "^1.2.2",
 				"whatwg-mimetype": "^3.0.0",
 				"ws": "^8.18.3"
 			},
--- a/packages/happy-dom/package.json
+++ b/packages/happy-dom/package.json
@@ -33,10 +33,11 @@
 		"test:debug": "vitest run --inspect-brk --no-file-parallelism"
 	},
 	"dependencies": {
-		"@types/whatwg-mimetype": "^3.0.2",
 		"@types/node": ">=20.0.0",
+		"@types/whatwg-mimetype": "^3.0.2",
 		"@types/ws": "^8.18.1",
 		"entities": "^4.5.0",
+		"parsel-js": "^1.2.2",
 		"whatwg-mimetype": "^3.0.0",
 		"ws": "^8.18.3"
 	},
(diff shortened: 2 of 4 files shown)


TrevorBurnham:
Looks like this also fixes #1912.

TrevorBurnham:
Also fixes #2034.

capricorn86:
It seems like this has a noticeable performance impact. Also "parcel-js" has quite few downloads and hasn't been updated for some time.

I would prefer to fix the logic in Happy DOM if possible. Each library added needs to be maintained and adds some risk.

TrevorBurnham:
> It seems like this has a noticeable performance impact…

I understand your concerns, but I suspect that any regex-based solution will always have edge-case bugs. I could imagine a hybrid solution where trivial selectors are parsed with regex and those with more complexity are handled by a more robust parser.

In the meantime, I've identified some low-hanging optimization fruit in parsel-js and sent over a PR: https://github.com/LeaVerou/parsel/pull/85 Let's see where that goes.

capricorn86:
@TrevorBurnham Thank you for your contribution! As mentioned I prefer to make a performant version in Happy DOM.

I created #2068 with a fix that uses 2-3 layers of RegExp  to separate the groups, but still in an efficient way, which I released now. I've added unit tests for all issues and it also covers the tests in this PR.

capricorn86:
I will close this PR now.

PR #2014, opened 2026-01-20
fix: [#2013] Flaky cache revalidation tests using Date.now() mocking

TrevorBurnham:
Fixes #2013

Fixes flaky cache revalidation tests in `Fetch.test.ts` and `SyncFetch.test.ts` that were intermittently failing in CI due to timing-dependent cache expiration logic.

## Problem

Seven tests that verify HTTP cache revalidation behavior (using `If-Modified-Since` and `If-None-Match` headers) were using `max-age=0.0001` (0.1 milliseconds) combined with `setTimeout(resolve, 100)` to wait for cache expiration. This approach is inherently unreliable because:

1. The extremely short `max-age` value (0.0001 seconds) leaves almost no margin for timing variations
2. `setTimeout` has no guaranteed ordering relative to other async operations
3. CI environments under load can introduce timing variations that cause the cache to not be marked as stale when expected

Example failure:
```
AssertionError: expected { 'content-type': 'text/html', …(3) } to deeply equal { 'content-type': 'text/html', …(3) }

- Expected
+ Received

{
-   "Cache-Control": "max-age=1",
-   "Last-Modified": "Mon, 11 Dec 2023 02:00:00 GMT",
+   "cache-control": "max-age=0.0001",
    "content-length": "9",
    "content-type": "text/html",
+   "last-modified": "Mon, 11 Dec 2023 01:00:00 GMT",
}
```

## Solution

Replace the unreliable `setTimeout` approach with deterministic `Date.now()` mocking, following the same pattern already used in `ResponseCache.test.ts`:

```typescript
let dateNowOffset = 0;
const originalDateNow = Date.now;
vi.spyOn(Date, 'now').mockImplementation(() => originalDateNow() + dateNowOffset);

// First request caches response with max-age=1
const response1 = await window.fetch(url);

// Advance time by 2 seconds to expire the cache
dateNowOffset = 2000;

// Second request triggers cache revalidation
const response2 = await window.fetch(url);
```

## Changes

- Mock `Date.now()` to control time deterministically instead of using real `setTimeout` delays
- Change `max-age` from `0.0001` to `1` second for more realistic cache durations
- Advance mocked time by 2 seconds to reliably expire the cache
- Update expected header assertions to match the new `max-age=1` value

## Affected Tests

**Fetch.test.ts:**
- Revalidates cache with a "If-Modified-Since" request for a GET response with "Cache-Control" set to a "max-age"
[...]

Issue #2013, opened 2026-01-20
Flaky test: Cache revalidation race condition


TrevorBurnham:
I notice that the test `Revalidates cache with a "If-Modified-Since" request for a GET response with "Cache-Control" set to a "max-age".` sometimes fails in CI ([example](https://github.com/capricorn86/happy-dom/actions/runs/21178246301/job/60913001167?pr=2011)). This looks similar to another flaky test involving cache revalidation (#1985) that was recently fixed (#1986). That fix should be extended to all tests that have similar timing issues.

PR #2015, opened 2026-01-20
fix: [#1918] XMLSerializer encodes non-breaking spaces as &#160; instead of &nbsp;

### LINKED ISSUE #1918: XMLSerializer incorrectly encodes non-breaking spaces in SVG as `&nbsp;`
**Describe the bug**
`XMLSerializer.serializeToString()` converts non-breaking spaces in text nodes to `&nbsp;`, which is [not a named entity in XML](https://developer.mozilla.org/en-US/docs/Web/XML/Guides/XML_introduction#character_references).

**To Reproduce**
```js
// the <tspan> here contains a non-breaking space
const document = new DOMParser().parseFromString('<body><svg viewBox="0 0 240 40" xmlns="http://www.w3.org/2000/svg"><text x="10" y="30" class="small">test<tspan> </tspan></text></svg></body>', 'text/html');
const svg = document.querySelector('svg');
const serializer = new XMLSerializer();
serializer.serializeToString(svg); // <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 240 40"><text x="10" y="30" class="small">test<tspan>&nbsp;</tspan></text></svg>
```

**Expected behavior**
Non-breaking spaces should be encoded as a numeric entity (`&#160;`) or left as-is

TrevorBurnham:
Fixes #1918 - `XMLSerializer.serializeToString()` now encodes non-breaking spaces as `&#160;` (numeric character reference) instead of `&nbsp;` (HTML named entity).

## Problem

`&nbsp;` is not a valid named entity in XML - only the five predefined entities (`&lt;`, `&gt;`, `&amp;`, `&quot;`, `&apos;`) are supported. When serializing SVG or other XML content containing non-breaking spaces, the output was invalid XML:

```js
const svg = document.querySelector('svg');
new XMLSerializer().serializeToString(svg);
// Before: <tspan>&nbsp;</tspan>  ❌ Invalid XML
// After:  <tspan>&#160;</tspan>  ✅ Valid XML
```

## Solution

Updated `XMLEncodeUtility.encodeTextContent()` to encode non-breaking spaces (`\xA0`) as `&#160;` instead of `&nbsp;`.

## Testing

Added test case verifying that `XMLSerializer` outputs `&#160;` and not `&nbsp;` for non-breaking spaces.

### DIFF
--- a/PR.md
+++ b/PR.md
@@ -1,0 +1,24 @@
+# fix: [#1918] XMLSerializer encodes non-breaking spaces as &#160; instead of &nbsp;
+
+## Summary
+
+Fixes #1918 - `XMLSerializer.serializeToString()` now encodes non-breaking spaces as `&#160;` (numeric character reference) instead of `&nbsp;` (HTML named entity).
+
+## Problem
+
+`&nbsp;` is not a valid named entity in XML - only the five predefined entities (`&lt;`, `&gt;`, `&amp;`, `&quot;`, `&apos;`) are supported. When serializing SVG or other XML content containing non-breaking spaces, the output was invalid XML:
+
+```js
+const svg = document.querySelector('svg');
+new XMLSerializer().serializeToString(svg);
+// Before: <tspan>&nbsp;</tspan>  ❌ Invalid XML
+// After:  <tspan>&#160;</tspan>  ✅ Valid XML
+```
+
+## Solution
+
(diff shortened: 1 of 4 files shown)


TrevorBurnham:
Test failure in CI appears to be related to #2013, and should be fixed by #2014.

PR #2024, opened 2026-01-21
fix: [#1938] Prevent util.inspect from throwing on href getters

### LINKED ISSUE #1938: Inspecting `Response.headers` with `%o` / `util.inspect(value, true)` throws “Cannot read private member #htmlHyperlinkElementUtility” in happy-dom v20.0.6
**Describe the bug**
A regression in `happy-dom` v20.0.6 causes logging a `Response.headers` instance to throw:

```
TypeError: Cannot read private member #htmlHyperlinkElementUtility from an object whose class did not declare it
    at get href (.../node_modules/happy-dom/lib/nodes/html-anchor-element/HTMLAnchorElement.js:54:21)
    at isURL (node:internal/util/inspect:208:23)
    at formatRaw (node:internal/util/inspect:1107:16)
    at formatValue (node:internal/util/inspect:922:10)
    at formatProperty (node:internal/util/inspect:2212:11)
    at formatRaw (node:internal/util/inspect:1141:9)
    at formatValue (node:internal/util/inspect:922:10)
    at formatProperty (node:internal/util/inspect:2212:11)
    at formatRaw (node:internal/util/inspect:1141:9)
    at formatValue (node:internal/util/inspect:922:10)
```

Note: `console.log(response.headers)` (no formatter) does **not** throw. The crash happens when the object is formatted with `%o` / `util.inspect(value, true)`.

**To Reproduce**

1. Install `happy-dom@20.0.6`.
2. Use Node.js v22.20.0.
3. Run the script below.

```ts
import util from 'node:util';

const window = new Window({ url: 'https://github.com' });

const response = await window.fetch('https://github.com');

// OK (no formatter):
console.log(response.headers);

// ❌ Throws the error above (Node's `%o` uses util.inspect(value, true)):
console.log('headers: %o', response.headers);

// ❌ Also throws:
console.log('headers: %s', util.inspect(response.headers, true));

// ✅ Does not throw:
console.log('headers: %s', util.inspect(response.headers, false));
```

**Expected behavior**
Inspecting / formatting a `Headers` object for logging should not throw.

**Screenshots**
N/A

**Device:**

 - OS: Ubuntu 25.04
 - Browser: N/A (Node.js environment)
 - Version:
   - happy-dom: v20.0.6 (commit aab20c3936ab5023c27f2b5a2ad7eaa42bd6d068)
   - Node.js: v22.20.0

**Additional context**
Reproduces after updating to `v20.0.6`; earlier versions did not throw with the same snippet.

TrevorBurnham:
Fixes #1938

## Problem

When logging a `Response.headers` instance with `util.inspect(value, true)` or `console.log('%o', value)`, Node.js throws:

```
TypeError: Cannot read private member #htmlHyperlinkElementUtility from an object whose class did not declare it
```

This happens because Node.js's `util.inspect` with `showHidden: true` traverses the entire object graph and checks if objects are URLs by accessing their `href` property. When it encounters `HTMLAnchorElement.prototype` or `HTMLAreaElement.prototype`, it calls the `href` getter which uses a private field (`#htmlHyperlinkElementUtility`). Since the getter is called on the prototype (not an instance), the private field access fails.

## Solution

Changed `HTMLAnchorElement` and `HTMLAreaElement` to use a module-scoped Symbol for storing the utility instance instead of a private field. Added an `instanceof` check in the getter to return an empty string when called on non-instances (like the prototype).

Also fixed `HTMLBaseElement` and `HTMLLinkElement` which had similar issues with their `href` getters accessing instance properties when called on the prototype.

## Changes

- `HTMLAnchorElement.ts`: Replace `#htmlHyperlinkElementUtility` private field with Symbol-based storage and add `instanceof` check
- `HTMLAreaElement.ts`: Same fix as HTMLAnchorElement
- `HTMLBaseElement.ts`: Add `instanceof` check to `href` getter
- `HTMLLinkElement.ts`: Add `instanceof` check to `href` getter

## Testing

```javascript
import util from 'node:util';

const window = new Window({ url: 'https://github.com' });

// These now work without throwing:
util.inspect(window.HTMLAnchorElement.prototype, { showHidden: true });
util.inspect(window.HTMLAreaElement.prototype, { showHidden: true });

// Anchor elements still work correctly:
const anchor = window.document.createElement('a');
anchor.href = 'https://example.com/path?query=1#hash';
console.log(anchor.href);      // https://example.com/path?query=1#hash
console.log(anchor.protocol);  // https:
console.log(anchor.hostname);  // example.com
```

### DIFF
--- a/packages/happy-dom/src/nodes/html-anchor-element/HTMLAnchorElement.ts
+++ b/packages/happy-dom/src/nodes/html-anchor-element/HTMLAnchorElement.ts
@@ -7,6 +7,26 @@ import HTMLHyperlinkElementUtility from '../html-hyperlink-element/HTMLHyperlink
 import IHTMLHyperlinkElement from '../html-hyperlink-element/IHTMLHyperlinkElement.js';
 import MouseEvent from '../../event/events/MouseEvent.js';
 
+// Used for caching the utility instance - module-scoped symbol to avoid prototype issues
+const HYPERLINK_UTILITY = Symbol('hyperlinkUtility');
+
+/**
+ * Returns the hyperlink utility for an element, creating it if necessary.
+ * Returns null if called on a non-instance (e.g., prototype).
+ *
+ * @param element The element to get the utility for.
+ * @returns The hyperlink utility or null.
+ */
+function getHyperlinkUtility(element: HTMLAnchorElement): HTMLHyperlinkElementUtility | null {
+	if (!(element instanceof HTMLAnchorElement)) {
+		return null;
+	}
+	if (!element[HYPERLINK_UTILITY]) {
+		element[HYPERLINK_UTILITY] = new HTMLHyperlinkElementUtility(element);
+	}
+	return element[HYPERLINK_UTILITY];
+}
+
 /**
  * HTML Anchor Element.
  *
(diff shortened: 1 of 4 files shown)


kalkv:
Any progress on it?

Issue #1605, opened 2024-11-12
[radio] keyboard navigation not fired when using `react-aria-components`


csantos-nydig:
**Describe the bug**

when using `RadioGroup` from react-aria-components and trying to navigate using keyboard (`ArrowDown`) the selection does not change.

**To Reproduce**
1. Go to https://stackblitz.com/edit/stackblitz-starters-erttrv?file=package.json,jsdom.spec.tsx,happy-dom.spec.tsx
2. Run `npm run test`
3. Notice it fails for HAPPY-DOM, but works for JSDOM

**Expected behavior**
I'd expect it to work in HAPPY-DOM

**Additional context**
I don't think this is an error in react-aria-components per se, because their components work fine in JSDOM; but my guess is that they are using some APIs that are not yet supported by HAPPY-DOM

I'm cross-linking both issues, so hopefully we can collaborate and see what needs to be done and where 🙏 https://github.com/adobe/react-spectrum/issues/7358

capricorn86:
Thank you for reporting @csantos-nydig! 🙂 

This has been fixed now in [v17.1.4](https://github.com/capricorn86/happy-dom/releases/tag/v17.1.4).

PR #2026, opened 2026-01-22
chore: [#1867] Add regression test for TreeWalker sibling traversal

### LINKED ISSUE #1867: TreeWalker stops traversing with sibling nodes of parent elements.
**Describe the bug**
TreeWalker does not continue to traverse when acceptNode returns FILTER_ACCEPT

**To Reproduce**
Just use this script:
```javascript

// Create a happy-dom window
const window = new Window();

// Create a simple nested DOM structure
document.body.innerHTML = `
<div>
  <form>
    <div>
      <label>Label 1</label>
    </div>
  </form>
  <div class="sibling">
    <label>Label 2</label>
  </div>
</div>
`;

// Create TreeWalker that accepts all elements
const walker = document.createTreeWalker(
  document.body,
  window.NodeFilter.SHOW_ELEMENT,
  { acceptNode: () => window.NodeFilter.FILTER_ACCEPT }
);

// Traverse all nodes
while (walker.nextNode()) {
  const element = walker.currentNode;
  const tag = element.tagName.toLowerCase();
  const className = element.className ? `.${element.className}` : '';
  console.log(`${tag}${className}`);
```
It outputs:
```
div
form
div
label
```

**Expected behavior**
Output:
```
div
form
div
label
div.sibling
label
```
I tested in Chrome and jsdom, their outputs are right.

**Device:**
 - OS: Debian
 - runtime: chrome/node/bun
 - Version: v138.0.7204.184/v24.3.0/v1.2.19

**happy-dom version**
v18.0.1

TrevorBurnham:
Closes #1867

#1867 reported that `TreeWalker.nextNode()` stops traversing when it should continue to sibling nodes of parent elements. 

**This issue is already fixed**. The fix was included in v17.1.4 (PR #1737) which addressed issue #1605.

This PR adds an explicit regression test for the exact scenario reported in #1867.

## Test Case Added

The new test verifies that TreeWalker correctly traverses this structure:
```html
<div>
  <form>
    <div>
      <label>Label 1</label>
    </div>
  </form>
  <div class="sibling">
    <label>Label 2</label>
  </div>
</div>
```

Expected traversal order: `div`, `form`, `div`, `label`, `div.sibling`, `label`

### DIFF
--- a/packages/happy-dom/test/tree-walker/TreeWalker.test.ts
+++ b/packages/happy-dom/test/tree-walker/TreeWalker.test.ts
@@ -141,6 +141,35 @@ describe('TreeWalker', () => {
 			]);
 		});
 
+		it('Traverses sibling nodes after deeply nested children (#1867).', () => {
+			const div = document.createElement('div');
+			div.innerHTML = `
+				<div>
+					<form>
+						<div>
+							<label>Label 1</label>
+						</div>
+					</form>
+					<div class="sibling">
+						<label>Label 2</label>
+					</div>
+				</div>
+			`;
+
+			const treeWalker = document.createTreeWalker(div, NodeFilter.SHOW_ELEMENT, {
+				acceptNode: () => NodeFilter.FILTER_ACCEPT
+			});
+
+			const tagNames: string[] = [];
+			while (treeWalker.nextNode()) {
+				const element = <Element>treeWalker.currentNode;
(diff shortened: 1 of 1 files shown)


capricorn86 (approved):
Thank you for your contribution @TrevorBurnham! ⭐

PR #2036, opened 2026-01-23
fix: [#2034] QuerySelector fails to match attribute values containing quotes

### LINKED ISSUE #2034: querySelector fails to match attribute values containing apostrophes
## Description

`querySelector` with an attribute selector fails to find elements when the attribute value contains an apostrophe (single quote), even when the selector uses double quotes as delimiters.

## Reproduction

https://github.com/JalilArfaoui/happy-dom-apostrophe-bug

```sh
git clone git@github.com:JalilArfaoui/happy-dom-apostrophe-bug.git
cd happy-dom-apostrophe-bug
npm install
npm test
```

Output:

Test 1: Apostrophe in double-quoted attribute value
  Selector: [id="type d'activité-label"]
  Result: ❌ NOT FOUND (bug!)
  Expected: ✅ FOUND

Test 2: Double quote in single-quoted attribute value
  Selector: [id='say "hello"']
  Result: ❌ NOT FOUND (bug!)
  Expected: ✅ FOUND

Test 3: Simple ID (control test)
  Selector: [id="simple"]
  Result: ✅ FOUND
  Expected: ✅ FOUND

The selector [id="type d'activité-label"] should match the element, since the value is delimited by double quotes and the apostrophe inside should not interfere.

This works correctly in all browsers (Chrome, Firefox, Safari).

Environment
- happy-dom version: 15.11.7
- Node.js version: 22.x
- Discovered via @testing-library/react with Vitest

JalilArfaoui:
Fixes #2034

## Root cause

The regex that captures attribute values used [^"']* which excludes both quote types regardless of which delimiter is actually used:

```
["']{1}([^"']*)["']{1}
```

## Fix

Use lookbehind/lookahead assertions to match the value based on the actual delimiter:

```
["']((?<=")[^"]*(?=")|(?<=')[^']*(?='))["']
```

This captures the value correctly without adding any new capture groups.

## Test plan                                                                                                                                                                                                                            
                                                                                                                                                                                                                                       
  - Added test for apostrophe in double-quoted value                                                                                                                                                                                   
  - Added test for double quote in single-quoted value                                                                                                                                                                                 
  - All existing tests pass (7149 tests)


TrevorBurnham:
This should be addressed by #2010, which replaces the custom CSS selector parser completely. I've added your test cases to that PR to confirm.

capricorn86:
Thank you for contributing @JalilArfaoui :star: 

I believe this was fixed by #2068, so I will close this now.

PR #2038, opened 2026-01-23
fix: [#2031] Replace regex-based HTML parser with parse5

### LINKED ISSUE #2031: Replace custom HTMLParser with parse5
There's a fast, spec-compliant HTML parser library called [parse5](https://github.com/inikulin/parse5). It's used by jsdom, and happy-dom would likely benefit from using it as well instead of its own regex-based [HTMLParser](https://github.com/capricorn86/happy-dom/blob/28f36e1aab20b1235aba5ff10758b8ce4cfe722b/packages/happy-dom/src/html-parser/HTMLParser.ts).

[As Jeff Atwood put it](https://blog.codinghorror.com/parsing-html-the-cthulhu-way/):

> If you attempt to parse HTML with regular expressions, you’re succumbing to the temptations of the dark god [Cthulhu’s](https://en.wikipedia.org/wiki/Cthulhu)… er… code.

TrevorBurnham:
Fixes #2031

This PR replaces the custom regex-based HTML parser with a [parse5](https://github.com/inikulin/parse5)-based implementation for spec-compliant HTML parsing.

## Changes

- **New `Parse5TreeAdapter`**: A tree adapter that allows parse5 to build Happy DOM nodes directly during parsing, avoiding post-processing conversion
- **Refactored `HTMLParser`**: Now uses parse5's `parse()` and `parseFragment()` functions with the custom tree adapter
- **Script execution fix**: Temporary documents created during parsing have `isConnected = false` to prevent premature script execution
- **Updated test expectations**: Adjusted whitespace expectations in tests to match parse5's output format

## Technical Details

The parse5 library implements the WHATWG HTML5 parsing algorithm, which handles:
- Implicit tag closing
- Optional tags
- Malformed HTML recovery
- Proper nesting rules
- Foreign content (SVG, MathML)

The `Parse5TreeAdapter` implements parse5's `TreeAdapter` interface to create Happy DOM nodes directly, supporting:
- Document, DocumentFragment, Element, Text, Comment, DocumentType nodes
- Namespace handling (HTML, SVG, MathML, XLink, XML, XMLNS)
- Template element content
- Script/link evaluation control via `disableEvaluation` flag

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9963,6 +9963,30 @@
 				"url": "https://github.com/sponsors/sindresorhus"
 			}
 		},
+		"node_modules/parse5": {
+			"version": "8.0.0",
+			"resolved": "https://registry.npmjs.org/parse5/-/parse5-8.0.0.tgz",
+			"integrity": "sha512-9m4m5GSgXjL4AjumKzq1Fgfp3Z8rsvjRNbnkVwfu2ImRqE5D0LnY2QfDen18FSY9C573YU5XxSapdHZTZ2WolA==",
+			"license": "MIT",
+			"dependencies": {
+				"entities": "^6.0.0"
+			},
+			"funding": {
+				"url": "https://github.com/inikulin/parse5?sponsor=1"
+			}
+		},
+		"node_modules/parse5/node_modules/entities": {
+			"version": "6.0.1",
+			"resolved": "https://registry.npmjs.org/entities/-/entities-6.0.1.tgz",
+			"integrity": "sha512-aN97NXWF6AWBTahfVOIrB/NShkzi5H7F9r1s9mD3cDj4Ko5f2qhhVoYMibXF7GlLveb/D2ioWay8lxI97Ven3g==",
+			"license": "BSD-2-Clause",
+			"engines": {
+				"node": ">=0.12"
+			},
+			"funding": {
+				"url": "https://github.com/fb55/entities?sponsor=1"
+			}
+		},
 		"node_modules/parseurl": {
 			"version": "1.3.3",
 			"resolved": "https://registry.npmjs.org/parseurl/-/parseurl-1.3.3.tgz",
@@ -13855,6 +13879,7 @@
 				"@types/whatwg-mimetype": "^3.0.2",
 				"@types/ws": "^8.18.1",
 				"entities": "^4.5.0",
+				"parse5": "^8.0.0",
 				"whatwg-mimetype": "^3.0.0",
 				"ws": "^8.18.3"
 			},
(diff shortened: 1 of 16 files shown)


aki05162525:
@TrevorBurnham 
Performance question: do we have any rough benchmarks comparing the old regex parser vs the new parse5-based implementation (e.g., parse(), parseFragment())? Just want to make sure we’re not introducing a noticeable regression for common workloads.

TrevorBurnham:
That's a good point. I threw together some benchmarks ([branch](https://github.com/TrevorBurnham/happy-dom/tree/htmlparser-benchmarks)) and the key findings were:

* Simple/small HTML shows ~100-120% slowdown (the overhead of parse5 is more noticeable on small inputs)
* Large documents show smaller relative slowdown (~10-37%) - the parse5 overhead is amortized over more content
* Template elements +131% slower
* SVG parsing is the biggest regression at +604% slower (likely due to namespace handling overhead in parse5)
* Only mixedContent (innerHTML) improved by 23.5%

I think that's likely a reasonable tradeoff to make for correctness, but it's not ideal.

TrevorBurnham:
I checked to see if there was any low-hanging fruit for optimizations in parse5 and it looks like it's already pretty well-optimized. I sent over one optimization, but it's not going to move the needle for our purposes: https://github.com/inikulin/parse5/pull/1679 I might dig deeper when I have the time.

I'd still say that the improvement in correctness justifies the performance hit.

aki05162525:
Thanks a lot for digging into this and running the benchmarks — really appreciate the detailed breakdown.

I agree that the correctness improvements from parse5 are valuable and probably justify the tradeoff overall.

That said, speaking as a user of happy-dom, one of the main reasons I (and likely others) choose it over jsdom is performance. As the [wiki](https://github.com/capricorn86/happy-dom/wiki) says:

> "Happy DOM focuses heavily on performance and can be used as an alternative to JSDOM."

So even if correctness comes first, I do hope happy-dom can continue to evolve in a direction where performance remains a core differentiator.

TrevorBurnham:
> I do hope happy-dom can continue to evolve in a direction where performance remains a core differentiator.

I feel the same way: sluggish tests are the biggest reason I started contributing to happy-dom. I'm continuing to look for optimization opportunities in both happy-dom and parse5.

capricorn86:
I've started on a branch to refactor the HTML and XML parsing to better follow the spec (but still in an optimized way):
https://github.com/capricorn86/happy-dom/pull/1666

My plan is to go through the open PRs and any outstanding bugs/feature requests and then continue on that.

TrevorBurnham:
@capricorn86 My sense is that parse5 is going to be tough to beat, but I respect the ambition! 🫡

Spixmaster:
I saw that https://github.com/capricorn86/happy-dom/pull/2010 was closed. Hence, I took a look at this pull request because of my general interest in disposing regex solutions. From my experience, they are a nightmare to maintain and will often have errors in some edge cases.

Personally, I programme in C++ and use [lexbor](https://github.com/lexbor/lexbor) for parsing HTML. It also has CSS parsing although it might be immature. Unfortunately, it does not have bindings for JavaScript. However, [gumbo-parser](https://www.npmjs.com/package/gumbo-parser) has. Maybe use that short-term and lexbor long-term as soon as bindings exist.

PR #2044, opened 2026-01-24
chore: [#deps] Upgrade whatwg-mimetype to v5.0.0

TrevorBurnham:
This PR upgrades [whatwg-mimetype](https://www.npmjs.com/package/whatwg-mimetype) from v3 to v5. This is a routine dependency upgrade that should have no impact on behavior. ~~The new package version is more modern and includes the TypeScript types, removing the need for the separate `@types/whatwg-mimetype` dep.~~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3412,9 +3412,9 @@
 			"license": "MIT"
 		},
 		"node_modules/@types/whatwg-mimetype": {
-			"version": "3.0.2",
-			"resolved": "https://registry.npmjs.org/@types/whatwg-mimetype/-/whatwg-mimetype-3.0.2.tgz",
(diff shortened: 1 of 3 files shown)


ocavue:
In your PR description:

> The new package version is more modern and includes the TypeScript types, removing the need for the separate `@types/whatwg-mimetype` dep.

But in your code:

> ```
> // @ts-expect-error whatwg-mimetype v5 has no TypeScript declarations
> import WhatwgMIMEType from 'whatwg-mimetype/lib/mime-type.js';
> ```

It seems that one of them is incorrect.

TrevorBurnham:
🤦‍♂️ Sorry about that! I see that there's a pending PR to update the `@types/whatwg-mimetype` package for v5: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74483 I'll update this PR to use those types once they're merged in.

TrevorBurnham:
I've updated this PR to use both `whatwg-mimetype@5.0.0` and the newly-minted `@types/whatwg-mimetype@5.0.0`.

Issue #2022, opened 2026-01-21
Flaky test: Timeout in DetachedWindowAPI


TrevorBurnham:
I'm seeing this test failure in CI ([example](https://github.com/capricorn86/happy-dom/actions/runs/21220922732/job/61055134221?pr=2021)):

```
 FAIL  test/window/DetachedWindowAPI.test.ts > DetachedWindowAPI > abort() > Cancels all ongoing asynchrounous tasks.
Error: Test timed out in 500ms.
```

I'm fairly certain that the change on the branch being tested is unrelated.

PR #2048, opened 2026-01-24
fix: [#2045] Replace JSON.parse(JSON.stringify()) with structuredClone()

### LINKED ISSUE #2045: Performance: Use structuredClone instead of JSON.parse(JSON.stringify(obj))
In multiple code paths, happy-dom uses `JSON.parse(JSON.stringify(obj))` to clone an object. We should be able to improve performance (and address any potential edge cases with JSON serialization) by using [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone#browser_compatibility), which has been available since Node.js 17. We currently have a stated dependency on Node.js 20+.

TrevorBurnham:
Resolves #2045

This PR uses `structuredClone()` for deep cloning instead of `JSON.parse(JSON.stringify())`.

## Files changed
- Headers.ts: Clone entries when constructing from another Headers instance
- CSSStyleDeclarationPropertyManager.ts: Clone properties in clone() method
- MediaStreamTrack.ts: Clone CAPABILITIES and SETTINGS constants

### DIFF
--- a/packages/happy-dom/src/css/declaration/property-manager/CSSStyleDeclarationPropertyManager.ts
+++ b/packages/happy-dom/src/css/declaration/property-manager/CSSStyleDeclarationPropertyManager.ts
@@ -539,7 +539,7 @@ export default class CSSStyleDeclarationPropertyManager {
 		const _class = <typeof CSSStyleDeclarationPropertyManager>this.constructor;
 		const clone: CSSStyleDeclarationPropertyManager = new _class();
 
-		clone.properties = JSON.parse(JSON.stringify(this.properties));
+		clone.properties = structuredClone(this.properties);
 		clone.definedPropertyNames = Object.assign({}, this.definedPropertyNames);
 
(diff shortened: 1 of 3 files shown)


TrevorBurnham:
I believe the CI failures are due to #2022, which is fixed by #2023.

capricorn86:
Is `structuredClone()` really faster?

Many benchmarks suggests that `JSON.stringify()` is much faster for certain objects as it doesn't have to round-trip to the JavaScript engine.

Here are some benchmarks:
https://www.measurethat.net/Benchmarks/Show/18541/0/jsonstringify-vs-structuredclone#latest_results_block

Would be interesting to benchmark with the actual code in Happy DOM.

TrevorBurnham:
Huh! I'm surprised, but my own checks back up those benchmarks. Hopefully future Node.js versions will improve `structuredClone` performance. Until then, closing this out.

PR #2043, opened 2026-01-24
feat: [#241] Add canvas adapter pattern for pluggable rendering support

RAprogramm:
## Summary

Adds `ICanvasAdapter` interface and static `setCanvasAdapter()` method to `HTMLCanvasElement`, allowing users to plug in real canvas implementations like node-canvas.

This is a **zero-dependency solution** - Happy-DOM remains lightweight, but users who need real canvas rendering can opt-in by installing node-canvas and setting the adapter.

## Changes

- Add `ICanvasAdapter` interface with `getContext()`, `toDataURL()`, `toBlob()` methods
- Add `HTMLCanvasElement.setCanvasAdapter()` and `getCanvasAdapter()` static methods  
- Delegate canvas operations to adapter when set (fallback to current behavior when not set)
- Add `NodeCanvasAdapter` implementation using node-canvas library
- Add 44 new tests (all 7195 tests pass)

Closes #241

PR #2069, opened 2026-02-20
feat: [#241] Add canvas adapter pattern for pluggable rendering support (vol.2)

### LINKED ISSUE #241: Add support for Canvas
e.g. `document.createElement('canvas').getContext('2d')`

RAprogramm:
the part 2 of [this](https://github.com/capricorn86/happy-dom/pull/2043) accidentally closed PR

Message from previous PR 
https://github.com/capricorn86/happy-dom/pull/2043#pullrequestreview-3775791974

---

Closes #241


RAprogramm:
@capricorn86, CI is failing because of tests in `@happy-dom/node-canvas-adapter` — it requires the native canvas library, which isn't available in the CI environment.

Since the canvas adapter is an optional dependency, what's the best approach:
 1. Skip these tests in CI?
 2. Add canvas dependencies installation to the workflow?
 3. Use mocks for tests in CI?

Could you advise on the right way to handle this?

Thanks!

capricorn86:
> @capricorn86, CI is failing because of tests in `@happy-dom/node-canvas-adapter` — it requires the native canvas library, which isn't available in the CI environment.
> 
> Since the canvas adapter is an optional dependency, what's the best approach:
> 
> 1. Skip these tests in CI?
> 2. Add canvas dependencies installation to the workflow?
> 3. Use mocks for tests in CI?
> 
> Could you advise on the right way to handle this?
> 
> Thanks!

@RAprogramm It looks really good! I will soon look into this more closely and see if something is missing, but I don't think so.

I think that the library can be added as dev dependency (in addition to be an optional dependency). It should solve the problem as it will only be available in the dev environment and not in the published package.

capricorn86:
Update: I've been working on adding types and make sure that functions such as `drawImage()` will work with various types.

PR #2070, opened 2026-02-20
fix: [#2053] Normalize invalid input type attribute to "text" per HTML spec

### LINKED ISSUE #2053: Invalid input type attribute not normalized to "text"
When setting an invalid `type` attribute on an `<input>` element, happy-dom keeps the invalid value instead of normalizing it to `"text"` as browsers do per the HTML spec.

## Expected behavior

Per [HTML spec](https://html.spec.whatwg.org/multipage/input.html#attr-input-type), when an invalid type is set, the input should behave as `type="text"` and the `type` IDL attribute should return `"text"`.

```js
const el = document.createElement('input');
el.setAttribute('type', '4'); // invalid type
console.log(el.type); // Should be "text"
```

## Actual behavior

```js
const el = document.createElement('input');
el.setAttribute('type', '4');
console.log(el.type); // Returns "4" (incorrect)
```

## Comparison with jsdom

jsdom correctly normalizes invalid types:

```js
// jsdom
el.setAttribute('type', '4');
el.type // → "text" ✓

// happy-dom  
el.setAttribute('type', '4');
el.type // → "4" ✗
```

## Real-world impact

This breaks libraries that check `el.type` to determine input behavior. For example, when a spread operator sets `type: 4` (a number coerced to string), the input should still behave as a text input.

## Environment

- happy-dom version: 16.7.2
- Node.js version: v22.x

atzzCokeK:
## Summary

- When an invalid value was set on the `type` attribute of an `<input>` element, `element.type` returned the invalid value instead of "text"
- This fix makes `element.type` return `"text"` for any invalid type value, matching browser behavior and the HTML spec
- `getAttribute('type')` still returns the raw value (this is correct per spec)

Fixes #2053

## Changes

- HTMLInputElement.ts: HTMLInputElement.ts: Define the 22 valid input type states from the [HTML spec](https://html.spec.whatwg.org/multipage/input.html#attr-input-type)
- HTMLInputElement.test.ts: Add tests for invalid type normalization

### DIFF
--- a/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
+++ b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
@@ -21,6 +21,33 @@ import MouseEvent from '../../event/events/MouseEvent.js';
 import type NodeList from '../node/NodeList.js';
 import ElementEventAttributeUtility from '../element/ElementEventAttributeUtility.js';
 
+// Valid input type states per HTML spec:
+// https://html.spec.whatwg.org/multipage/input.html#attr-input-type
+const INPUT_TYPE_STATES = new Set([
+	'hidden',
+	'text',
+	'search',
+	'tel',
+	'url',
+	'email',
+	'password',
+	'date',
+	'month',
+	'week',
+	'time',
+	'datetime-local',
+	'number',
+	'range',
+	'color',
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Thank you for your contribution @atzzCokeK! ⭐

PR #2072, opened 2026-02-21
fix: [#2054] Events created with native Event constructor now bubble correctly

### LINKED ISSUE #2054: Events created with `new Event()` don't bubble to parent elements
Events created using the `Event` constructor with `{ bubbles: true }` do not bubble to parent elements, even though `event.bubbles` is correctly set to `true`.

## Reproduction

```js
const window = new Window();

const parent = document.createElement('div');
parent.appendChild(child);
document.body.appendChild(parent);

let bubbled = false;
parent.addEventListener('click', () => {
  bubbled = true;
});

const event = new Event('click', { bubbles: true });
console.log('event.bubbles:', event.bubbles); // true ✓

child.dispatchEvent(event);
console.log('bubbled to parent:', bubbled); // false ✗ (should be true)
```

## Expected behavior

The event should bubble from `child` to `parent`, triggering the parent's event listener.

## Actual behavior

- `event.bubbles` is correctly `true`
- Event fires on `child`
- Event does NOT bubble to `parent`
- Parent's listener is never called

## Workaround

Using the legacy `createEvent`/`initEvent` API works correctly:

```js
const event = document.createEvent('Event');
event.initEvent('click', true, true); // (type, bubbles, cancelable)
child.dispatchEvent(event);
console.log('bubbled to parent:', bubbled); // true ✓
```

## Impact

This breaks any code using the modern Event constructor API for custom events or synthetic event dispatching. Many testing libraries and frameworks rely on this for simulating user interactions.

## Environment

- happy-dom version: 16.7.2
- Node.js version: v22.x

TrevorBurnham:
Closes #2054

### Problem

Events created with the native `Event` constructor (e.g. `new Event('click', { bubbles: true })`) don't bubble to parent elements when dispatched via `dispatchEvent()`. Events created with `document.createEvent()`/`initEvent()` or `new window.Event()` work fine.

The root cause: happy-dom's dispatch logic reads event properties via internal Symbol keys (`event[PropertySymbol.bubbles]`, `event[PropertySymbol.eventPhase]`, etc.). Native `Event` objects don't have these Symbol properties, so `event[PropertySymbol.bubbles]` returns `undefined` (falsy) and the bubbling phase is skipped entirely. Additionally, `event[PropertySymbol.eventPhase]` being `undefined` caused the capturing phase to incorrectly invoke bubbling-phase listeners.

### Changes

**`packages/happy-dom/src/event/EventTarget.ts`**
- At the top of `dispatchEvent()`, detect non-happy-dom events (where `event[PropertySymbol.type]` is `undefined`) and copy standard properties (`type`, `bubbles`, `cancelable`, `composed`, etc.) to the internal Symbol properties. Also override `composedPath()` to traverse happy-dom's DOM tree.
- In `#callDispatchEventListeners()`, read `eventPhase` from the Symbol property instead of the native getter, so the capturing/bubbling phase distinction works correctly for native events.

**`packages/happy-dom/test/event/EventBubbling.test.ts`** (new)
- Tests that native `Event` objects bubble to parent elements.
- Tests that native `Event` objects with `bubbles: false` do not bubble.
- Tests that native `Event` objects fire on the target element.

### DIFF
--- a/packages/happy-dom/src/event/EventTarget.ts
+++ b/packages/happy-dom/src/event/EventTarget.ts
@@ -1,5 +1,6 @@
 import * as PropertySymbol from '../PropertySymbol.js';
 import type Event from './Event.js';
+import HappyDOMEvent from './Event.js';
 import type IEventListenerOptions from './IEventListenerOptions.js';
 import EventPhaseEnum from './EventPhaseEnum.js';
 import WindowBrowserContext from '../window/WindowBrowserContext.js';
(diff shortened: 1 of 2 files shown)


capricorn86 (changes requested):
I believe that the correct behavior according to what the Browser does is to throw `new TypeError("Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'.")`

capricorn86:
I'm closing this as this has been fixed in #2092

PR #2073, opened 2026-02-22
fix: [#1848] Implement PageTransitionEvent with persisted property

### LINKED ISSUE #1848: `persisted` property of PageTransitionEvent not preserved
**Describe the bug**
The `persisted` property of [PageTransitionEvent](https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent/persisted), e.g. a `pageshow` event disappears between being dispatched and being handled.

**To Reproduce**
```ts
describe("pageTransitionAPI", () => {
    test("preserves `persisted` property in PageTransitionEvent", () => {
        let event: PageTransitionEvent | null = null;

        window.addEventListener("pageshow", (e) => {
            event = e;
        });

        window.dispatchEvent(
            new PageTransitionEvent("pageshow", { persisted: true }),
        );

        expect(event).not.toBeNull(); // Passes
        expect(event!.type).toBe("pageshow"); // Passes
        expect(event!.persisted).toBe(true); // Fails
    });
});
```

TrevorBurnham:
Closes #1848

### Problem

`PageTransitionEvent` was aliased to the base `Event` class, so the `persisted` property from the init object was silently dropped. Dispatching a `new PageTransitionEvent('pageshow', { persisted: true })` and reading `event.persisted` in the handler returned `undefined`.

### Changes

- Created `PageTransitionEvent` class extending `Event` with a readonly `persisted` property, defaulting to `false`.
- Created `IPageTransitionEventInit` interface.
- Updated `BrowserWindow` to use the new class instead of the `Event` alias.
- Exported both from the package index.
- Added tests covering constructor defaults, `persisted: true`, and round-tripping through `dispatchEvent`.

### DIFF
--- a/packages/happy-dom/src/event/events/IPageTransitionEventInit.ts
+++ b/packages/happy-dom/src/event/events/IPageTransitionEventInit.ts
@@ -1,0 +1,5 @@
+import type IEventInit from '../IEventInit.js';
+
+export default interface IPageTransitionEventInit extends IEventInit {
+	persisted?: boolean;
+}
--- a/packages/happy-dom/src/event/events/PageTransitionEvent.ts
+++ b/packages/happy-dom/src/event/events/PageTransitionEvent.ts
@@ -1,0 +1,23 @@
+import Event from '../Event.js';
+import type IPageTransitionEventInit from './IPageTransitionEventInit.js';
+
+/**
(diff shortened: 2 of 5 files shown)


IGx89:
Just ran into this limitation myself and would love to have this merged!

PR #2074, opened 2026-02-22
fix: [#1847] Navigation with location.href now adds to history

### LINKED ISSUE #1847: Navigation history not added to when navigating with `location.href`
**Describe the bug**
When navigating with `location.href`, it appears we are not adding to history

**To Reproduce**
```ts
describe("history.length", () => {
    test("changes as we navigate with history.pushState", () => {
        expect(history.length).toBe(1);

        history.pushState({}, "", "http://example.com");

        expect(history.length).toBe(2); // Passes
    });

    test("changes as we navigate with location.href", () => {
        expect(history.length).toBe(1);

        location.href = "http://example.com";

        expect(history.length).toBe(2); // Fails
    });
});
```

TrevorBurnham:
Closes #1847

### Problem

Setting `location.href` doesn't increment `history.length`. The `href` setter calls `browserFrame.goto()` → `BrowserFrameNavigator.navigate()`, which pushes to history for full navigations. However, when using a detached window (e.g. `new Window()`, the common test scenario), `validateFrameNavigation()` returns `false` and the code falls through to a path that only updates the URL via `setURL()` without pushing a history entry.

### Changes

**`packages/happy-dom/src/browser/utilities/BrowserFrameNavigator.ts`**
- In the `validateFrameNavigation` fallback path (used by detached windows), push a history entry before calling `setURL()`, matching the behavior of the normal navigation path.

**`packages/happy-dom/test/history/History.test.ts`**
- Added test verifying `history.length` increases after setting `location.href` on a detached window.

### DIFF
--- a/packages/happy-dom/src/browser/utilities/BrowserFrameNavigator.ts
+++ b/packages/happy-dom/src/browser/utilities/BrowserFrameNavigator.ts
@@ -128,6 +128,20 @@ export default class BrowserFrameNavigator {
 
 		if (!BrowserFrameValidator.validateFrameNavigation(frame)) {
 			if (!frame.page.context.browser.settings.navigation.disableFallbackToSetURL) {
+				if (!disableHistory) {
+					const history = frame[PropertySymbol.history];
+
+					history.push({
+						title: '',
+						href: targetURL.href,
+						state: null,
+						popState: false,
+						scrollRestoration: HistoryScrollRestorationEnum.auto,
+						method: method || (formData ? 'POST' : 'GET'),
+						formData: formData || null
+					});
+				}
+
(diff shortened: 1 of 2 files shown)


capricorn86 (changes requested):
History should not be changed when inside a window of a `DetachedBrowser` as there is no navigation happening. The user stays on the same page. For it to work with unit tests, it can fallback to setting the current URL without navigating.

TrevorBurnham:
> History should not be changed when inside a window of a `DetachedBrowser` as there is no navigation happening.

But it's already the case that `history.pushState()` (for example) works in `DetachedBrowser`. Why would we simulate realistic behavior for that method but not for the `href` setter?

PR #2080, opened 2026-02-24
fix: Properly decode CSS escape sequences in attribute selector values

silverwind:
## Summary

- `querySelectorAll` with attribute selectors failed to match elements when the selector contained CSS hex escape sequences (e.g. `\30 ` for `"0"`) as produced by `CSS.escape()`
- The parser's `ESCAPED_CHARACTER_REGEXP` (`/\\/g`) only stripped backslashes, turning `\30 abc` into `30 abc` instead of correctly decoding it to `0abc`
- Added a `cssUnescape()` method that properly handles both hex escapes (`\30 ` → `0`, `\0041` → `A`) and character escapes (`\:` → `:`) per the [CSS Syntax Level 3 spec](https://www.w3.org/TR/css-syntax-3/#escape-a-character)

## Repro

```js
const el = document.createElement('div');
el.innerHTML = '<div data-key="0abc"></div>';
el.querySelectorAll('[data-key="\\30 abc"]'); // returns [] (should return [div])

// Or via CSS.escape:
const escaped = CSS.escape('0abc'); // "\\30 abc"
el.querySelectorAll(`[data-key="${escaped}"]`); // returns [] (should return [div])
```

## Test plan

- [x] Added test for hex escape sequences in attribute values
- [x] Added test for character escape sequences in attribute values
- [x] Added test for `CSS.escape()` integration with class + attribute selector
- [x] All 7243 existing tests still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/happy-dom/src/query-selector/SelectorParser.ts
+++ b/packages/happy-dom/src/query-selector/SelectorParser.ts
@@ -651,7 +651,7 @@ export default class SelectorParser {
 	 *
 	 * Handles hex escapes (e.g. "\30 " → "0", "\0041" → "A") and character escapes (e.g. "\:" → ":").
 	 *
-	 * @see https://www.w3.org/TR/css-syntax-3/#escape-a-character
+	 * @see https://www.w3.org/TR/css-syntax-3/#consume-escaped-code-point
 	 * @param value Escaped CSS value.
 	 * @returns Unescaped value.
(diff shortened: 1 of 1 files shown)


capricorn86 (approved):
Thank you for your contribution @silverwind! ⭐

PR #2083, opened 2026-02-26
fix: [#1422] Make inert attribute block click and focus interactions

### LINKED ISSUE #1422: `inert` not blocking user interaction
when `inert` is present, it should block any interaction (focus, click, hover) on the given component.

**To Reproduce**
1. Go to https://stackblitz.com/edit/vitest-dev-vitest-dkwbqj?file=test%2Fbasic.test.tsx,package.json
2. See error in the test console

**Expected behavior**
Should not be able to click the button with `inert` property present

**Current behavior**
The button is still clickable

**Device:**

**Additional context**

`inert` should block any interaction on all descendent children, for example:

```
<div inert>
  <button />
  <input />
</div>
```

Should block interaction in both elements: `button` and `input`.

coffeeandwork:
## Summary
- When an element or any of its ancestors has the `inert` attribute, `click()` now returns without dispatching the event and `focus()` is prevented
- Adds `HTMLElementUtility.isInert()` helper that walks the ancestor tree checking for the `inert` attribute
- Per the [HTML spec](https://html.spec.whatwg.org/multipage/interaction.html#inert-subtrees), inert elements should not receive focus or respond to user interaction

## Test plan
- Added tests verifying `click()` is blocked on inert elements
- Added tests verifying `click()` is blocked on descendants of inert elements
- Added tests verifying `focus()` is blocked on inert elements
- Added tests verifying `focus()` is blocked on descendants of inert elements
- Added tests verifying behavior is restored when `inert` is removed
- Added tests for deeply nested elements under inert ancestors
- Verified all existing tests still pass

Closes #1422

### DIFF
--- a/packages/happy-dom/src/nodes/html-element/HTMLElement.ts
+++ b/packages/happy-dom/src/nodes/html-element/HTMLElement.ts
@@ -947,9 +947,6 @@ export default class HTMLElement extends Element {
 	 * Triggers a click event.
 	 */
 	public click(): void {
-		if (HTMLElementUtility.isInert(this)) {
-			return;
-		}
 		this.dispatchEvent(
 			new PointerEvent('click', {
 				bubbles: true,
--- a/packages/happy-dom/src/nodes/html-element/HTMLElementUtility.ts
+++ b/packages/happy-dom/src/nodes/html-element/HTMLElementUtility.ts
@@ -8,25 +8,6 @@ import type SVGElement from '../svg-element/SVGElement.js';
  */
 export default class HTMLElementUtility {
 	/**
-	 * Returns whether an element or any of its ancestors has the inert attribute.
-	 *
-	 * @param element Element to check.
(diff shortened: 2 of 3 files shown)


capricorn86 (changes requested):
Thank you for your contribution @coffeeandwork! ⭐ 

It seems like this isn't the behavior when dispatching an event pragmatically.

Try running this in the browser console:
```typescript
var container = document.createElement('div');
var div = document.createElement('div');
var button = document.createElement('button');

div.setAttribute('inert', '');
button.textContent = 'Click me!';
div.appendChild(button);
container.appendChild(div);
document.body.appendChild(container);

container.addEventListener('click', () => console.log('Button clicked!'));

button.click();
```

capricorn86:
Looking a bit further into this. It seems like "inert" should just block "focusing" an element, but not clicking it.

Issue #1845, opened 2025-06-11
error TS2420: Class 'VirtualConsole' incorrectly implements interface 'Console'.


Spixmaster:
v18.0.1

In tsconfig.json with `"skipLibCheck": false` I get following error.

```sh
$ typedoc --cleanOutputDir --darkHighlightTheme dracula --name Argo --out ./documentation-web/ --readme ./README.md --tsconfig ./tsconfig.json ./src-web/typescript/**/*.ts
node_modules/happy-dom/lib/console/VirtualConsole.d.ts:8:22 - error TS2420: Class 'VirtualConsole' incorrectly implements interface 'Console'.
  Type 'VirtualConsole' is missing the following properties from type 'Console': write, [Symbol.asyncIterator]

8 export default class VirtualConsole implements Console {
                       ~~~~~~~~~~~~~~

[error] Found 1 errors and 0 warnings
```

capricorn86:
@Spixmaster which version of Node.js do you use?

Spixmaster:
Bun v1.3.10 currently. The error still persists.

I tested with happy-dom v20.8.3. Let me check again.

v20.8.4 still has the same issue.

capricorn86:
Something must have been introduced in a later version of Bun (or Node if the types are for Node) that makes VirtualConsole incompatible with the interface.

YevheniiKotyrlo:
I dug into this more — the root cause is that [Bun's type definitions extend the global `Console` interface](https://github.com/oven-sh/bun/blob/main/packages/bun-types/globals.d.ts) with `write()` and `[Symbol.asyncIterator]()`. These aren't in `@types/node` or `lib.dom.d.ts`, so when `VirtualConsole` declares `implements Console`, it needs to satisfy Bun's extended interface too.

I've submitted a fix in #2102 that adds both methods. Hope this helps!

capricorn86:
@Spixmaster, thanks to @YevheniiKotyrlo there is now a fix in

PR #2095, opened 2026-03-05
fix: [#1845] Replace ConsoleConstructor import with indexed access type

YevheniiKotyrlo:
**Update:** This PR fixed the `ConsoleConstructor` import (TS2305) but did not fix the TS2420 reported in #1845 — I misattributed the issue. I should have filed a separate issue for the ConsoleConstructor problem rather than linking it to #1845. The actual fix for #1845 is in #2102.

---

## Problem

`VirtualConsole.ts` imports `ConsoleConstructor` from the Node.js `'console'` module:

```typescript
import type { ConsoleConstructor } from 'console';
```

This import is emitted into the published `lib/console/VirtualConsole.d.ts`, causing `TS2305: Module '"console"' has no exported member 'ConsoleConstructor'` for consumers whose TypeScript configuration doesn't resolve Node.js built-in module types — for example, projects with `skipLibCheck: false` and a `moduleResolution` that doesn't map the `'console'` specifier to `@types/node`'s console declarations.

### Reproduction

Any consumer project with `skipLibCheck: false` that imports something from happy-dom touching VirtualConsole's type will get:

```
node_modules/happy-dom/lib/console/VirtualConsole.d.ts(2,15): error TS2305: Module '"console"' has no exported member 'ConsoleConstructor'.
```

This is related to #1845, which reports a similar type compatibility issue with VirtualConsole.

## Fix

Replace the `ConsoleConstructor` import with an indexed access type `Console['Console']`. This resolves to the same type (the `Console` property on the global `Console` interface points to `ConsoleConstructor`) but avoids the `import from 'console'` module specifier entirely.

### Before

```typescript
import type { ConsoleConstructor } from 'console';
public declare Console: ConsoleConstructor;
```

### After

```typescript
// No import needed
public declare Console: Console['Console'];
```

## Verification

- `npm run compile`: passes
- `npm run lint` (root): passes (zero warnings)
- `npm test` (packages/happy-dom): all 48 console tests pass. 4 pre-existing failures in unrelated test files (SyncFetch, ModuleURLUtility, BrowserWindow, DetachedWindowAPI) — identical on clean `master`.
- Consumer project with `skipLibCheck: false`: zero ConsoleConstructor errors after fix (confirmed `TS2305` without fix)

### DIFF
--- a/packages/happy-dom/src/console/VirtualConsole.ts
+++ b/packages/happy-dom/src/console/VirtualConsole.ts
@@ -2,7 +2,6 @@ import type IVirtualConsolePrinter from './IVirtualConsolePrinter.js';
 import VirtualConsoleLogLevelEnum from './enums/VirtualConsoleLogLevelEnum.js';
 import VirtualConsoleLogTypeEnum from './enums/VirtualConsoleLogTypeEnum.js';
 import type IVirtualConsoleLogGroup from './IVirtualConsoleLogGroup.js';
-import type { ConsoleConstructor } from 'console';
 
 /**
  * Virtual Console.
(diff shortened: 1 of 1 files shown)


Spixmaster:
This reads like a typical LLM pull request. Convinced without doubt to solve an issue but in the end the error is the same as before.

capricorn86:
> This reads like a typical LLM pull request. Convinced without doubt to solve an issue but in the end the error is the same as before.

@Spixmaster I know it was an LLM and I know that it didn't solve #1845, but I guess getting rid of the import is a small improvement (even though it probably didn't have any real effect), so I approved it.

YevheniiKotyrlo:
Hey, wanted to follow up on this — my PR didn't actually fix #1845. I was looking at the wrong error (TS2305 from the ConsoleConstructor import) while the real issue was TS2420 (missing interface members).

After digging deeper, I found that [Bun's type definitions extend the global `Console` interface](https://github.com/oven-sh/bun/blob/main/packages/bun-types/globals.d.ts) with `write()` and `[Symbol.asyncIterator]()`, which aren't in `@types/node` or `lib.dom.d.ts`. That's what causes the TS2420 for Bun users with `skipLibCheck: false`.

I've submitted a proper fix in #2102 that adds both methods. Sorry for the confusion on this one.

PR #2102, opened 2026-03-14
fix: [#1845] Replace implements Console with custom IVirtualConsole interface

YevheniiKotyrlo:
Bun's type definitions extend the global `Console` interface with `write()` and `[Symbol.asyncIterator]()`. These aren't in `@types/node` or `lib.dom.d.ts`, so Bun users with `skipLibCheck: false` get TS2420 because `VirtualConsole` doesn't have them.

My previous PR (#2095) fixed the wrong error — it addressed TS2305 (the `ConsoleConstructor` import), not the TS2420 that @Spixmaster reported. Sorry about that.

This adds both methods:
- `write()` — routes data through the printer, returns 0
- `[Symbol.asyncIterator]()` — empty async generator (no stdin in a virtual console)

**Reproduction:**

```bash
mkdir repro && cd repro && npm init -y
npm install happy-dom@20.8.3 bun-types@1.3.10 typescript@5.8.3
```

tsconfig.json with `"skipLibCheck": false` and `"types": ["bun-types"]`, then any file that imports VirtualConsole:

```
# before fix:
VirtualConsole.d.ts(8,22): error TS2420 — missing write, [Symbol.asyncIterator]

# after fix:
0 errors
```

Tests added for both methods.

### DIFF
--- a/packages/happy-dom/src/window/BrowserWindow.ts
+++ b/packages/happy-dom/src/window/BrowserWindow.ts
@@ -395,15 +395,15 @@ export default class BrowserWindow extends EventTarget implements INodeJSGlobal 
 	public readonly DocumentType = DocumentType;
 
 	// Nodes that can be created using "new" keyword (populated by WindowContextClassExtender)
-	declare public readonly Document: typeof Document;
-	declare public readonly HTMLDocument: typeof HTMLDocument;
-	declare public readonly XMLDocument: typeof XMLDocument;
-	declare public readonly DocumentFragment: typeof DocumentFragment;
-	declare public readonly Text: typeof Text;
-	declare public readonly Comment: typeof Comment;
-	declare public readonly Image: typeof Image;
-	declare public readonly Audio: typeof Audio;
-	declare public readonly WebSocket: typeof WebSocket;
+	public declare readonly Document: typeof Document;
+	public declare readonly HTMLDocument: typeof HTMLDocument;
+	public declare readonly XMLDocument: typeof XMLDocument;
+	public declare readonly DocumentFragment: typeof DocumentFragment;
+	public declare readonly Text: typeof Text;
+	public declare readonly Comment: typeof Comment;
+	public declare readonly Image: typeof Image;
+	public declare readonly Audio: typeof Audio;
+	public declare readonly WebSocket: typeof WebSocket;
 
 	// HTML Element classes
 	public readonly HTMLAnchorElement = HTMLAnchorElement;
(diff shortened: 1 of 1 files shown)


YevheniiKotyrlo:
You're right that adding `write()` and `[Symbol.asyncIterator]()` is wrong. I went back and traced the actual root cause this time.

## Why `implements Console` breaks

The `implements Console` clause in the compiled `.d.ts` is what triggers TS2420. When consumers use `skipLibCheck: false`, TypeScript re-checks this clause against *their* `Console` definition — and that definition varies across runtimes.

I compared the [WHATWG Console spec](https://console.spec.whatwg.org/#namespacedef-console) against the three main TypeScript type sources:

| Member | [WHATWG Spec](https://console.spec.whatwg.org/#namespacedef-console) | [lib.dom.d.ts](https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts) | [@types/node](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/console.d.ts) | [bun-types](https://bun.sh/reference/globals/Console) |
|---|---|---|---|---|
| 19 core methods | Yes | Yes | Yes | Yes |
| `timeStamp()` | No | Yes | Yes | Yes |
| `profile()` / `profileEnd()` | No | No | Yes | Yes |
| `Console` constructor property | No | No | Yes | Yes |
| `write()` | No | No | No | Yes |
| `[Symbol.asyncIterator]()` | No | No | No | Yes |

The WHATWG spec defines `console` as a **namespace**, not an interface or class — there's no `Console` type in the spec at all. Each runtime's TypeScript definitions invent their own `Console` interface with different members. So `implements Console` ties happy-dom to whatever `Console` the consumer's environment provides.

## Fix: custom `IVirtualConsole` interface

I created an `IVirtualConsole` interface owned by happy-dom, containing all standard console methods with VirtualConsole's actual signatures. Then replaced `implements Console` with `implements IVirtualConsole` and changed all `Console` type annotations throughout the browser/window layer to `IVirtualConsole`.

This works because `Console` (any runtime) is a superset of `IVirtualConsole` — `globalThis.console` remains assignable to `IVirtualConsole`, so `new Window({ console: globalThis.console })` still works.

Changes:
- New file: `IVirtualConsole.ts` (22 methods, excludes deprecated `exception()`)
- `VirtualConsole`: `implements IVirtualConsole` instead of `implements Console`, removed `Console['Console']` property
- 8 files: `Console` → `IVirtualConsole` type annotations (BrowserWindow, Window, BrowserPage, Browser, DetachedBrowser, DetachedBrowserPage, IBrowserPage, IBrowser)
- `IVirtualConsole` exported from index.ts

I noticed that [d7c7d45b](https://github.com/capricorn86/happy-dom/commit/d7c7d45b9190663ca545aa6184a28eb812ab46ca) (294 files, -6168 lines) moved *away* from custom interfaces toward `implements` on global interfaces. This partially reverses that for Console, but the reason is different — it's not about circular dependencies, it's about runtime type divergence. The same `implements <global interface>` pattern is used for `Request`, `Response`, and `Attr`, and they'd be similarly vulnerable if any runtime extends those interfaces. For reference, [jsdom](https://github.com/jsdom/jsdom) avoids this entirely — their [VirtualConsole](https://github.com/jsdom/jsdom/blob/main/lib/jsdom/virtual-console.js) extends `EventEmitter` and doesn't implement `Console` at all.

## Validation

I set up a [reproduction repo](https://github.com/YevheniiKotyrlo/repro-happy-dom-virtual-console) that type-checks happy-dom's `.d.ts` across three consumer environments with `skipLibCheck: false`:

- [`main` branch](https://github.com/YevheniiKotyrlo/repro-happy-dom-virtual-console/actions?query=branch%3Amain) — unpatched happy-dom@20.8.3, shows the failures
- [`fix` branch](https://github.com/YevheniiKotyrlo/repro-happy-dom-virtual-console/actions?query=branch%3Afix) — same version with the [patch](https://github.com/YevheniiKotyrlo/repro-happy-dom-virtual-console/blob/fix/patches/happy-dom%4020.8.3.patch) that applies #2095 + the custom interface, all passing

| Environment | `main` (unpatched) | `fix` (patched) |
|---|---|---|
| Node.js + @types/node | pass | pass |
| Bun + bun-types | **fail** (TS2305 + TS2420) | pass |
| Browser (lib.dom) | **fail** (TS2305) | pass |

The TS2305 errors on `main` are from the `ConsoleConstructor` import — already fixed by #2095 but not yet released in 20.8.3. The patch includes both fixes.

I'll push the reworked code to this PR now.

PR #2107, opened 2026-03-17
fix: [#2106] `Request.formData()` should honor `Content-Type` header

### LINKED ISSUE #2106: `Request.formData()` fails on string body with explicit `Content-Type` header
**Describe the bug**
`Request.formData()` throws when the body is a string and the `Content-Type` header is explicitly set to `application/x-www-form-urlencoded`. It checks the body-inferred content type (`text/plain;charset=UTF-8`) instead of the `Content-Type` header.

`Response.formData()` does not have this bug — it already reads from `this.headers.get('Content-Type')`.

**To Reproduce**
```js
const window = new Window();

const req = new window.Request('https://example.com', {
  method: 'POST',
  headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
  body: 'key1=value1&key2=value2',
});

await req.formData();
// DOMException: Failed to construct FormData object: The "content-type" header
// is neither "application/x-www-form-urlencoded" nor "multipart/form-data".
```

**Expected behavior**
`formData()` should parse the body as URL-encoded form data, since the `Content-Type` header is `application/x-www-form-urlencoded`. Node's native `Request` handles this correctly.

**Device:**
- OS: Linux
- Version: happy-dom 20.8.4

brianhelba:
`Request.formData()` was checking the internal `PropertySymbol.contentType` (inferred from the body init) instead of the actual `Content-Type` header. This caused `formData()` to fail when a string body was paired with an explicit `Content-Type` header, since the body-inferred type was always `text/plain;charset=UTF-8`.

This aligns `Request.formData()` with `Response.formData()`, which already reads from headers.

Fixes #2106.

### DIFF
--- a/packages/happy-dom/src/fetch/Request.ts
+++ b/packages/happy-dom/src/fetch/Request.ts
@@ -432,7 +432,7 @@ export default class Request implements Request {
 		const window = this[PropertySymbol.window];
 		const asyncTaskManager = new WindowBrowserContext(window).getAsyncTaskManager()!;
 
-		const contentType = this[PropertySymbol.contentType];
+		const contentType = this.headers.get('Content-Type') ?? this[PropertySymbol.contentType];
 
 		if (this.body && contentType && /multipart/i.test(contentType)) {
 			if (this[PropertySymbol.bodyUsed]) {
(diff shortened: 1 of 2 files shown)


brianhelba:
@capricorn86 Please take a look.

Issue #2099, opened 2026-03-11
Timeout and PROMISE leaks reported by vitest 4.1-beta --detect-async-leaks are unfixable from userland


jtnix:
**Describe the bug**
Tests using `happy-dom` as the Vitest DOM environment report spurious `Timeout leaking` and `PROMISE leaking` warnings when running with `--detect-async-leaks`. These are false positives — the leaks originate entirely inside happy-dom's own internal implementations and cannot be resolved from userland test code.

**To Reproduce**
1. Set up a Vitest 4.1-beta project with `environment: 'happy-dom'`
2. Write a test that renders a component using `@tanstack/react-virtual` (or any component with a debounced `useEffect`), or any test that consumes a `fetch` response via `Response.json()`
3. Run `vitest run --detect-async-leaks`
4. Observe `Timeout leaking` and/or `PROMISE leaking` reported against the test file

**Expected behavior**
happy-dom should cancel all internally-owned timers and close any open `ReadableStream`s during window/environment teardown so that `--detect-async-leaks` reports a clean slate after each test file.

**Screenshots**
N/A — relevant stack frames from the leak output:
```
Timeout leaking
❯ Window.setTimeout  happy-dom/lib/window/BrowserWindow.js:1205
❯ @tanstack/virtual-core  observeElementOffset (layout effect)

PROMISE leaking
❯ Window.setTimeout  happy-dom/lib/window/BrowserWindow.js:1205
❯ Response.json → ReadableStream (internal fetch polyfill)
```

**Device:**
- OS: macOS / Linux (CI)
- Browser: N/A (Vitest / Node.js test environment)
- Version: happy-dom v20.x, Vitest v4.1-beta

**Additional context**
Both leak categories are owned by happy-dom internals:

- **`Timeout leaking`** — `BrowserWindow.setTimeout` is called during React's commit phase (e.g. virtualizer layout effects, search input debounces). These timers are never exposed to test code so there is no way to call `clearTimeout` on them in `afterEach`.
- **`PROMISE leaking`** — happy-dom's `Response.json()` implementation leaves unresolved `ReadableStream` microtask chains after the response body is consumed. No userland workaround exists.

Switching to `jsdom` reduces some of these but introduces other tradeoffs. The fix needs to live in happy-dom's window teardown logic.

capricorn86:
[...]

PR #2114, opened 2026-03-25
fix: [#2099] Lazily create ReadableStream for Request and Response body

55728:
## Problem

When constructing a `Request` or `Response` with a body that can be synchronously buffered (e.g. `string`, `Buffer`, `Blob`, `ArrayBuffer`, `URLSearchParams`), `getBodyStream()` eagerly creates a `ReadableStream` via `toReadableStream()` even though the buffer is already available.

Methods like `text()`, `json()`, `arrayBuffer()`, and `blob()` check for the internal buffer first and skip stream consumption entirely when a buffer exists. This means the eagerly created `ReadableStream` is never read — it is constructed and then abandoned.

Node.js `ReadableStream` internally uses microtask-based state transitions, so abandoned streams leave unresolved promises. Vitest 4.1's `--detect-async-leaks` flag detects these as `PROMISE leaking`.

## Solution

- **`FetchBodyUtility.getBodyStream()`**: Return `stream: null` when a `buffer` is already available (6 call sites).
- **`Response`**: Change `body` from a `public readonly` field to a getter backed by `[PropertySymbol.body]`. The getter lazily creates a `ReadableStream` from the buffer on first access, matching the pattern already used by `Request`.
- **`Request`**: Add the same lazy creation logic to the existing `body` getter.
- **`Request` constructor**: Replace `this.body` getter access with direct internal field checks (`[PropertySymbol.body]` / `[PropertySymbol.bodyBuffer]`) for `contentLength` determination, avoiding unnecessary stream creation.
- **`FetchRequestValidationUtility.validateBody()`**: Use internal fields instead of the `body` getter to validate GET/HEAD requests, preventing lazy getter side effects.
- **`FetchBodyUtility.cloneBodyStream()`**: Move the buffer check before the `body === null` check so that cloning a buffered body does not trigger the lazy getter. Also simplify the Request/Response branching — both now use `[PropertySymbol.body]`, so the conditional is no longer needed.

## Verification

Reproduction project using Vitest 4.1 with `--detect-async-leaks`:

```ts
// Before: PROMISE leaking
const response = new Response(JSON.stringify({ hello: 'world' }), {
  headers: { 'Content-Type': 'application/json' },
});
const data = await response.json();
```

| | Before | After |
|---|---|---|
| PROMISE leaks | 1 | **0** |
| Timeout leaks | 2 | 2 (unrelated, see below) |

All existing tests pass (341 tests in `test/fetch/`), plus 9 new tests verifying lazy body stream behavior:

- Stream is not created eagerly on construction
- Stream is created lazily on first `body` getter access
- `text()` / `json()` / `arrayBuffer()` consume the buffer without creating a stream
- `clone()` works correctly without triggering the lazy getter
- `body` returns `null` when no body is set

## Out of scope

The `Timeout leaking` reports in #2099 are caused by a different mechanism (environment teardown ordering with `asyncTaskManager`) and will need a separate fix.

Related: #2099

### DIFF
--- a/packages/happy-dom/src/fetch/Request.ts
+++ b/packages/happy-dom/src/fetch/Request.ts
@@ -116,7 +116,11 @@ export default class Request implements Request {
 
 		if (contentLength) {
 			this[PropertySymbol.contentLength] = contentLength;
-		} else if (!this.body && (this.method === 'POST' || this.method === 'PUT')) {
+		} else if (
+			!this[PropertySymbol.body] &&
+			!this[PropertySymbol.bodyBuffer] &&
+			(this.method === 'POST' || this.method === 'PUT')
+		) {
 			this[PropertySymbol.contentLength] = 0;
 		}
 
@@ -187,6 +191,11 @@ export default class Request implements Request {
 	 * @returns Body.
 	 */
 	public get body(): ReadableStream | null {
+		if (!this[PropertySymbol.body] && this[PropertySymbol.bodyBuffer]) {
+			this[PropertySymbol.body] = FetchBodyUtility.toReadableStream(
+				this[PropertySymbol.bodyBuffer]
+			);
+		}
 		return this[PropertySymbol.body];
 	}
 
--- a/packages/happy-dom/src/fetch/Response.ts
+++ b/packages/happy-dom/src/fetch/Response.ts
@@ -31,7 +31,6 @@ export default class Response implements Response {
 	protected declare [PropertySymbol.window]: BrowserWindow;
 
 	// Public properties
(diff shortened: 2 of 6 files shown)


55728:
**Note:** The failing test (`Fetch > send() > Revalidates cache with "If-Modified-Since"...`) is a pre-existing flaky test unrelated to this change — the same failure can be seen in [#1910's CI run](https://github.com/capricorn86/happy-dom/actions/runs/23460830862/job/68261775986).

55728:
Closing this as stale. Happy to reopen if needed. Thanks!

PR #2121, opened 2026-04-01
fix: [#2120] Use WeakMap for window-to-frame mapping to prevent leak

### LINKED ISSUE #2120: Window leaks memory
**Describe the bug**
`new Window()` is leaking memory.

**To Reproduce**
1. Create a new `Window`
2. Close the window
3. See used heap has gone up

https://codesandbox.io/p/sandbox/hungry-sun-pxs7g3

**Expected behavior**
Creating and closing a `Window` should not take any additional memory.

**Device:**
 - Node version 24.14.1
 - happy-dom version 20.8.

This also occurs when running the [server-side rendering simple example as per the documentation](https://github.com/capricorn86/happy-dom/wiki/Server%E2%80%90Side-Rendering#simple-example) like:
```ts

async function run() {
  console.log(process.memoryUsage().heapUsed);
  const window = new Window();
  await window.happyDOM.waitUntilComplete();
  await window.happyDOM.close();
}

while (true) {
  run();
}
```

etrepum:
The static Map in WindowBrowserContext held strong references from window internalId to browserFrame, preventing garbage collection of Window objects that were not explicitly closed via happyDOM.close().

Replace with a WeakMap keyed by the Window object itself so entries are automatically removed when the Window is GC'd. This also removes the unnecessary internalId indirection.

Closes #2120

(This PR was researched and authored with Claude, but it took a fair bit of human supervision)


etrepum:
Just checking up on this PR from 6 months ago - would be great to address this memory leak. We're using happy-dom in `@lexical/headless`. Let me know if there's anything I can do here to move things along.

PR #2134, opened 2026-04-13
fix: [#2088] Handle CSS hex escape sequences in querySelector

### LINKED ISSUE #2088: querySelector throws SyntaxError on selectors containing CSS hex escapes (e.g. #\39 7e...)
**Describe the bug**

`querySelector` and `querySelectorAll` throw a `SyntaxError` (or silently return `null`) when the selector contains CSS hex escape sequences produced by `CSS.escape()`.

When an element ID starts with a digit (e.g. a UUID like `97e356d3-...`), `CSS.escape()` correctly produces a hex escape: `\39 7e356d3-...`. Per [CSS Syntax 3 §4.3.7](https://drafts.csswg.org/css-syntax-3/#consume-escaped-code-point), the `\39` is 1-6 hex digits representing code point U+0039 (`'9'`), and the trailing space is a delimiter consumed by the tokenizer — it is **not** a descendant combinator.

happy-dom's `SelectorParser` uses `SELECTOR_GROUP_REGEXP` to split selectors at combinators, but this regex is not escape-aware. It sees the space inside `\39 ` and incorrectly treats it as a descendant combinator, splitting `#\39 7e356d3-...` into two fragments: `#\39` and `7e356d3-...`.

Additionally, `ESCAPED_CHARACTER_REGEXP` (`/\\/g`) only strips backslashes rather than decoding CSS hex escapes (`\39 ` → `9`), so even selectors that don't throw will never match.

The `SyntaxError` throw was introduced in v20.6.3 (commit e6a64da4, PR #2068). The underlying hex escape parsing bug existed before that, but manifested as a silent `null` return rather than a hard error.

**To Reproduce**

```js
const id = '97e356d3-601d-42ca-9c13-3446228274ac';

const container = document.createElement('div');
child.id = id;
container.appendChild(child);

// CSS.escape() produces: \39 7e356d3-...
container.querySelector(`#${CSS.escape(id)}`);
// Throws: SyntaxError: Failed to execute 'querySelectorAll' on 'Element':
//   '#\39 7e356d3-...' is not a valid selector.
```

A reproduction test is available at: https://stackblitz.com/edit/vitejs-vite-5gnp3mau?file=src%2Fcss-escape.test.ts

**Expected behavior**

1. `querySelector(`#${CSS.escape(id)}`)` should **not throw** — the selector is valid CSS.
2. It should **find and return the element** — a spec-compliant parser decodes `\39 ` as `'9'`, producing the ID `97e356d3-...`.

Both behaviors work correctly in Chrome, Firefox, and Safari.

**Device:**
- OS: Any (Node.js / Bun)
- Browser: happy-dom
- Version: >=20.6.3 throws `SyntaxError`; <20.6.3 silently returns `null` (doesn't throw but doesn't find the element either)

**Additional context**

Spec references:

- [CSS Syntax 3 §2.1 "Escaping"](https://drafts.csswg.org/css-syntax-3/#escaping) — `\` + 1-6 hex digits + optional whitespace
- [CSS Syntax 3 §4.3.7 "Consume an escaped code point"](https://drafts.csswg.org/css-syntax-3/#consume-escaped-code-point) — normative algorithm for decoding hex escapes
- [CSSOM §2.1 "Common Serializing Idioms"](https://drafts.csswg.org/cssom/#common-serializing-idioms) — defines "escape as code point" as `\` + hex + SPACE
- [CSSOM §8.1 "The CSS.escape() Method"](https://drafts.csswg.org/cssom/#the-css.escape()-method) — uses the above for leading digits

There are two things to fix:

1. **`SELECTOR_GROUP_REGEXP`** (or its replacement) must be escape-aware: when encountering `\` followed by 1-6 hex digits and an optional space, that space must not be treated as a combinator.
2. **`ESCAPED_CHARACTER_REGEXP`** must be replaced with a proper CSS hex escape decoder that converts `\XX ` to `String.fromCodePoint(parseInt('XX', 16))` per the spec, rather than just stripping backslashes.

**Workaround:** Use attribute selectors instead of `CSS.escape()`:
```js
container.querySelector(`[id="${id}"]`);
```

coffeeandwork:
querySelector and querySelectorAll throw a SyntaxError (or silently return null) when the selector contains CSS hex escape sequences like `#\39 7e356d3-...`, which is what `CSS.escape()` produces for IDs starting with a digit.

The root cause was in SelectorParser:

1. `SELECTOR_GROUP_REGEXP` wasn't escape-aware — it saw the trailing space in `\39 ` as a descendant combinator and split the selector in half.
2. `SELECTOR_REGEXP`'s ID/class/attribute patterns only handled single-character escapes (`\.`), not hex escapes (`\39 `), so they couldn't consume the full escaped identifier.
3. IDs, classes, and attribute names were "unescaped" by just stripping backslashes (`/\\/g`), which turned `\39 7e...` into `397e...` instead of decoding `\39` as the character `9`. The `cssUnescape()` method that handles this correctly already existed — it just wasn't being used for these cases.

### Changes

- Added a non-capturing escape sequence alternative to `SELECTOR_GROUP_REGEXP` so hex/character escapes are consumed before their trailing space can be misread as a combinator
- Updated `SELECTOR_REGEXP` to match hex escapes (`\\[0-9a-fA-F]{1,6}\s?`) in ID, class, attribute name, and unquoted attribute value patterns
- Switched ID/class/attribute name handling from the naive backslash strip to `cssUnescape()`
- Removed the now-unused `ESCAPED_CHARACTER_REGEXP`

### Testing

- Added tests for `CSS.escape()` with a UUID ID, hex escapes in ID selectors, and hex escapes in class selectors
- All existing querySelector tests still pass

Fixes #2088

### DIFF
--- a/packages/happy-dom/src/query-selector/SelectorParser.ts
+++ b/packages/happy-dom/src/query-selector/SelectorParser.ts
@@ -9,10 +9,12 @@ import NodeTypeEnum from '../nodes/node/NodeTypeEnum.js';
 /**
  * Selector group RegExp.
  *
+ * Non-capturing match: CSS escape sequences (hex or character), consumed to avoid
+ * misinterpreting the optional trailing space in hex escapes as a descendant combinator.
  * Group 1: Combinator (" ", ",", "+", ">", "̣~")
  * Group 2: Parentheses or brackets.
  */
-const SELECTOR_GROUP_REGEXP = /(\s*[\s,+>~]\s*)|([\[\]\(\)"'])/gm;
+const SELECTOR_GROUP_REGEXP = /(?:\\[0-9a-fA-F]{1,6}\s?|\\[^\n])|(\s*[\s,+>~]\s*)|([\[\]\(\)"'])/gm;
 
 /**
  * Selector RegExp.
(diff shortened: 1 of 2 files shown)


coffeeandwork:
The CI failure is in `SyncFetch.test.ts` — specifically the cache revalidation test that checks `If-Modified-Since` behavior with a small `max-age`. It's a timing issue: when CI is slow, the cache expires before the second request goes out, so the test sees stale headers (`max-age=0.0001`) instead of the revalidated ones (`max-age=1`).

This is a known flaky test — #2013 tracks it and there are already PRs (#2014, #2116) trying to fix it. Unrelated to the selector parser changes here. All 157 querySelector tests passed.

PR #2135, opened 2026-04-13
fix: [#2096] Add instance-level readyState enums to XMLHttpRequest

### LINKED ISSUE #2096: XMLHttpRequest does not implement instance-level `readyState` enums
**Describe the bug**

`XMLHttpRequest` instances constructed in HappyDOM do not implement the instance-level `readyState` enums at all (e.g. `xhr.DONE`, `xhr.OPENED`, etc). Those properties are not defined on the request instances and result in `undefined` always being returned.

**To Reproduce**

```js

const window = new Window()
const xhr = new window.XMLHttpRequest()
console.assert(xhr.DONE === 4)
```

**Expected behavior**

`xhr.DONE`, as well as all the other `readyState` enums, return the corresponding number value, never `undefined`.

coffeeandwork:
XMLHttpRequest only defined `UNSENT`, `OPENED`, `HEADERS_RECEIVED`, `LOADING`, and `DONE` as static properties. In browsers these are also available on instances (they sit on the prototype), so `xhr.DONE === 4` works. In happy-dom it returned `undefined`, which silently breaks the common pattern `xhr.readyState === xhr.DONE`.

Added readonly instance properties that mirror the static ones.

Fixes #2096

### Testing

Added tests verifying the instance properties return the correct values and match their static counterparts. All existing XMLHttpRequest tests still pass.

### DIFF
--- a/packages/happy-dom/src/xml-http-request/XMLHttpRequest.ts
+++ b/packages/happy-dom/src/xml-http-request/XMLHttpRequest.ts
@@ -34,12 +34,19 @@ export default class XMLHttpRequest extends XMLHttpRequestEventTarget {
 	// Injected by WindowContextClassExtender
 	protected declare [PropertySymbol.window]: BrowserWindow;
 
-	// Constants
+	// Static constants
 	public static UNSENT = XMLHttpRequestReadyStateEnum.unsent;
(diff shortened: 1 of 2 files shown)


bitboxer:
Can confirm, this fixes the issue we just ran into.

PR #2143, opened 2026-04-20
fix: [#2142] Cache parsed query selectors to mitigate a performance regression in 20.6.3

### LINKED ISSUE #2142: Performance regression in selector parsing since 20.6.3
**Describe the bug**

20.6.3 introduced changes to selector parsing that cause test suites using React Testing Library (and probably also DOM Testing Library) to run significantly slower than they did in 20.6.2.

I first noticed this performance regression in a (closed source) real-world project with thousands of RTL-based tests, where it resulted in a 2x slowdown across the full run and caused some tests to exceed timeouts, especially in a slow CI environment.

**To Reproduce**

See <https://github.com/rgrove/happy-dom-perf-regression> for a synthetic repro case that demonstrates the performance regression.

```
git clone git@github.com:rgrove/happy-dom-perf-regression.git
cd happy-dom-perf-regression

npm install
npm test # measure on happy-dom 20.9.0

npm install happy-dom@20.6.2
npm test # measure on happy-dom 20.6.2
```

The Vitest suite in that project (`Form.test.jsx`) renders a small form 100 times and runs typical React Testing Library queries against it. On my machine (Apple M1 Pro, Node.js 24.14.1), this results in the following test durations:

| happy-dom | Duration |
| --------- | -------- |
| 20.6.2    | 4.97s     |
| 20.9.0    | 8.46s     |

**Expected behavior**

happy-dom 20.6.3+ should have similar performance characteristics to 20.6.2.

**Additional context**

CPU profiling shows `SelectorParser.getSelectorGroups` consuming a majority of total runtime.

In `lib/query-selector/SelectorParser.js`, `getSelectorGroups` is invoked for every selector string on every `querySelectorAll` call (a new `SelectorParser` is constructed per call by `QuerySelector`). A "simple selector" fast path that short-circuited single-tag/class/id selectors in 20.6.2 was removed, and no caching of parsed selector groups was introduced to replace it. As a result, the same selector strings are re-parsed thousands of times during a typical test run.

happy-dom does appear to cache `querySelectorAll` results per node, but that cache is invalidated on every DOM mutation. React Testing Library mounts and unmounts components between tests, so the result cache provides little benefit and the selector-string parser bears the brunt of the work.

I have a working fix that addresses this by introducing a window-scoped selector group cache that avoids re-parsing selectors that can be safely cached. PR to follow.

rgrove:
This introduces a window-scoped LRU cache for parsed selectors to address the case described in #2142 that can cause the same selector strings to be re-parsed frequently in common use cases like unit tests in which a window is shared across tests but the contents of the DOM are cleaned up after each test.

To avoid unbounded growth, the cache is limited to 1,024 selectors, and is cleared when the window is destroyed. In my testing on a large project using React Testing Library, I found that a cache size of 1,024 seemed to be a sweet spot that achieved similar performance to 20.6.2 (prior to the regression).

Since `SelectorItem` instances carry scope references, we can't cache them directly. Instead, the cache stores a plain object (a "snapshot") with enough information to allow us to later instantiate a new `SelectorItem` with a different scope. Some pseudo-selectors with selector arguments (e.g. `:not`, `:has`, etc.) aren't safe to cache because they hold references to other `SelectorItem` instances, so we skip caching them.

Fixes #2142


capricorn86:
Thank you for your contribution @rgrove! In general it was a great approach, but I felt that I needed to start a separate branch to test various approaches to solve this because of the references to elements and parse scope.

I started with a RegExp cache, but in the end I also ended up with a `SelectorItem` cache, but where `SelectorItem` doesn't contain any references to elements and parse scope.

I merged the branch in PR #2214 now and will close this. Feel free to take a look at the PR and let me know if I missed something in my implementation.

rgrove:
Thanks @capricorn86! Your solution looks more comprehensive, and in my large private project where I first noticed the performance regression 20.10.5 appears to be slightly faster than my own patch. 🚀

PR #2158, opened 2026-05-20
fix: treat '<' followed by non-letter as text in HTML parser

### LINKED ISSUE #2156: Bug: `innerHTML` incorrectly parses `<` followed by a digit as an HTML tag
**Describe the bug**
 Bug: `innerHTML` incorrectly parses `<` followed by a digit as an HTML tag

  When setting `innerHTML` with HTML content that contains a `<` character immediately
  followed by a digit (e.g. `<3`), happy-dom incorrectly treats it as an opening HTML
  tag instead of plain text. This diverges from the WHATWG HTML5 parsing specification.

**To Reproduce**
  ```js

  const win = new Window();
  const doc = win.document;
  const container = doc.createElement('div');
  container.innerHTML = '<p>text <3 more</p>';

  const p = container.childNodes[0];
  console.log(p.childNodes.length);       // actual: 2
  console.log(p.childNodes[1].tagName);   // actual: "3"
```

**Expected behavior**
  Per the WHATWG HTML5 tokenization spec,
  when the parser is in the tag open state and encounters a character that is not
  an ASCII alpha, !, /, or ?, it must emit a U+003C LESS-THAN SIGN character
  token and reconsume the current character in the data state.

  p.childNodes.length   // → 1
  p.childNodes[0]       // → Text node: "text <3 more"

  This is the behaviour produced by both the browser's native DOMParser and parse5.

**Actual Behavior**
  happy-dom creates a child element with tagName = "3" inside the `<p>`, and
  swallows all subsequent content as its descendants. The resulting DOM tree is
  structurally invalid.

**Device:**
  - happy-dom version: 20.9.0 (latest)
  - Node.js version: 22.14.0

**Additional context**

### PR DESCRIPTION
**Pull request by mixelburg**
Fixes #2156

Per the WHATWG HTML5 tokenization spec, when the parser encounters `<` followed by a character that is not an ASCII alpha, `!`, `/`, or `?`, it should emit `<` as a character token and reconsume the character in the data state. Previously, the `MARKUP_REGEXP` pattern `<([^\s/!>?]+)` would match `<` followed by any non-whitespace character (including digits), incorrectly treating `<3` as an opening tag with name `"3"`.

This fix restricts tag name matching to start with an ASCII letter (`[a-zA-Z]`), which aligns with the HTML5 spec and matches browser behavior.

**Before:**
```js
container.innerHTML = '<p>text <3 more</p>';
// Creates a "<3>" element inside <p>, swallowing subsequent content
```

**After:**
```js
container.innerHTML = '<p>text <3 more</p>';
// Correctly produces a single text node: "text <3 more"
```

### DIFF
--- a/packages/happy-dom/src/html-parser/HTMLParser.ts
+++ b/packages/happy-dom/src/html-parser/HTMLParser.ts
@@ -32,7 +32,7 @@ import type Text from '../nodes/text/Text.js';
  * Group 7: End of self closing start tag (e.g. "/>" in "<img/>").
  * Group 8: End of start tag or comment tag (e.g. ">" in "<div>").
  */
-const MARKUP_REGEXP = /<([^\s/!>?]+)|<\/([^\s/!>?]+)\s*>|(<!--)|(-->|--!>)|(<!)|(<\?)|(\/>)|(>)/gm;
+const MARKUP_REGEXP = /<([a-zA-Z][^\s/!>?]*)|<\/([a-zA-Z][^\s/!>?]*)\s*>|(<!--)|(-->|--!>)|(<!)|(<\?)|(\/>)|(>)/gm;
 
 /**
  * Attribute RegExp.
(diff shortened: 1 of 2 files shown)


mixelburg:
Closing due to inactivity — over 40 days with no engagement. Happy to reopen if there's interest.

PR #2162, opened 2026-05-23
fix: [#2152] Fix TreeWalker.previousNode() skipping nodes when sibling has FILTER_SKIP children

### LINKED ISSUE #2152: TreeWalker.previousNode() returns null when previous sibling has skipped child nodes
**Describe the bug**
TreeWalker.previousNode() appears to incorrectly return null in cases where the previous visible node is an element that contains child nodes excluded by whatToShow.

This breaks libraries that rely on backward TreeWalker traversal, for example @react-aria/focus’s focusManager.focusPrevious(), which is used by react-aria-components Toolbar keyboard navigation. In our tests, ArrowRight navigation works, but ArrowLeft navigation fails because focusPrevious() receives null.

**Minimal reproduction**
const root = document.createElement('div');

root.innerHTML = `
  <button>Bold</button>
  <button>Italic</button>
`;

document.body.appendChild(root);

const buttons = root.querySelectorAll('button');

const walker = document.createTreeWalker(root, NodeFilter.SHOW_ELEMENT, {
  acceptNode(node) {
    return node.tagName === 'BUTTON'
      ? NodeFilter.FILTER_ACCEPT
      : NodeFilter.FILTER_SKIP;
  },
});

walker.currentNode = buttons[1];

const previous = walker.previousNode();

console.log(previous === buttons[0]); // expected: true, actual: false
console.log(previous); // expected: first button, actual: null

**Expected behavior**
walker.previousNode() should return the first <button> element.

**Actual behavior**
walker.previousNode() returns null.

**Environment:**
 - happy-dom: 20.3.7
 - Also reproduced by inspecting latest 20.9.0
 - Runtime: Vitest with environment: "happy-dom"

mixelburg:
Fixes #2152

`previousNode()` used `let node = sibling` inside the inner while loop, shadowing the outer `node` variable. After the loop finished, the outer code called `.parentNode` on the stale outer `node` (still `currentNode`) instead of on the deepest node visited in the sibling subtree.

Per the WHATWG spec the algorithm uses a single `node` variable throughout — removing `let` fixes the shadowing and matches the spec behavior.

**Example** (from the issue):

```js
const root = document.createElement('div');
root.innerHTML = '<button>Bold</button><button>Italic</button>';
const buttons = root.querySelectorAll('button');

const walker = document.createTreeWalker(root, NodeFilter.SHOW_ELEMENT, {
  acceptNode: node => node.tagName === 'BUTTON' ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_SKIP
});

walker.currentNode = buttons[1];
walker.previousNode(); // returned null — now correctly returns buttons[0]
```

All 20 existing TreeWalker tests pass. Added a regression test.

### DIFF
--- a/packages/happy-dom/src/tree-walker/TreeWalker.ts
+++ b/packages/happy-dom/src/tree-walker/TreeWalker.ts
@@ -133,7 +133,7 @@ export default class TreeWalker {
 			let sibling = node?.previousSibling || null;
 
 			while (sibling !== null) {
-				let node = sibling;
+				node = sibling;
 				let result = this[PropertySymbol.filterNode](node);
 
 				while (result !== NodeFilter.FILTER_REJECT && node[PropertySymbol.nodeArray].length) {
--- a/packages/happy-dom/test/tree-walker/TreeWalker.test.ts
+++ b/packages/happy-dom/test/tree-walker/TreeWalker.test.ts
@@ -284,6 +284,31 @@ describe('TreeWalker', () => {
 				expectedPreviousNode = currentNode;
 			}
 		});
+
+		it('Returns correct previous node when previous sibling has FILTER_SKIP children.', () => {
+			const root = document.createElement('div');
+			root.innerHTML = `
+				<button>Bold</button>
+				<button>Italic</button>
+			`;
+			document.body.appendChild(root);
+
+			const buttons = root.querySelectorAll('button');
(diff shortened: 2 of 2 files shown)


mixelburg:
Closing due to inactivity (37 days with no reviews). Happy to reopen if there's interest.

PR #2183, opened 2026-06-05
fix: [#2182] Node.prototype.nodeName getter returns correct value for subclass instances

### LINKED ISSUE #2182: `Node.prototype.nodeName` getter returns `''` when invoked on an element (breaks DOMPurify ≥ 3.4.8 anti-clobbering)
**Describe the bug**

happy-dom defines an **own** `nodeName` getter on **both** `Node.prototype` and `Element.prototype`. Normal property access (`el.nodeName`) resolves to the `Element.prototype` getter and returns the correct value (`"P"`). However, the getter installed on **`Node.prototype`** returns an **empty string** when invoked directly on an element via `Object.getOwnPropertyDescriptor(Node.prototype, 'nodeName').get.call(el)`.

This matters because security libraries deliberately read `nodeName` through the un-clobberable `Node.prototype` getter rather than via the (clobberable) instance property. **DOMPurify 3.4.8** (released 2026-06-03) did exactly this as part of a security fix (GHSA-x4vx-rjvf-j5p4 — *"IN_PLACE mode trusts attacker-controlled `nodeName` … allowing script retention and XSS"*):

```js
// DOMPurify 3.4.8, dist/purify.js
const getNodeName = Node && Node.prototype ? lookupGetter(Node.prototype, 'nodeName') : null;
// ...
const tagName = transformCaseFunc(getNodeName ? getNodeName(currentNode) : currentNode.nodeName);
```

Under happy-dom, `getNodeName(currentNode)` returns `''` for every element, so DOMPurify computes an empty tag name for every node and its entire `ALLOWED_TAGS` allow/block logic misfires: allowed tags are dropped and disallowed tags (including XSS vectors) are retained. jsdom and real browsers are unaffected.

This is a continuation of the anti-clobbering work started in #1165, which exposed `cloneNode`/`nextSibling`/`childNodes`/`parentNode` on `Node.prototype` but did not cover `nodeName` (DOMPurify only began reading `nodeName` this way in 3.4.8).

**To Reproduce**

Minimal reproduction — the core defect (happy-dom only, no DOMPurify)

```js
import { Window } from 'happy-dom' // 20.10.1

const window = new Window()
const el = window.document.createElement('p')

console.log(el.nodeName) // "P"  ✅ (Element.prototype getter)

const desc = Object.getOwnPropertyDescriptor(window.Node.prototype, 'nodeName')
console.log(typeof desc.get)     // "function"
console.log(desc.get.call(el))   // ""  ❌  expected "P"
```

Both `Node.prototype` and `Element.prototype` have their **own** `nodeName` getter; the `Node.prototype` one returns `''` when called on an element.

End-to-end reproduction (with DOMPurify 3.4.8)

```js
import createDOMPurify from 'dompurify' // 3.4.8
import { Window } from 'happy-dom'      // 20.10.1
import { JSDOM } from 'jsdom'

const input = '<p><strong>boop</strong> <img src=x onerror=alert(1)></p>'
const cfg = { ALLOWED_TAGS: ['p', 'strong'] }

const hd = createDOMPurify(new Window())
console.log('happy-dom:', hd.sanitize(input, cfg))
// happy-dom: <strong>boop</strong> <img src="x" onerror="alert(1)">
//            ^ allowed <p> dropped, disallowed <img onerror> retained

const jd = createDOMPurify(new JSDOM('').window)
console.log('jsdom    :', jd.sanitize(input, cfg))
// jsdom    : <p><strong>boop</strong> </p>   ✅ correct
```

**Expected behavior**

The `nodeName` getter on `Node.prototype` should return the node's correct name (`"P"`) when invoked on an element, matching browsers and jsdom. Per #1165, properties that DOMPurify reads for anti-clobbering must be functions/getters on `Node.prototype` that work when called on any `Node` subclass instance and must not be shadowed by a child-class override that the base getter cannot fall through to.

**Screenshots**

**Device:**

- happy-dom: **20.10.1** (also reproduces on 20.9.0)
- dompurify: **3.4.8** (3.4.7 is unaffected — it read `currentNode.nodeName` directly; 3.4.8 switched to the `Node.prototype` getter)
- node: **v24.16.0**

**Additional context**

- DOMPurify's `Node.prototype` read is intentional and security-motivated (GHSA-x4vx-rjvf-j5p4); reverting it isn't an option on their side. The DOMPurify maintainer's in-progress PR cure53/DOMPurify#1443 carries a commit explicitly noting happy-dom support is *"too broken still,"* and does **not** change the `getNodeName` logic — so the fix needs to land here.
- Related: #1165 (anti-clobbering Node prototype properties), #1629 (Support for DOMPurify), #1403, #1810.

mixelburg:
Fixes #2182

The `Node.prototype.nodeName` getter returned an empty string for all node types. When libraries like DOMPurify extract the getter from `Node.prototype` directly (via `Object.getOwnPropertyDescriptor`) and call it on an element, the result was always `''` instead of the tag name.

This happened because the base `Node` getter was hardcoded to `return ''`, and while subclasses override it, those overrides don't apply when the getter is called from `Node.prototype` directly.

Fixed by making `Node.prototype.nodeName` dispatch based on `nodeType`:
- Element: returns `tagName`
- Text: `#text`
- CDATASection: `#cdata-section`
- ProcessingInstruction: returns `target`
- Comment: `#comment`
- Document: `#document`
- DocumentType: returns `name`
- DocumentFragment: `#document-fragment`

All existing tests pass (4007 passed, 14 pre-existing failures unrelated to this change). Added regression test verifying the getter returns correct values when extracted from `Node.prototype` and called on subclass instances.

### DIFF
--- a/packages/happy-dom/src/nodes/node/Node.ts
+++ b/packages/happy-dom/src/nodes/node/Node.ts
@@ -26,6 +26,8 @@ import type HTMLTextAreaElement from '../html-text-area-element/HTMLTextAreaElem
 import type HTMLSlotElement from '../html-slot-element/HTMLSlotElement.js';
 import NodeFactory from '../NodeFactory.js';
 import type SVGStyleElement from '../svg-style-element/SVGStyleElement.js';
+import type ProcessingInstruction from '../processing-instruction/ProcessingInstruction.js';
+import type DocumentType from '../document-type/DocumentType.js';
 
 /**
  * Node.
@@ -230,7 +232,26 @@ export default class Node extends EventTarget {
 	 * @returns Node name.
 	 */
 	public get nodeName(): string {
-		return '';
+		switch (this[PropertySymbol.nodeType]) {
+			case NodeTypeEnum.elementNode:
+				return (<Element>(<unknown>this))[PropertySymbol.tagName] ?? '';
+			case NodeTypeEnum.textNode:
+				return '#text';
+			case NodeTypeEnum.cdataSectionNode:
+				return '#cdata-section';
+			case NodeTypeEnum.processingInstructionNode:
+				return (<ProcessingInstruction>(<unknown>this))[PropertySymbol.target];
+			case NodeTypeEnum.commentNode:
+				return '#comment';
+			case NodeTypeEnum.documentNode:
+				return '#document';
+			case NodeTypeEnum.documentTypeNode:
+				return (<DocumentType>(<unknown>this)).name;
+			case NodeTypeEnum.documentFragmentNode:
+				return '#document-fragment';
+			default:
+				return '';
+		}
 	}
 
 	/**
--- a/packages/happy-dom/test/nodes/node/Node.test.ts
+++ b/packages/happy-dom/test/nodes/node/Node.test.ts
@@ -154,6 +154,25 @@ describe('Node', () => {
 		it('Returns emptry string.', () => {
 			expect(NodeFactory.createNode(document, Node).nodeName).toBe('');
 		});
+
+		it('Returns correct value when getter is extracted from Node.prototype and called on subclasses.', () => {
+			const nodeProtoGetter = Object.getOwnPropertyDescriptor(Node.prototype, 'nodeName')!.get!;
+
+			const div = document.createElement('div');
+			expect(nodeProtoGetter.call(div)).toBe('DIV');
+
+			const text = document.createTextNode('hello');
+			expect(nodeProtoGetter.call(text)).toBe('#text');
+
+			const comment = document.createComment('test');
+			expect(nodeProtoGetter.call(comment)).toBe('#comment');
+
+			expect(nodeProtoGetter.call(document)).toBe('#document');
+
+			if (document.doctype) {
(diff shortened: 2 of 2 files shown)


mixelburg:
Closing as this hasn't received any feedback. Happy to reopen if there's interest.

ElVisPL:
@mixelburg there is not feedback for any PRs currently, I think this one is really needed

mvolkert:
Hi,

I'm still affected by the issue discussed in #2182.

After upgrading DOMPurify from 3.4.7 to 3.4.8, our Vitest tests running with happy-dom started failing, while the exact same code still works correctly in a real browser. The regression appears exactly between 3.4.7 and 3.4.8. 【1-855cdd】【2-52f05f】

One example:

Input:
text1<br>text2

Expected:
text1<br>text2

Actual:
<br>text2

The leading text node disappears.

I also updated happy-dom to 20.13.2, but the behavior is unchanged.

Since this still looks related to the nodeName handling described in #2182, would you consider reopening the issue/PR for another look?

Thanks!

PR #2189, opened 2026-06-06
fix: [#2184] Exclude disabled select and textarea elements from FormData

### LINKED ISSUE #2184: FormData constructor does not exclude disabled <select> and <textarea> elements
**Describe the bug**

When a `FormData` object is constructed from an `HTMLFormElement`, disabled `<select>` and `<textarea>` controls are still included in the entry list. The HTML standard requires *all* disabled submittable elements to be skipped. happy-dom already excludes disabled `<input>` elements, but not disabled `<select>`/`<textarea>`, so the behavior is inconsistent across control types.

**To Reproduce**

```js

const window = new Window();

const form = document.createElement("form");
form.innerHTML = `
  <select name="sel" disabled><option value="x" selected>x</option></select>
  <textarea name="ta" disabled>hello</textarea>
  <input name="in" type="text" value="i" disabled>
`;

console.log([...new window.FormData(form).keys()]);
// happy-dom 20.10.1 → [ 'sel', 'ta' ]   ❌ disabled <select>/<textarea> included
//                                          (note: the disabled <input> IS correctly excluded)
// Chrome/Firefox/Safari & jsdom → [ ]    ✅ all disabled controls excluded
```

**Expected behavior**

Disabled controls of every type are excluded from the constructed entry list, so `[...new window.FormData(form).keys()]` is `[]`. Verified in Chrome, and in jsdom 29 (which returns `['sel2']` when an extra enabled `<select name="sel2">` is added — i.e. only the enabled control is kept).

**Spec references**

- [XHR §`FormData()` constructor](https://xhr.spec.whatwg.org/#dom-formdata) → "set this's entry list to the result of *constructing the entry list* for form and submitter."
- [HTML §Constructing the entry list](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-the-form-data-set), step 5.1: *"If any of the following is true … **The field element is disabled.** … then continue."* — this applies to every submittable element (`input`, `select`, `textarea`, `button`), not only `input`.
- [HTML §concept of being "disabled"](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-fe-disabled) — an element is disabled if it has the `disabled` attribute **or** is a descendant of a `<fieldset disabled>` (outside its first `<legend>`).

**Root cause**

[`packages/happy-dom/src/form-data/FormData.ts`](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/form-data/FormData.ts#L101-L104) — the `INPUT` branch guards on `disabled`, but the `TEXTAREA`/`SELECT` branch appends unconditionally:

```ts
case 'TEXTAREA':
case 'SELECT':
    this.append(name, (<HTMLInputElement>item).value); // no disabled check
    break;
```

Both `HTMLSelectElement` and `HTMLTextAreaElement` already expose a `disabled` getter, so the data is available.

**Suggested fix**

Minimal, targeted:

```ts
case 'TEXTAREA':
case 'SELECT':
    if ((<HTMLInputElement>item).disabled) {
        break;
    }
    this.append(name, (<HTMLInputElement>item).value);
    break;
```

Cleaner, spec-aligned alternative — since "disabled" applies uniformly to all submittable elements, hoist a single check above the `switch` and drop the per-`INPUT` one:

```ts
for (const item of items) {
    const name = item.name;
    if (!name || (<HTMLInputElement>item).disabled) {
        continue;
    }
    switch (item[PropertySymbol.tagName]) { /* … */ }
}
```

(Related, out of scope here: `item.disabled` reads only the element's own attribute, so a control inside a `<fieldset disabled>` is still wrongly included. A fully spec-compliant fix would consult the element's inherited "actually disabled" state per the concept linked above.)

**Suggested test** (mirrors the existing *"Supports disabled input elements."* test in `test/form-data/FormData.test.ts`)

```ts
it('Excludes disabled <select> and <textarea> elements.', () => {

    const select = document.createElement('select');
    select.name = 'select';
    const option = document.createElement('option');
    option.value = 'option value';
    option.selected = true;
    select.appendChild(option);
    select.disabled = true;

    const textarea = document.createElement('textarea');
    textarea.name = 'textarea';
    textarea.value = 'textarea value';
    textarea.disabled = true;

    form.appendChild(select);
    form.appendChild(textarea);

    const formData = new window.FormData(form);

    expect(formData.get('select')).toBe(null);
    expect(formData.get('textarea')).toBe(null);
});
```

**Workaround**

Wrap `FormData` to blank the `name` of disabled `<select>`/`<textarea>` during construction, then restore it:

```js
const OriginalFormData = window.FormData;
window.FormData = class extends OriginalFormData {
  constructor(form, submitter) {
    const renamed = [];
    if (form) {
      for (const el of form.elements) {
        if (el.disabled && el.name && (el.tagName === "SELECT" || el.tagName === "TEXTAREA")) {
          renamed.push([el, el.name]);
          el.removeAttribute("name");
        }
      }
    }
    try {
      super(form, submitter);
    } finally {
      for (const [el, name] of renamed) el.setAttribute("name", name);
    }
  }
};
```

**Environment**
- happy-dom: 20.10.1 (also reproduced on 20.9.0; code path unchanged on `main`)
- Node.js: 24
- Reference: Chrome / Firefox / Safari all exclude disabled controls; jsdom 29 matches

mixelburg:
Fixes #2184

The FormData constructor already excluded disabled `<input>` elements but skipped the disabled check for `<select>` and `<textarea>` controls. Added the same guard.

Per the spec, all disabled submittable elements should be excluded from the entry list regardless of element type.

### DIFF
--- a/packages/happy-dom/src/form-data/FormData.ts
+++ b/packages/happy-dom/src/form-data/FormData.ts
@@ -100,6 +100,9 @@ export default class FormData implements Iterable<[string, string | File]> {
 						break;
 					case 'TEXTAREA':
 					case 'SELECT':
+						if ((<HTMLInputElement>item).disabled) {
+							break;
+						}
 						this.append(name, (<HTMLInputElement>item).value);
 						break;
 				}
--- a/packages/happy-dom/test/form-data/FormData.test.ts
+++ b/packages/happy-dom/test/form-data/FormData.test.ts
@@ -307,6 +307,37 @@ describe('FormData', () => {
 			expect(formData.getAll('radioInput')).toEqual([]);
 			expect(formData.getAll('checkboxInput')).toEqual([]);
 		});
+
+		it('Excludes disabled select and textarea elements.', () => {
+			const form = document.createElement('form');
+
+			const select = document.createElement('select');
+			select.name = 'sel';
+			select.disabled = true;
+			const option = document.createElement('option');
+			option.value = 'x';
+			option.selected = true;
+			select.appendChild(option);
+
+			const textarea = document.createElement('textarea');
+			textarea.name = 'ta';
+			textarea.disabled = true;
+			textarea.value = 'hello';
+
+			const input = document.createElement('input');
+			input.type = 'text';
+			input.name = 'in';
+			input.value = 'i';
+			input.disabled = true;
+
+			form.appendChild(select);
+			form.appendChild(textarea);
+			form.appendChild(input);
+
+			const formData = new window.FormData(form);
+
+			expect([...formData.keys()]).toEqual([]);
+		});
 	});
 
 	describe('forEach()', () => {


mixelburg:

PR #2192, opened 2026-06-09
fix: [#2187] Return default validationMessage for built-in constraint violations

### LINKED ISSUE #2187: validationMessage returns an empty string for built-in constraint violations
> This is the `validationMessage` half of #941, filed separately as a focused, independently-fixable bug. (#941 also tracks a distinct problem — `ValidityState` properties not being enumerable — which is still open and not covered here.)

**Describe the bug**

`validationMessage` is only ever populated by `setCustomValidity()`. For a *built-in* constraint violation (e.g. an empty `required` field), it returns `""` even though the element is invalid. Browsers and jsdom return a non-empty message.

**To Reproduce**

```js

const window = new Window();
const input = window.document.createElement("input");
input.required = true;
input.value = "";

console.log(input.checkValidity());      // false
console.log(input.validity.valueMissing); // true
console.log(JSON.stringify(input.validationMessage));
// happy-dom 20.10.1 → ""                         ❌
// jsdom 29          → "Constraints not satisfied" ✅
// browsers          → locale-specific, e.g. "Please fill out this field." (non-empty)
```

**Expected behavior**

A non-empty message when the element is a candidate for constraint validation and fails it. Real browsers use locale-specific text; jsdom uses the locale-independent `"Constraints not satisfied"`, which is the pragmatic value for a test DOM (verified in jsdom 29 for `input`, `textarea`, and `select`; jsdom returns `""` for a barred element such as a disabled control).

**Spec references**

- [HTML §The constraint validation API — `validationMessage`](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage): *"If the element is not a candidate for constraint validation or if the element satisfies its constraints, **the attribute must return the empty string**. Otherwise, the attribute must return a suitably localized message that the user agent would show the user…"* — i.e. non-empty when invalid **and** a validation candidate.
- [HTML §`willValidate`](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-willvalidate) — defines "candidate for constraint validation" (false for disabled/readonly/etc.).
- jsdom precedent (cited in #941): [`DefaultConstraintValidation-impl.js`](https://github.com/jsdom/jsdom/blob/main/lib/jsdom/living/constraint-validation/DefaultConstraintValidation-impl.js) returns `"Constraints not satisfied"`.

**Root cause**

The `validationMessage` getter on `HTMLInputElement`, `HTMLTextAreaElement`, and `HTMLSelectElement` returns a stored field that only `setCustomValidity()` writes — e.g. [`HTMLInputElement.ts`](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts#L277-L279):

```ts
public get validationMessage(): string {
    return this[PropertySymbol.validationMessage]; // only set by setCustomValidity()
}
```

Built-in violations (reflected correctly in `validity`) are never turned into a message.

**Suggested fix**

Fall back to a generic message when the element is a validation candidate and invalid, keeping any custom message. Applied to each of the three getters (or a shared helper):

```ts
public get validationMessage(): string {
    const customMessage = this[PropertySymbol.validationMessage];
    if (customMessage) {
        return customMessage; // from setCustomValidity()
    }
    // Empty for elements barred from constraint validation, or that are valid.
    if (!this.willValidate || this.validity.valid) {
        return '';
    }
    // Browsers return a locale-specific message; match jsdom's locale-independent default.
    return 'Constraints not satisfied';
}
```

The `!this.willValidate` gate is what keeps barred controls (e.g. disabled) returning `""` — confirmed necessary: a disabled `required`-empty `input` has `validity.valid === false` but `willValidate === false`.

⚠️ **One dependency:** `HTMLTextAreaElement` currently has **no `willValidate` getter** (`HTMLInputElement` and `HTMLSelectElement` do). It should be added there too (textarea is a validation candidate per spec regardless of this issue); otherwise the textarea getter must gate on `!this.validity.valid` alone.

**Suggested test** (extends the existing `describe('get validationMessage()')` in each element's test file)

```ts
it('Returns a non-empty message for a built-in constraint violation.', () => {
    element.required = true;
    element.value = '';
    expect(element.validity.valueMissing).toBe(true);
    expect(element.validationMessage).toBe('Constraints not satisfied');
});

it('Returns an empty message when valid.', () => {
    element.required = true;
    element.value = 'x';
    expect(element.validationMessage).toBe('');
});
```

**Workaround**

Override the getter to derive a message from `validity` (shown for `input`; same for `HTMLTextAreaElement`/`HTMLSelectElement`):

```js
const descriptor = Object.getOwnPropertyDescriptor(
  window.HTMLInputElement.prototype,
  "validationMessage",
);
const originalGet = descriptor.get;
Object.defineProperty(window.HTMLInputElement.prototype, "validationMessage", {
  configurable: true,
  get() {
    const message = originalGet.call(this);
    if (message) return message;
    if (this.validity && !this.validity.valid) return "Constraints not satisfied";
    return message;
  },
});
```

**Environment**
- happy-dom: 20.10.1 (also 20.9.0; code path unchanged on `main`)
- Node.js: 24
- Reference: jsdom 29 returns `"Constraints not satisfied"`; browsers return locale-specific non-empty text

mixelburg:
Fixes #2187

The `validationMessage` getter on HTMLInputElement, HTMLTextAreaElement, HTMLSelectElement, HTMLButtonElement, HTMLObjectElement, and HTMLOutputElement only returned the custom message set via `setCustomValidity()`. When the element is a validation candidate with a built-in constraint violation (e.g. `required` + empty value), it returned `""` instead of a descriptive message.

Per the [HTML spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-validationmessage), `validationMessage` must return a non-empty message when the element is a validation candidate and fails its constraints. This fix returns `'Constraints not satisfied'` in that case, matching jsdom's behavior.

The `ValidityState.customError` getter reads the stored field directly, so this change does not create a circular dependency — `customError` remains `false` for built-in violations.

Tests added for input (valueMissing), select (required with no value), and button (disabled) elements. All existing tests continue to pass.


capricorn86:
Thank you for your contribution! `npm run compile` is unfortunately failing

mixelburg:

capricorn86 (approved):

capricorn86:
I have fixed the remaining work now and it's ready to be released. It failed because `willValidate` was not implemented in some elements. I also added tests for it and updated some logic according to spec.

PR #2193, opened 2026-06-10
fix: [#2190] Run click listeners on disabled form controls when dispatched programmatically

### LINKED ISSUE #2190: Disabled <button>/<input> drops a scripted dispatchEvent('click') instead of running its listeners
**Describe the bug**

A `click` dispatched via `dispatchEvent()` on a disabled `<button>` or `<input>` is dropped entirely: `dispatchEvent` returns `false` and **no `click` listeners run**. Per the DOM standard, `dispatchEvent` has no disabled provision — it must always invoke listeners along the propagation path. The HTML standard only suppresses clicks *queued on the user interaction task source* (real user clicks) on disabled controls, and the `click()` method returns early for a disabled control — but a scripted `dispatchEvent(new MouseEvent('click'))` is neither of those. Chromium, Firefox, WebKit, and jsdom all run the listeners (and the activation behavior).

**To Reproduce**

```js

const window = new Window();
const button = window.document.createElement("button");
button.disabled = true;
let clicks = 0;
button.addEventListener("click", () => clicks++);
window.document.body.appendChild(button);

const returned = button.dispatchEvent(
  new window.MouseEvent("click", { bubbles: true, cancelable: true }),
);
console.log(clicks, returned);
// happy-dom 20.10.1              → 0 false   ❌ listener never runs
// Chromium/Firefox/WebKit & jsdom → 1 true    ✅
```

**Expected behavior**

The listeners run and `dispatchEvent` returns `true` (unless a listener calls `preventDefault()`), and the control's activation behavior runs — exactly as in real browsers. Verified across Chromium, Firefox, and WebKit (Playwright), and jsdom 29:

| `dispatchEvent('click')` on a disabled… | listener fires | returns | activation |
|---|---|---|---|
| `<button>` | ✅ | `true` | — |
| `<input type="checkbox">` | ✅ | `true` | **toggles** (unless `preventDefault`) |
| submit `<button>` inside a `<form>` | ✅ | `true` | no submission (disabled submitter) |

For contrast, the *synthetic-click* paths correctly **do** suppress on disabled in real browsers (and should remain so):
- `element.click()` → no-op (no listener, no toggle)
- clicking a `<label>` whose control is disabled → control not activated (0 clicks)

**Spec references**

- DOM — [`dispatchEvent()`](https://dom.spec.whatwg.org/#dom-eventtarget-dispatchevent) / [event dispatch](https://dom.spec.whatwg.org/#concept-event-dispatch): no disabled provision; listeners are invoked along the path regardless.
- HTML — [the `disabled` attribute](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-disabled): *"A form control that is disabled must prevent any click events that are **queued on the user interaction task source** from being dispatched on the element."* (only user-queued clicks, not scripted `dispatchEvent`).
- HTML — [the `click()` method](https://html.spec.whatwg.org/multipage/interaction.html#dom-click): *"…**If this element is a form control that is disabled, then return.** … Fire a synthetic pointer event named click…"* → the disabled gate belongs on the `click()` path, not on `dispatchEvent`.

**Root cause**

[`HTMLButtonElement.dispatchEvent`](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts#L357-L365) and [`HTMLInputElement.dispatchEvent`](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts#L1432-L1444) short-circuit with `return false` when the event is a click and the element is disabled — *before* `super.dispatchEvent` (which runs the listeners):

```ts
// HTMLInputElement.dispatchEvent
if (this.disabled) {
    return false; // listeners never run, activation never runs
}
```

`HTMLElement.click()` is implemented as `dispatchEvent(new PointerEvent('click'))`, and since `PointerEvent extends MouseEvent`, this short-circuit makes `.click()` a correct no-op on disabled — but it also catches a raw `dispatchEvent(new MouseEvent('click'))`, which must not be suppressed.

**Suggested fix**

Move the disabled gate off `dispatchEvent` and onto the synthetic-click path:

1. Remove the `disabled` short-circuit from `HTMLButtonElement.dispatchEvent` and `HTMLInputElement.dispatchEvent`, so listeners + activation run. For the button, gate only the submit/reset activation block with `!this.disabled`; the input's checkbox/radio toggle should run (browsers toggle).
2. Gate [`HTMLElement.click()`](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-element/HTMLElement.ts#L949-L956) on disabled instead, per the `click()` spec steps:
   ```ts
   public click(): void {
       if ((this as { disabled?: boolean }).disabled) return;
       this.dispatchEvent(new PointerEvent('click', { bubbles: true, composed: true, cancelable: true }));
   }
   ```
3. Also guard the label forwarding in [`HTMLLabelElement.dispatchEvent`](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElement.ts#L110-L127) — it forwards via `control.dispatchEvent(new MouseEvent('click'))` and currently relies on the very suppression being removed. Without a guard, clicking a label would start activating a disabled control. Add `!control.disabled` (or forward via `control.click()`):
   ```ts
   if (control && event.target !== control && !(control as { disabled?: boolean }).disabled) {
       control.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true }));
   }
   ```
   Verified: clicking a label whose control is disabled activates the control **0** times in Chromium/Firefox/WebKit.

(Existing tests that assert the disabled-drop on `dispatchEvent` in `HTMLButtonElement.test.ts` / `HTMLInputElement.test.ts` would need updating to the browser-accurate behavior.)

**Suggested test**

```ts
it('Runs click listeners on a disabled control (scripted dispatchEvent).', () => {
    input.type = 'checkbox';
    input.disabled = true;
    let clicks = 0;
    input.addEventListener('click', () => clicks++);
    document.body.appendChild(input);

    const returned = input.dispatchEvent(
        new window.MouseEvent('click', { bubbles: true, cancelable: true }),
    );

    expect(clicks).toBe(1);
    expect(returned).toBe(true);
    expect(input.checked).toBe(true); // activation still runs
});

it('click() is a no-op on a disabled control.', () => {
    input.type = 'checkbox';
    input.disabled = true;
    let clicks = 0;
    input.addEventListener('click', () => clicks++);
    document.body.appendChild(input);

    input.click();

    expect(clicks).toBe(0);
    expect(input.checked).toBe(false);
});

it('Forwards a label click to its control only when the control is enabled.', () => {
    const make = (disabled) => {
        const label = document.createElement('label');
        const checkbox = document.createElement('input');
        checkbox.type = 'checkbox';
        checkbox.disabled = disabled;
        label.append(checkbox, document.createTextNode('x'));
        document.body.appendChild(label);
        let clicks = 0;
        checkbox.addEventListener('click', () => clicks++);
        label.dispatchEvent(new window.MouseEvent('click', { bubbles: true, cancelable: true }));
        return { clicks, checked: checkbox.checked };
    };

    expect(make(false)).toEqual({ clicks: 1, checked: true });  // enabled  → forwarded
    expect(make(true)).toEqual({ clicks: 0, checked: false }); // disabled → not forwarded
});
```

The label-forwarding test is the regression guard for fix step 3 (the `!control.disabled` check in `HTMLLabelElement.dispatchEvent`): once `dispatchEvent` stops self-suppressing on disabled, only this guard keeps a label click from activating a disabled control — matching all three browsers (verified: 0 clicks).

**Workaround**

Route the disabled-control click through the base `EventTarget` dispatch so the listeners run:

```js
function clickDisabled(element, event) {
  if (
    event instanceof MouseEvent &&
    event.type === "click" &&
    (element instanceof HTMLButtonElement || element instanceof HTMLInputElement) &&
    element.disabled
  ) {
    return EventTarget.prototype.dispatchEvent.call(element, event);
  }
  return element.dispatchEvent(event);
}
```

(This restores listener invocation, but not the activation behavior — it won't toggle a checkbox — since the activation lives in the bypassed `HTMLInputElement.dispatchEvent`.)

**Environment**
- happy-dom: 20.10.1 (code path unchanged on `master`)
- Node.js: 24
- Reference: Chromium / Firefox / WebKit (Playwright) and jsdom 29 all run the listeners + activation

(Likely the same root cause as the older, narrower #546.)

mixelburg:
Fixes #2190

`dispatchEvent(new MouseEvent('click'))` on a disabled button/input was returning `false` without running any listeners, matching neither browsers nor the DOM spec. Per the HTML spec, only `element.click()` should suppress on disabled — `dispatchEvent()` has no disabled provision and must always invoke listeners.

Changes:
- **HTMLButtonElement / HTMLInputElement**: Removed the disabled early-return from `dispatchEvent`. Click listeners now run on disabled elements, but activation behavior (form submit/reset, checkbox toggle, input/change events) is skipped.
- **HTMLButtonElement / HTMLInputElement**: Added `click()` override that returns early when disabled, matching the spec's `click()` semantics.
- **HTMLLabelElement**: Added `!control.disabled` guard to prevent label clicks from forwarding to disabled controls.

All 426 existing tests pass + 5 new tests added.

### DIFF
--- a/packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts
+++ b/packages/happy-dom/src/nodes/html-button-element/HTMLButtonElement.ts
@@ -354,19 +354,18 @@ export default class HTMLButtonElement extends HTMLElement {
 	/**
 	 * @override
 	 */
-	public override dispatchEvent(event: Event): boolean {
-		if (
-			event.type === 'click' &&
-			event instanceof MouseEvent &&
-			event.eventPhase === EventPhaseEnum.none &&
-			this.disabled
-		) {
-			return false;
+	public override click(): void {
+		if (this.disabled) {
+			return;
 		}
+		super.click();
+	}
 
+	public override dispatchEvent(event: Event): boolean {
 		const returnValue = super.dispatchEvent(event);
 
 		if (
+			!this.disabled &&
 			!event[PropertySymbol.defaultPrevented] &&
 			event.type === 'click' &&
 			event.eventPhase === EventPhaseEnum.none &&
--- a/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
+++ b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
@@ -1429,6 +1429,13 @@ export default class HTMLInputElement extends HTMLElement {
 	/**
 	 * @override
 	 */
+	public override click(): void {
+		if (this.disabled) {
+			return;
+		}
+		super.click();
+	}
+
 	public override dispatchEvent(event: Event): boolean {
 		if (
 			event[PropertySymbol.type] !== 'click' ||
@@ -1438,11 +1445,6 @@ export default class HTMLInputElement extends HTMLElement {
 			return super.dispatchEvent(event);
 		}
 
-		// Do nothing if the input element is disabled and the event is a click event.
-		if (this.disabled) {
-			return false;
-		}
-
 		let previousCheckedValue: boolean | null = null;
 		let previousIndeterminateValue: boolean = this[PropertySymbol.indeterminate];
 
@@ -1450,7 +1452,7 @@ export default class HTMLInputElement extends HTMLElement {
 		// However, the value has to be restored if preventDefault() is called on the click event.
 		const type = this.type;
 
-		if (type === 'checkbox' || type === 'radio') {
+		if (!this.disabled && (type === 'checkbox' || type === 'radio')) {
 			previousCheckedValue = this.checked;
 			this.#setChecked(type === 'checkbox' ? !previousCheckedValue : true);
 			if (type === 'checkbox') {
@@ -1462,7 +1464,7 @@ export default class HTMLInputElement extends HTMLElement {
 		// Dispatches the event
 		const returnValue = super.dispatchEvent(event);
 
-		if (
+		if (this.disabled ||
 			event[PropertySymbol.type] !== 'click' ||
 			event[PropertySymbol.eventPhase] !== EventPhaseEnum.none ||
 			!(event instanceof MouseEvent)
--- a/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElement.ts
+++ b/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElement.ts
@@ -118,7 +118,7 @@ export default class HTMLLabelElement extends HTMLElement {
 			event instanceof MouseEvent
 		) {
 			const control = this.control;
-			if (control && event.target !== control) {
+			if (control && !control.disabled && event.target !== control) {
 				control.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true }));
 			}
 		}
--- a/packages/happy-dom/test/nodes/html-button-element/HTMLButtonElement.test.ts
+++ b/packages/happy-dom/test/nodes/html-button-element/HTMLButtonElement.test.ts
@@ -536,5 +536,48 @@ describe('HTMLButtonElement', () => {
 
 			expect(resetTriggeredCount).toBe(1);
 		});
+
+		it('Runs click listeners on disabled button when using dispatchEvent.', () => {
+			const button = <HTMLButtonElement>document.createElement('button');
+			button.disabled = true;
+			let clicks = 0;
+			button.addEventListener('click', () => clicks++);
+			document.body.appendChild(button);
+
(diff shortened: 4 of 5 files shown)


capricorn86:
Thank you for your contribution! `npm run compile` is unfortunately failing

mixelburg:

PR #2207, opened 2026-06-13
fix: [#2206] Coerce null qualifiedName to empty string in createDocument

### LINKED ISSUE #2206: createDocument(null, null, null) throws TypeError in 20.10.2 (worked in 20.10.1)
**Describe the bug**

After upgrading to 20.10.2, calling `document.implementation.createDocument(null, null, null)` throws:

```
TypeError: Cannot read properties of null (reading 'split')
    at DOMImplementation.createDocument (node_modules/happy-dom/lib/dom-implementation/DOMImplementation.js:30:31)
```

I bisected the npm releases: 20.10.0 and 20.10.1 are fine, it broke in 20.10.2.

The new `qualifiedName.split(':')` switch in `src/dom-implementation/DOMImplementation.ts` doesn't account for a null qualifiedName. Per the DOM spec the parameter is declared as `[LegacyNullToEmptyString] DOMString qualifiedName`, so null is supposed to be coerced to `""` rather than rejected. Chrome, Firefox and Safari all accept this call and return an XML document with no document element.

We ran into it because some code under test calls `createDocument(null, null, null)` to get an empty scratch document, which has always worked in browsers.

**To Reproduce**

```js
import { Window } from "happy-dom"; // 20.10.2

const window = new Window();
window.document.implementation.createDocument(null, null, null);
// TypeError: Cannot read properties of null (reading 'split')
```

**Expected behavior**

Returns a document instead of throwing, same as 20.10.1 and real browsers (null qualifiedName behaves like an empty string).

**Device:**
 - OS: Linux
 - Node v24.10.0

**Additional context**

Spec references: https://dom.spec.whatwg.org/#dom-domimplementation-createdocument and https://webidl.spec.whatwg.org/#LegacyNullToEmptyString

Firer:
Fixes #2206.

`document.implementation.createDocument(null, null, null)` started throwing `TypeError: Cannot read properties of null (reading 'split')` in 20.10.2. It worked in 20.10.1 and works in Chrome, Firefox and Safari.

The `createDocument` rewrite in #2188 added a `qualifiedName.split(':')` switch but dereferences `qualifiedName` directly. Per the DOM spec the parameter is declared [`[LegacyNullToEmptyString] DOMString`](https://webidl.spec.whatwg.org/#LegacyNullToEmptyString), so `null` should be coerced to `""` rather than rejected (see the [createDocument steps](https://dom.spec.whatwg.org/#dom-domimplementation-createdocument)).

This coerces `null` to an empty string before the switch — matching the existing pattern in `MediaList.ts` — and widens the parameter types to accept `null`. An empty/`null` qualified name falls through to the default branch and returns an `HTMLDocument`, restoring the 20.10.1 behaviour.

One note on scope: the spec has `createDocument` always return an `XMLDocument`, but the current implementation is an intentional stub (`TODO: Not fully implemented`) that keys off the qualified name and returns `HTMLDocument` by default. I kept that behaviour so this stays a focused regression fix rather than a rework of document-type selection — happy to follow up separately if you'd like it to track the spec more closely.

While I was in the file I also fixed the neighbouring `"xml"` test, which passed `"svg"` and so never actually exercised the `xml` branch.

### Testing

Added a regression test for `createDocument(null, null, null)`. It fails on `master` with the reported `TypeError` and passes with this change. Lint and the `DOMImplementation` suite are green.

### DIFF
--- a/packages/happy-dom/src/dom-implementation/DOMImplementation.ts
+++ b/packages/happy-dom/src/dom-implementation/DOMImplementation.ts
@@ -27,12 +27,21 @@ export default class DOMImplementation {
 	 * @param qualifiedName Qualified name.
 	 * @param [_docType] Document type.
 	 */
-	public createDocument(_namespaceURI: string, qualifiedName: string, _docType?: string): Document {
+	public createDocument(
+		_namespaceURI: string | null,
+		qualifiedName: string | null,
+		_docType?: string | null
+	): Document {
 		if (arguments.length < 2) {
 			throw new this.#document[PropertySymbol.window].TypeError(
 				`Failed to execute 'createDocument' on 'DOMImplementation': 2 arguments required, but only ${arguments.length} present.`
 			);
 		}
+
+		// The spec declares "qualifiedName" as "[LegacyNullToEmptyString] DOMString",
+		// so "null" is coerced to an empty string instead of being rejected.
+		qualifiedName = qualifiedName === null ? '' : String(qualifiedName);
+
 		switch (qualifiedName.split(':').pop()) {
 			case 'svg':
 			case 'xml':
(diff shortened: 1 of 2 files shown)


capricorn86 (approved):
Thank you for your contribution @Firer!

PR #2212, opened 2026-06-15
fix: [#2203] Allow replaceWith() on document element

### LINKED ISSUE #2203: HTMLElement::replaceWith() doesn't work
```js
const parsed = new DOMParser().parseFromString(html, 'text/html');

container.documentElement.replaceWith(
  container.importNode(parsed.documentElement, true)
);
```
```
Error: Failed to execute 'insertBefore' on 'Node': Only one element on document allowed.
```

**Device:**
 - Version 20.9.0

mixelburg:
Fixes #2203

`document.documentElement.replaceWith(newElement)` throws "Only one element on document allowed" because `ChildNodeUtility.replaceWith()` calls `insertBefore(newNode, oldNode)` before `removeChild(oldNode)`. The Document's `insertBefore` check rejects a second element child even though the old one is about to be removed.

Per the DOM spec, `replaceWith` uses the internal insert algorithm which doesn't enforce the one-element-per-document constraint. The fix adds a check: when the reference node in `insertBefore` is itself an element, skip the one-element guard since the reference node will be removed (as in the replaceWith/replaceChild flow).

### DIFF
--- a/packages/happy-dom/src/nodes/html-document/HTMLDocument.ts
+++ b/packages/happy-dom/src/nodes/html-document/HTMLDocument.ts
@@ -70,7 +70,8 @@ export default class HTMLDocument extends Document {
 
 		if (
 			newNode[PropertySymbol.nodeType] === NodeTypeEnum.elementNode &&
-			this[PropertySymbol.elementArray].length !== 0
+			this[PropertySymbol.elementArray].length !== 0 &&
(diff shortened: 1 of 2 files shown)


mixelburg:
Closing due to inactivity (40 days with no reviews). Happy to reopen if there's interest.

stevenvachon:
@capricorn86

PR #2216, opened 2026-06-17
fix: [#2213] Expose waitForNavigation() on window.happyDOM

### LINKED ISSUE #2213: Expose `waitForNavigation()` on `window.happyDOM`
**Describe the bug**

`waitForNavigation()` exists on `DetachedBrowserFrame` but is not exposed through `DetachedWindowAPI` (`window.happyDOM`).

`DetachedWindowAPI` already exposes `waitUntilComplete()` by delegating to the frame, so `waitForNavigation()` should be consistent with that.

As an alternative, `waitForNavigation()` should be exposed _somewhere_ that it can be called from a Vitest context.

**To Reproduce**

1. Create a `Window` instance (detached mode, as Vitest does with `environment: 'happy-dom'`)
2. Attempt to call `window.happyDOM.waitForNavigation()`
3. The method does not exist

**Expected behavior**

`waitForNavigation()` should be exposed on `DetachedWindowAPI` as `window.happyDOM.waitForNavigation()`, allowing Vitest users to write:

```typescript
await window.happyDOM.waitForNavigation();
```

instead of polling:

```typescript
await vi.waitUntil(() => new URL(window.location.href).hostname === 'example.com');
```

**Device and details:**

- OS: Linux
- Node version: 26.1.0
- Package version: 20.9.0

**Additional context**

This came up while testing an OAuth2 browser-side library with Vitest. The library triggers a redirect via `window.location.href` as a side effect of an async call, and there is currently no direct API available to await that navigation completing.

mixelburg:
Fixes #2213

`waitForNavigation()` exists on `DetachedBrowserFrame` and `DetachedBrowserPage` but was not exposed through `DetachedWindowAPI` (`window.happyDOM`). This made it inaccessible from Vitest's `happy-dom` environment where the only entry point to the browser API is `window.happyDOM`.

Added `waitForNavigation()` to `DetachedWindowAPI`, delegating to `this.#browserFrame.waitForNavigation()` — the same delegation pattern already used by `waitUntilComplete()`, `abort()`, and `close()`.

### DIFF
--- a/packages/happy-dom/src/window/DetachedWindowAPI.ts
+++ b/packages/happy-dom/src/window/DetachedWindowAPI.ts
@@ -46,6 +46,15 @@ export default class DetachedWindowAPI {
 	}
 
 	/**
+	 * Returns a promise that is resolved when the frame has navigated and the response HTML has been written to the document.
+	 *
+	 * @returns Promise.
+	 */
+	public waitForNavigation(): Promise<void> {
+		return this.#browserFrame.waitForNavigation();
+	}
+
+	/**
 	 * Waits for all async tasks to complete.
 	 *
 	 * @deprecated Use waitUntilComplete() instead.


mixelburg:
Closing due to inactivity (38 days with no reviews). Happy to reopen if there's interest.

PR #2221, opened 2026-06-21
fix: [#2186] Batch requestAnimationFrame callbacks per frame with shared timestamp

### LINKED ISSUE #2186: requestAnimationFrame callbacks are not batched per frame
**Describe the bug**

Each `requestAnimationFrame` call schedules its **own** `setImmediate` and reads `performance.now()` **separately** when it fires. As a result, callbacks registered for the *same* frame:

1. receive **different timestamps** (the spec gives every callback in a frame one shared `now`), and
2. run in **separate tasks** instead of one atomic "run the animation frame callbacks" pass, so unrelated microtasks/queued work interleave between same-frame callbacks and one callback can observe another's mid-frame DOM mutations.

jsdom (`pretendToBeVisual`) and real browsers batch all of a frame's callbacks into a single pass with one timestamp.

**To Reproduce**

```js

const window = new Window();

const timestamps = [];
window.requestAnimationFrame((t) => timestamps.push(t));
window.requestAnimationFrame((t) => {
  timestamps.push(t);
  console.log("same frame, same timestamp?", timestamps[0] === timestamps[1], timestamps);
});
// happy-dom 20.10.1             → false  [ 111.000667, 111.013792 ]   ❌ two timestamps
// Chrome/Firefox/Safari & jsdom → true   [ 111.0,      111.0       ]   ✅ one timestamp
```

**Expected behavior**

All callbacks registered before a frame run together in one task, each invoked with the *same* timestamp; a callback registered *during* the run is deferred to the next frame. Verified against jsdom 29 (`pretendToBeVisual: true`) → `same timestamp: true`.

**Spec references**

- [HTML §`requestAnimationFrame(callback)`](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#dom-animationframeprovider-requestanimationframe) — *"Increment target's animation frame callback identifier by one, and let `handle` be the result … Return `handle`."* → the return value is an **`unsigned long` integer**.
- [HTML §run the animation frame callbacks](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#run-the-animation-frame-callbacks):
  1. Let `callbacks` be target's map of animation frame callbacks.
  2. Let `callbackHandles` be the result of getting the keys of `callbacks`. *(snapshot — callbacks added during the run are deferred to the next frame)*
  3. For each `handle` in `callbackHandles`, if it still exists: remove it and **invoke callback with « `now` »**. *(the same single `now` for every callback)*

**Root cause**

[`packages/happy-dom/src/window/BrowserWindow.ts` → `requestAnimationFrame()`](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/window/BrowserWindow.ts#L2603-L2656) schedules one `setImmediate` *per call* and reads `this.performance.now()` independently inside each:

```ts
const id = TIMER.setImmediate(() => {
    this.#browserFrame[PropertySymbol.asyncTaskManager].endImmediate(id);
    // …
    callback(this.performance.now()); // per-call task, per-call timestamp
});
```

There is no shared per-frame queue, so there's no single timestamp and no atomic batch. (Relatedly, it returns the `NodeJS.Immediate` object rather than an integer handle.)

**Suggested fix**

Collect callbacks into a per-frame map keyed by an integer handle, and flush them in **one** `setImmediate` with a single timestamp, snapshotting the handles so callbacks added mid-frame defer to the next frame (the existing `closed` / `preventTimerLoops` guards and error capturing stay):

```ts
#animationFrameCallbacks = new Map<number, (timestamp: number) => void>();
#animationFrameId = 0;
#animationFrameImmediate: NodeJS.Immediate | null = null;

public requestAnimationFrame(callback: (timestamp: number) => void): number {
    if (this.closed) {
        return 0;
    }

    // … existing settings.timer.preventTimerLoops detection stays here …

    const handle = ++this.#animationFrameId;
    this.#animationFrameCallbacks.set(handle, callback);

    if (this.#animationFrameImmediate === null) {
        const useTryCatch =
            !settings ||
            (!settings.disableErrorCapturing &&
                settings.errorCapture === BrowserErrorCaptureEnum.tryAndCatch);

        const id = TIMER.setImmediate(() => {
            this.#browserFrame[PropertySymbol.asyncTaskManager].endImmediate(id);
            this.#animationFrameImmediate = null;

            const timestamp = this.performance.now(); // single shared timestamp
            // Snapshot handles registered before this frame; ones added during
            // the run keep new handles and run on the next frame.
            for (const currentHandle of [...this.#animationFrameCallbacks.keys()]) {
                const frameCallback = this.#animationFrameCallbacks.get(currentHandle);
                if (!frameCallback) {
                    continue; // cancelled during this frame
                }
                this.#animationFrameCallbacks.delete(currentHandle);
                if (useTryCatch) {
                    let result: any;
                    try {
                        result = frameCallback(timestamp);
                    } catch (error) {
                        this[PropertySymbol.dispatchError](<Error>error);
                    }
                    if (result instanceof Promise) {
                        result.catch((error: Error) => this[PropertySymbol.dispatchError](error));
                    }
                } else {
                    frameCallback(timestamp);
                }
            }
        });
        this.#animationFrameImmediate = id;
        this.#browserFrame[PropertySymbol.asyncTaskManager].startImmediate(id);
    }

    return handle;
}

public cancelAnimationFrame(handle: number): void {
    this.#animationFrameCallbacks.delete(handle);
}
```

⚠️ **Behavior change to flag:** this returns an **integer** handle (matching the spec, jsdom, and browsers) instead of a `NodeJS.Immediate`. The existing test *"Requests an animation frame."* (`test/window/BrowserWindow.test.ts`) asserts `timeoutId.constructor.name === 'Immediate'` and would need updating. There's no way to keep the `Immediate` return type while supporting per-callback `cancelAnimationFrame` in a batched model, and the integer handle is the spec-correct behavior anyway.

**Suggested test**

```ts
it('Runs all callbacks registered for the same frame with a single shared timestamp.', async () => {
    const timestamps: number[] = [];
    await new Promise((resolve) => {
        window.requestAnimationFrame((t) => timestamps.push(t));
        window.requestAnimationFrame((t) => {
            timestamps.push(t);
            resolve(null);
        });
    });
    expect(timestamps).toHaveLength(2);
    expect(timestamps[0]).toBe(timestamps[1]);
});

it('Defers a callback registered during the frame to the next frame.', async () => {
    const order: string[] = [];
    await new Promise((resolve) => {
        window.requestAnimationFrame(() => {
            order.push('a');
            window.requestAnimationFrame(() => {
                order.push('c'); // next frame
                resolve(null);
            });
        });
        window.requestAnimationFrame(() => order.push('b'));
    });
    expect(order).toEqual(['a', 'b', 'c']);
});
```

**Workaround**

Replace `requestAnimationFrame`/`cancelAnimationFrame` with a batching shim (one timer per frame, shared timestamp). `window.setTimeout` keeps it tracked by happy-dom's task manager:

```js
const pending = new Map();
let nextHandle = 1;
let scheduled = false;

window.requestAnimationFrame = (callback) => {
  const handle = nextHandle++;
  pending.set(handle, callback);
  if (!scheduled) {
    scheduled = true;
    window.setTimeout(() => {
      scheduled = false;
      const timestamp = window.performance.now();
      for (const handle of [...pending.keys()]) {
        const callback = pending.get(handle);
        if (!callback) continue;
        pending.delete(handle);
        callback(timestamp);
      }
    }, 0);
  }
  return handle;
};

window.cancelAnimationFrame = (handle) => {
  pending.delete(handle);
};
```

**Environment**
- happy-dom: 20.10.1 (also 20.9.0; code path unchanged on `main`)
- Node.js: 24
- Reference: Chrome / Firefox / Safari batch with one timestamp; jsdom 29 (`pretendToBeVisual: true`) matches

mixelburg:
Resolves #2186

### Description

requestAnimationFrame was scheduling a separate setImmediate per call, so callbacks registered for the same frame received different timestamps and ran in separate tasks instead of one atomic pass.

This batches callbacks into a per-frame map, flushed in a single setImmediate with one shared timestamp. Callbacks registered during a frame are deferred to the next frame (snapshot before iterating).

The return type changes from NodeJS.Immediate to number, matching the spec (unsigned long integer) and browser/jsdom behavior.

### Changes
- BrowserWindow.ts: Added animationFrameCallbacks map and animationFrameImmediate for batching. requestAnimationFrame returns number. cancelAnimationFrame removes from map.
- BrowserWindow.test.ts: Updated return type assertion. Added tests for shared timestamp and deferred-callback behavior.

### DIFF
--- a/packages/happy-dom/src/window/BrowserWindow.ts
+++ b/packages/happy-dom/src/window/BrowserWindow.ts
@@ -875,6 +875,9 @@ export default class BrowserWindow extends EventTarget implements INodeJSGlobal 
 	#zeroDelayTimeout: { timeouts: Array<Timeout> | null } = { timeouts: null };
 	#timerLoopStacks: string[] = [];
 	#timerLoopLimits: ITimerLoopsLimit[] = [];
+	#animationFrameCallbacks: Map<number, (timestamp: number) => void> = new Map();
+	#animationFrameId = 0;
+	#animationFrameImmediate: NodeJS.Immediate | null = null;
 
 	/**
 	 * Constructor.
@@ -2602,9 +2605,9 @@ export default class BrowserWindow extends EventTarget implements INodeJSGlobal 
 	 * @param callback Callback.
 	 * @returns ID.
 	 */
-	public requestAnimationFrame(callback: (timestamp: number) => void): NodeJS.Immediate {
+	public requestAnimationFrame(callback: (timestamp: number) => void): number {
 		if (this.closed) {
-			return <NodeJS.Immediate>{};
+			return 0;
 		}
 		const settings = this.#browserFrame.page.context.browser.settings;
 
@@ -2621,7 +2624,7 @@ export default class BrowserWindow extends EventTarget implements INodeJSGlobal 
 						? 1
 						: (settings.timer.preventTimerLoops.requestAnimationFrame ?? 1))
 				) {
-					return <NodeJS.Immediate>{};
+					return 0;
 				}
 			} else {
 				timerLoopStacks.push(stack);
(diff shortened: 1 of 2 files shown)


mixelburg:
Closing as this hasn't received any feedback in 35+ days. Happy to reopen if there's interest.

PR #2246, opened 2026-07-18
fix: [#2374] classic `<script>` top-level `var`/`function` declarations must reach the global object

### LINKED ISSUE #2374: classic `<script>` top-level `var` / `function` declarations don't reach `window`
### Describe the bug

A classic (non-module) `<script>` has its body wrapped in an anonymous function before it
runs, so top-level `var` and `function` declarations stay inside that wrapper's scope and
never become properties of the global object. Real browsers — and the HTML / ECMAScript
spec for classic scripts — put those declarations on the global object; only
`type="module"` scripts are supposed to get their own scope.

### To Reproduce

```js

const window = new Window();

const script = document.createElement('script');
script.text = `
  var foo = 123;
  function bar() { return 'bar'; }
`;
document.body.appendChild(script);

console.log(window.foo);    // happy-dom: undefined — real browsers: 123
console.log(window.bar);    // happy-dom: undefined — real browsers: [Function: bar]
```

### Expected behavior

`window.foo === 123` and `window.bar()` returns `'bar'`, same as a real browser.

### Real-world impact

Loading htmx as a classic `<script src>` breaks: the library's top-level `var htmx = ...`
never reaches `window`, even though the exact same file works in an actual browser.

### Environment

- happy-dom: latest
- Node: 20+
- OS: Linux

marciomazza:
### Description

Resolves #2374

happy-dom wrapped every classic script body in `(function anonymous($happy_dom) { ... })` before running it, trapping top-level "var" and function declarations inside that function's own scope instead of letting them become properties of the global object, as real browsers do (and as the ECMAScript spec requires for classic, non-module scripts). Only module scripts are supposed to get their own scope.

Example of the bug:

```js
const script = document.createElement('script');
script.text = 'var foo = 123;';
document.body.appendChild(script);

console.log(window.foo); // happy-dom: undefined — real browsers: 123
```

I ran into this while loading a real-world library (htmx) as a classic `<script src>`: the library's own top-level `var htmx = ...` never reached `window`, even though the exact same file works fine in an actual browser.

In this fix, JavaScriptCompiler no longer wraps the code in a function. `$happy_dom` (needed for `dynamicImport()`/`dispatchError()`) is now bound as a non-enumerable property on the window instead of a function parameter — it stays installed for the window's lifetime because async continuations of `dynamicImport()` reference it after the synchronous top level finishes. `execute()` runs the code directly through `evaluateScript()` so top-level declarations land on the global object. Adjacent behavior change: a syntax error now throws from `execute()` instead of `compile()` (e.g. a top-level `return` → `SyntaxError`), which is more spec-correct; both callers handle it.

### AI

I used **Claude Code** to write this, over multiple iterations. I guided and reviewed
it myself at every step.

### Tests

- [x] Make sure to add tests for your changes. Run your test in a real browser to make sure that the test tests what a real browser would do (e.g. by running the code in the browser console). _(Not verified in an actual browser console — behavior confirmed against the WHATWG/ECMAScript spec for classic-script top-level declarations, which is well established browser behavior.)_
- [x] Run the tests with `npm test` locally to make sure that all tests pass before submitting the PR. _(Full suite passes locally. Also ran `tsc --noEmit` and `eslint --max-warnings 0` clean.)_

### Title

### DIFF
--- a/packages/@happy-dom/server-renderer/test/ServerRendererBrowser.test.ts
+++ b/packages/@happy-dom/server-renderer/test/ServerRendererBrowser.test.ts
@@ -419,14 +419,14 @@ Timer #1
 					headers: { key1: 'value' },
 					outputFile: null,
 					pageConsole: `Error: Error
-    at https://example.com/gb/en/:1:59
+    at https://example.com/gb/en/:1:26
     at Timeout._onTimeout (/window/BrowserWindow.ts:0:0)
     at listOnTimeout (node:internal/timers:0:0)
     at processTimers (node:internal/timers:0:0)
 `,
 					pageErrors: [
 						`Error: Error
-    at https://example.com/gb/en/:1:59
+    at https://example.com/gb/en/:1:26
     at Timeout._onTimeout (/window/BrowserWindow.ts:0:0)
     at listOnTimeout (node:internal/timers:0:0)
     at processTimers (node:internal/timers:0:0)`
--- a/packages/happy-dom/src/javascript/JavaScriptCompiler.ts
+++ b/packages/happy-dom/src/javascript/JavaScriptCompiler.ts
@@ -81,17 +81,18 @@ export default class JavaScriptCompiler {
 
 		const regExp = new RegExp(STATEMENT_REGEXP);
 		const count = this.count;
-		let newCode = '(function anonymous($happy_dom) {';
+		let newCode = '';
 		let match: RegExpExecArray | null = null;
 		let precedingToken: string;
 		let textBetweenStatements: string;
 		let lastIndex = 0;
 		const importStartIndex = -1;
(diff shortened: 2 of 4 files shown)


marciomazza:
This PR's CI failure isn't caused by this diff — it's a known flaky test on `master`: the same cache-revalidation flake as `SyncFetch.test.ts` (stale content returned instead of the revalidated value), here in `Fetch.test.ts`, unrelated to what's changed here. Fixed in #2266.

Note there's a second unrelated flake still on `master` too — a `ServerRendererBrowser.test.ts` debug-trace race, fixed in #2270 — so a rebuild/rerun here could hit that one instead before both land. No action needed on this PR's own diff either way.

PR #2248, opened 2026-07-19
fix: [#2247] Improved contenteditable focus and selection behavior

### LINKED ISSUE #2247: Contenteditable elements don't get a selection when focused
**Describe the bug**
In browsers, when a `contenteditable="true"` elements gets focused, either by clicking or by keyboard navigation, the cursor appears. This happens because a `Selection` is associated with the element.

Happy DOM doesn't currently do this.

I have a fix submitted here: https://github.com/capricorn86/happy-dom/pull/2248

**To Reproduce**
1. Create a `<div contenteditable="true">hello, world</div>`
2. Click the div

**Expected behavior**
The div should be focused, and should have a collapsed `Selection` (in browsers this will appear where the user clicked, but since Happy DOM doesn't do layout, it is undefined where that caret should be).

In addition, `event.getTargetRanges()` should return the `StaticRange` objects associated with the `InputEvent`.

**Device and details:**

- OS: MacOS
- Node version: 22.22.2
- Package version: 20.10.6

**Additional context**
This is part of some work I'm doing to try to be able to write headless tests for Slate.

I have a related PR drafted against user-event https://github.com/erikpukinskis/user-event/pull/1 that fixes an issue with nested `contenteditable="false"` elements, but for the simple case, only this happy-dom PR is needed for caret placement, typing, etc to work.

erikpukinskis:
### Description

This branch adds some additional selection behavior to `contenteditable` elements. I needed this stuff to be able to test the DOM behavior in a [Slate](https://www.slatejs.org/)-based document editor I am writing.

Builds on the basic `contenteditable` attribute behavior added in #1757 and previous Selection work (#1953, #1954).

Resolves #2247

### Changes

1. **Caret placement on `focus()`** — when a `contenteditable` element gains focus, HappyDOM now collapses the selection to the start of the first editable text node, matching the behavior required by the Selection API spec.
2. **`StaticRange`** — a new class implementing the [StaticRange interface](https://developer.mozilla.org/en-US/docs/Web/API/StaticRange), which is a prerequisite for `InputEvent.getTargetRanges()`.
3. **`InputEvent.getTargetRanges()`** — adds the missing method and the corresponding `targetRanges` init option, allowing test code and tools like `@testing-library/user-event` to construct `beforeinput` events with proper target range data.

### Tests

### Title

#### Why I'm testing in Happy DOM and not a browser

It makes sense that Slate itself would want to use real browsers for its regression tests. It is a low level cross-browser library that deals with browser quirks.

But for actual editors based on top of Slate, there is still a lot of tricky DOM behaviors that aren't browser-specific: Toolbars are built when you make a selection. Keyboard events affect `Range`/`Selection` and DOM contents. These behaviors are not primarily visual, so testing DOM state makes sense.

#### Why this behavior belongs in Happy DOM

Caret placement on focus is explicitly required by the [Selection API spec](https://w3c.github.io/selection-api/): 

> When an editing host gains focus, if the selection has no range inside it, the user agent must place the caret at the beginning of the editing host.

This is a side effect of the `focus()` DOM method itself — the same category as firing `focusin` or updating `document.activeElement`, both of which Happy DOM already handles correctly in `HTMLElementUtility.focus()`.

`StaticRange` and `getTargetRanges()` are pure spec additions with no behavioral side effects.

### DIFF
--- a/packages/happy-dom/src/event/events/IInputEventInit.ts
+++ b/packages/happy-dom/src/event/events/IInputEventInit.ts
@@ -1,9 +1,11 @@
 import type DataTransfer from '../DataTransfer.js';
 import type IUIEventInit from '../IUIEventInit.js';
+import type StaticRange from '../../range/StaticRange.js';
 
 export default interface IInputEventInit extends IUIEventInit {
 	inputType?: string;
 	data?: string;
 	dataTransfer?: DataTransfer;
 	isComposing?: boolean;
+	targetRanges?: StaticRange[];
 }
--- a/packages/happy-dom/src/event/events/InputEvent.ts
+++ b/packages/happy-dom/src/event/events/InputEvent.ts
@@ -1,5 +1,6 @@
 import type DataTransfer from '../DataTransfer.js';
 import UIEvent from '../UIEvent.js';
+import type StaticRange from '../../range/StaticRange.js';
 import type IInputEventInit from './IInputEventInit.js';
 
 /**
@@ -10,6 +11,7 @@ export default class InputEvent extends UIEvent {
 	public readonly dataTransfer: DataTransfer | null;
 	public readonly inputType: string;
 	public readonly isComposing: boolean;
+	readonly #targetRanges: StaticRange[];
 
 	/**
 	 * Constructor.
@@ -24,5 +26,15 @@ export default class InputEvent extends UIEvent {
 		this.dataTransfer = eventInit?.dataTransfer ?? null;
 		this.inputType = eventInit?.inputType ?? '';
 		this.isComposing = eventInit?.isComposing ?? false;
+		this.#targetRanges = eventInit?.targetRanges ?? [];
+	}
+
+	/**
+	 * Returns the ranges that this event will affect.
+	 *
+	 * @returns A copy of the target ranges array.
+	 */
+	public getTargetRanges(): StaticRange[] {
+		return this.#targetRanges.slice();
 	}
 }
--- a/packages/happy-dom/src/index.ts
+++ b/packages/happy-dom/src/index.ts
@@ -181,6 +181,7 @@ import XMLDocument from './nodes/xml-document/XMLDocument.js';
 import PermissionStatus from './permissions/PermissionStatus.js';
 import Permissions from './permissions/Permissions.js';
 import Range from './range/Range.js';
+import StaticRange from './range/StaticRange.js';
 import ResizeObserver from './resize-observer/ResizeObserver.js';
 import Screen from './screen/Screen.js';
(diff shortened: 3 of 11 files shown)


erikpukinskis:
Additional questions I considered, for reference...

**Why change selection on `mousedown` if `focus` already does it?**

In the common case, this `mousedown` handler is redundant. Click on an unfocused element → `focus` is fired → selection changes. But it's still needed for a few rarer cases:

1. A `preventDefault` on `mousedown` could stop the `focus` from getting fired. That's not something a test/user simulation can easily map to `focus` behavior.

2. Clicks on children propagate up to the nearest `contenteditable` parent. Without the `mousedown` handler there's no easy way for tests/user simulation to determine which element should receive the `focus`.

3. Even if a `contenteditable` is focused, that doesn't guarantee there is a `Selection`. The spec says users must be able to move the carat within an editing house. And this _may_ be triggered as the default action of `mousedown`.

   So, to me it makes sense that Happy DOM does _something_ with the mousedown, even if actual browsers might do something _fancier_ (like hitbox testing to find the nearest text).


**Why is this behavior on `mousedown` if other elements handle things like focus on `click`?**

According to the spec, buttons, inputs, etc get focus behavior on click because for those elements click is considered an §6.5 "[activation behavior](https://html.spec.whatwg.org/multipage/interaction.html#activation)"

Contenteditable editing isn’t “activation behavior” in §6.5 terms. It lives under [§6.8 Editing](https://html.spec.whatwg.org/multipage/interaction.html#editing): moving the caret and changing selection inside an editing host are default actions of mousedown (and keydown), not click.

**Would this behavior fit better in `testing-library`?**

Determining that a "click" from a user would fire a `mousedown` on a certain element, and then a `click`, or a touch or whatever... that logic would make sense in `testing-library`. However the HTML spec specifies that selection changes should happen when `mousedown` is fired, which feels like a Happy DOM concern.

You might argue that determining _where_ the selection should go is "user interaction simulation". But even a keyboard user who just tabs into a contenteditable will expect to see a caret placed.

**If we're adding focus-on-click, selection-on-focus, etc, will this lead to handling arrow keys? Or typing text into a `contenteditable`?**

 - **Up/down arrow keys** — Almost certainly not. The behavior there depends on layout.
 - **Left/right arrow keys** — This _might_ make sense in Happy DOM. The HTML spec does require it. But it's not necessary for Slate-based editors—Slate fires preventDefault and handles these keys itself.
 - **Text** — Almost certainly not. You'd need to handle deleting ranges, splitting/merging nodes, handling plain/rich text, etc... basically, write an entire editor. It makes sense that Happy DOM would stub out just enough that an editor could function similarly to how it does in the browser, but not implement the editing itself.

Issue #2170, opened 2026-06-02
Regression (20.3.3): Node.contains() returns false for a genuine descendant of a proxied <form>/<select> when the subtree is assembled detached then attached


mrsvan:
**Describe the bug**

Since **v20.3.3**, `Node.contains()` returns `false` for an element that is a genuine descendant, when the subtree is assembled **while detached** and only afterwards attached to the document. The same code returns the correct `true` in v20.3.2 and earlier, and is still broken in the latest **v20.9.0**.

`form` → `div` → `input` is a real ancestry chain, so `form.contains(input)` must be `true`. The "build detached, then attach" order is exactly what React / React Testing Library does, so in practice this breaks `@testing-library/jest-dom`'s `toContainElement` (and any `contains()`-based assertion) for React-rendered DOM. It reproduces specifically with `<form>` and `<select>` ancestors (the elements that return a Proxy from their constructor).

**To Reproduce**

```js
import { Window } from 'happy-dom';

const document = new Window().document;

const input = document.createElement('input');
const div = document.createElement('div');
const form = document.createElement('form');

div.appendChild(input);          // input -> div
form.appendChild(div);           // div   -> form   (subtree assembled while DETACHED)
document.body.appendChild(form); // attach the assembled subtree to the document

console.log(form.contains(input));
// v20.3.2 and earlier: true  (correct — input IS a descendant of form)
// v20.3.3 … v20.9.0:   false (BUG)
```

- **Triggers:** assembling the nested subtree **before** attaching it to the document.
- **Does NOT trigger:** attaching the `form` first and then appending children top-down; or a single-level `form.appendChild(input)`.

**Expected behavior**

`form.contains(input)` returns `true` — `input` is unambiguously a descendant of `form`.

**Environment**

- happy-dom: 20.3.3 – 20.9.0 (latest). Last good: 20.3.2.
- Node.js: 24.16.0
- Originally surfaced with @testing-library/react 16.x + React 19 + Vitest 4, where `expect(form).toContainElement(input)` fails for a `<form><div><input/></div></form>` render.

**Additional context — bisect & root cause**

Patch-level bisect of the repro:

| version | `form.contains(input)` |
|---|---|
| 20.3.0 – 20.3.2 | `true` (correct) |
| **20.3.3** | `false` (regression enters here) |
| 20.3.4 … 20.9.0 | `false` |

[...]

PR #2254, opened 2026-07-25
fix: [#2253] Use proxy as parent node in Node.connectedToNode()

### LINKED ISSUE #2253: Inserting a <form>/<select> that already has children re-points the children's parentNode at the raw target instead of the Proxy
**Describe the bug**

`HTMLFormElement` and `HTMLSelectElement` return a `Proxy` from their constructor, and that proxy is the object every public API hands out — `createElement()`, `querySelector()`, `.children`, framework refs.

`Node[PropertySymbol.appendChild]()` and `Node[PropertySymbol.insertBefore]()` correctly store that proxy as the child's parent, using `const self = this[PropertySymbol.proxy] || this`. But `Node[PropertySymbol.connectedToNode]()`, which runs on the node being inserted, re-points every direct child's parent at bare `this` — the **raw target**, because symbol-keyed methods are bound to the target by `ClassMethodBinder`:

https://github.com/capricorn86/happy-dom/blob/084afd6/packages/happy-dom/src/nodes/node/Node.ts#L987-L991

```ts
const childNodes = this[PropertySymbol.nodeArray].slice();
for (let i = 0, max = childNodes.length; i < max; i++) {
    childNodes[i][PropertySymbol.parentNode] = this; // <- raw target, not this[PropertySymbol.proxy]
    childNodes[i][PropertySymbol.connectedToNode]();
}
```

So as soon as a `<form>`/`<select>` that **already has children** is inserted into any parent, its children's parent pointer holds an object that is `!==` the one user code holds. Everything that reads the parent chain then returns, or compares against, the wrong object:

| assertion | happy-dom 20.11.1 | jsdom 29.1.1 | Chromium 148 / Firefox 150 / WebKit 26.4 |
| --- | --- | --- | --- |
| `div.parentNode === form` | ❌ `false` | ✅ | ✅ |
| `div.parentElement === form` | ❌ `false` | ✅ | ✅ |
| `input.closest('form') === form` | ❌ `false` | ✅ | ✅ |
| `form.contains(input)` | ❌ `false` | ✅ | ✅ |
| `form.compareDocumentPosition(input)` | ❌ `2` (`PRECEDING`) | ✅ `20` | ✅ `20` |
| `event.currentTarget === form` (bubbled) | ❌ `false` | ✅ | ✅ |
| `option.parentNode === select` | ❌ `false` | ✅ | ✅ |
| `form.children[0] === div` | ✅ | ✅ | ✅ |
| `document.querySelector('form') === form` | ✅ | ✅ | ✅ |

The last two rows are why this is so easy to hit: `querySelector()` and `.children` correctly hand user code the proxy, so every identity comparison against the parent chain silently fails.

**To Reproduce**

```js
import * as PropertySymbol from 'happy-dom/lib/PropertySymbol.js';

const window = new Window();

form.appendChild(div);
div.appendChild(input);
document.body.appendChild(form); // <- form already has children when inserted

console.log(div.parentNode === form);             // false (expected true)
console.log(div.parentElement === form);          // false (expected true)
console.log(input.closest('form') === form);      // false (expected true)
console.log(form.contains(input));                // false (expected true)
console.log(form.compareDocumentPosition(input)); // 2     (expected 20)

// The parent pointer holds the raw target rather than the proxy:
console.log(div.parentNode[PropertySymbol.proxy] === form); // true

// Event propagation walks the same chain, so a listener on the form
// sees the raw target as currentTarget:
form.addEventListener('test', (event) => {
    console.log(event.currentTarget === form); // false (expected true)
});
input.dispatchEvent(new window.Event('test', { bubbles: true }));
```

Swapping the `<form>` for a `<div>` makes every identity check log `true` and `compareDocumentPosition` return `20` — only the two proxied elements are affected.

A plain React 19 render reproduces it without any manual DOM calls, because React builds the subtree detached and then inserts it:

```jsx
root.render(<form id="form"><div><input /></div></form>);

const form = container.querySelector('#form');
const input = container.querySelector('input');
input.closest('form') === form; // false
form.contains(input);           // false
```

**Expected behavior**

There is exactly one object per node. `div.parentNode`, `div.parentElement`, `input.closest('form')` and the event path must all yield the same object that `createElement()` / `querySelector()` returned, so `form.contains(input)` is `true` and `form.compareDocumentPosition(input)` is `20` (`DOCUMENT_POSITION_CONTAINED_BY | DOCUMENT_POSITION_FOLLOWING`).

**Device and details:**

- OS: macOS 26.5.2 (Darwin 25.5.0)
- Node version: 24.18.0
- Package version: happy-dom 20.11.1 (latest)

**Additional context**

**When it triggers.** Any insertion of a `<form>`/`<select>` that already has children — connection to the document is not required:

| scenario | result |
| --- | --- |
| `appendChild` the form into a **detached** `<div>` | ❌ broken |
| `appendChild` / `insertBefore` / `replaceChild` the form into `document.body` | ❌ broken |
| form never inserted anywhere | ✅ ok |
| children appended **after** the form is already inserted | ✅ ok |
| subtree built by `innerHTML` on a connected host | ✅ ok |

That last pair is why this mostly bites framework renders: React/Vue/Svelte build the subtree first and insert it afterwards, whereas the HTML parser appends into an already-inserted parent.

**Suggested fix.** Use the same `self` normalization the sibling insertion methods already use, in `Node[PropertySymbol.connectedToNode]()`:

```diff
+		// HTMLFormElement and HTMLSelectElement return a Proxy from their constructor.
+		const self = (<any>this)[PropertySymbol.proxy] || this;
+
 		const childNodes = this[PropertySymbol.nodeArray].slice();
 		for (let i = 0, max = childNodes.length; i < max; i++) {
-			childNodes[i][PropertySymbol.parentNode] = this;
+			childNodes[i][PropertySymbol.parentNode] = self;
 			childNodes[i][PropertySymbol.connectedToNode]();
 		}
```

With this change every row of both tables above turns ✅, and the existing suite is unaffected — **7628 tests / 297 files pass** on `master` (084afd6) with the patch applied.

**Relationship to #2170.** This looks like the root cause of #2170 as well. That issue reports `contains()` returning `false` for a descendant of a proxied `<form>`/`<select>`; the linked PR #2176 fixes it by normalizing both sides of the identity comparison inside `NodeUtility.isInclusiveAncestor`. That addresses `contains()`, but the parent pointer itself still holds the raw target, so `parentElement`, `closest()`, `compareDocumentPosition()` and `event.currentTarget` stay wrong.

Fixing `connectedToNode()` instead repairs `contains()` at the source — I verified the #2170 repro passes with only the diff above and no change to `NodeUtility` — while also fixing the other five. It also keeps the #1876 behaviour that #1991 introduced, since the proxy is what ends up in the tree.

**Suggested test.** Fails on 3 of 4 cases before the fix, passes after (the negative case guards against false positives):

```ts
it('keeps the proxy identity of a <form> ancestor when the subtree is inserted', () => {

	form.appendChild(div);
	div.appendChild(input);

	expect(div.parentNode).toBe(form);
	expect(div.parentElement).toBe(form);
	expect(input.closest('form')).toBe(form);
	expect(form.contains(input)).toBe(true);
	expect(form.compareDocumentPosition(input)).toBe(20);
});

it('keeps the proxy identity of a <select> ancestor when the subtree is inserted', () => {
	const select = document.createElement('select');
	const option = document.createElement('option');

	select.appendChild(option);
	document.body.appendChild(select);

	expect(option.parentNode).toBe(select);
	expect(select.contains(option)).toBe(true);
});

it('reports the proxy as currentTarget for an event bubbling to a <form>', () => {

	form.appendChild(input);

	let currentTarget: EventTarget | null = null;
	form.addEventListener('test', (event) => {
		currentTarget = event.currentTarget;
	});
	input.dispatchEvent(new window.Event('test', { bubbles: true }));

	expect(currentTarget).toBe(form);
});

it('does not report unrelated nodes as ancestors', () => {
	const sibling = document.createElement('input');

	form.appendChild(document.createElement('div'));
	document.body.appendChild(sibling);

	expect(form.contains(sibling)).toBe(false);
	expect(sibling.closest('form')).toBe(null);
	expect(sibling.parentElement).toBe(document.body);
});
```

Happy to open a PR with the fix and these tests if that's useful.

diegohaz:
### Description

`HTMLFormElement` and `HTMLSelectElement` return a `Proxy` from their constructor, and that proxy is the identity the public API hands out — `createElement()`, `querySelector()`, `.children`, framework refs.

`Node[PropertySymbol.appendChild]()` and `insertBefore()` already store that proxy as the child's parent via `const self = this[PropertySymbol.proxy] || this`. But `Node[PropertySymbol.connectedToNode]()` re-points every direct child at bare `this`, which is the **raw target** — `ClassMethodBinder` binds symbol-keyed methods to the target, so `this` is the underlying node even when the method is reached through the proxy.

So whenever a `<form>`/`<select>` **that already has children** is inserted into any parent, its children's parent pointer holds an object that is `!==` the one user code holds, and everything reading the parent chain returns or compares against the wrong object:

```js

form.appendChild(div);
div.appendChild(input);
document.body.appendChild(form); // <- form already has children when inserted

div.parentNode === form;             // false (expected true)
div.parentElement === form;          // false (expected true)
input.closest('form') === form;      // false (expected true)
form.contains(input);                // false (expected true)
form.compareDocumentPosition(input); // 2     (expected 20)
```

Event propagation walks the same chain, so a listener on the `<form>` receives the raw target as `currentTarget`. The same applies to `<select>`/`<option>`.

Connection to the document is not required — `insertBefore()` and `replaceChild()` into a detached parent hit it too. Appending children *after* the element is already in place is fine, which is why `innerHTML`-parsed markup works and framework renders do not: React/Vue/Svelte build the subtree first and insert it afterwards, so a plain React render reproduces all of it with no manual DOM calls.

Resolves #2253

### Fix

Normalize to the proxy in `connectedToNode()`, the same way `appendChild()` and `insertBefore()` already do. Non-proxied nodes are unchanged, since `this[PropertySymbol.proxy]` is undefined for them.

This also appears to be the root cause of #2170 — `contains()` returning `false` for a descendant of a proxied `<form>`/`<select>`. #2176 fixes that by normalizing both sides of the comparison inside `NodeUtility.isInclusiveAncestor`, which repairs `contains()` but leaves the parent pointer holding the raw target, so `parentElement`, `closest()`, `compareDocumentPosition()` and `event.currentTarget` stay wrong. Fixing the parent pointer repairs `contains()` at the source: the #2170 reproduction passes with this diff alone and no change to `NodeUtility`.

### Verification

Behaviour was checked against real browsers before writing the tests — Chromium 148, Firefox 150 and WebKit 26.4 (via Playwright), plus jsdom 29.1.1 as a second reference. All of them return the expected values for every assertion above, including `compareDocumentPosition()` in both directions (`20` parent→child, `10` child→parent). happy-dom was the only outlier.

- `npm test` in `packages/happy-dom`: **7638 passed / 297 files**, no failures.
- The 10 added tests: 9 fail on `master` without the source change and pass with it. The 10th is a negative case (a sibling is not reported as a descendant) that passes either way, guarding the normalization against false positives.
- `npm run lint`: clean.

The `@happy-dom/node-canvas-adapter` suite has 9 pre-existing failures on this machine (native-canvas PNG snapshot mismatches). I confirmed they fail identically on unmodified `master`, so they are unrelated to this change.

### Tests

Added to `Node.test.ts` and `Element.test.ts`, covering the public API only:

- `get parentElement()` — `<form>` and `<select>` inserted with children already appended, plus insertion into a parent that is not connected to the document
- `contains()` — nested elements in an inserted `<form>`/`<select>`, plus the negative sibling case
- `compareDocumentPosition()` — `20` parent→child and `10` child→parent
- `dispatchEvent()` — `target`/`currentTarget` for an event bubbling to a `<form>`
- `closest()` — `<form>` and `<select>` ancestors

### Tests

### Title

### AI tools

Disclosure: I used Claude Code as an assistant on this PR — for the root-cause investigation, the browser and jsdom cross-checks, and drafting the patch and tests. Every verification step described above was run locally against this branch. Happy to discuss or adjust any part of it.

### DIFF
--- a/packages/happy-dom/src/nodes/node/Node.ts
+++ b/packages/happy-dom/src/nodes/node/Node.ts
@@ -987,7 +987,7 @@ export default class Node extends EventTarget {
 		// HTMLFormElement and HTMLSelectElement return a Proxy from their constructor.
 		// The proxy is the identity handed out by the public API, so it has to be stored
 		// as the parent of the child nodes, the same way as in appendChild() and insertBefore().
-		const self = (<any>this)[PropertySymbol.proxy] || this;
+		const self = this[PropertySymbol.proxy] || this;
 
 		const childNodes = this[PropertySymbol.nodeArray].slice();
 		for (let i = 0, max = childNodes.length; i < max; i++) {
--- a/packages/happy-dom/src/nodes/node/Node.ts
+++ b/packages/happy-dom/src/nodes/node/Node.ts
@@ -984,9 +984,14 @@ export default class Node extends EventTarget {
 			}
 		}
 
+		// HTMLFormElement and HTMLSelectElement return a Proxy from their constructor.
+		// The proxy is the identity handed out by the public API, so it has to be stored
+		// as the parent of the child nodes, the same way as in appendChild() and insertBefore().
+		const self = (<any>this)[PropertySymbol.proxy] || this;
+
 		const childNodes = this[PropertySymbol.nodeArray].slice();
 		for (let i = 0, max = childNodes.length; i < max; i++) {
-			childNodes[i][PropertySymbol.parentNode] = this;
+			childNodes[i][PropertySymbol.parentNode] = self;
 			childNodes[i][PropertySymbol.connectedToNode]();
 		}
 
--- a/packages/happy-dom/test/nodes/element/Element.test.ts
+++ b/packages/happy-dom/test/nodes/element/Element.test.ts
@@ -1081,6 +1081,28 @@ describe('Element', () => {
 			expect(b.closest('div .span b') === b).toBe(true);
 			expect(b.closest('div .span article b') === b).toBe(true);
 		});
+
+		it('Finds the closest HTMLFormElement when it is inserted after its child nodes have been appended (issue #2253).', () => {
+			const form = document.createElement('form');
+			const div = document.createElement('div');
+			const input = document.createElement('input');
+
+			form.appendChild(div);
+			div.appendChild(input);
+			document.body.appendChild(form);
+
+			expect(input.closest('form') === form).toBe(true);
+		});
+
+		it('Finds the closest HTMLSelectElement when it is inserted after its child nodes have been appended (issue #2253).', () => {
+			const select = document.createElement('select');
+			const option = document.createElement('option');
+
+			select.appendChild(option);
+			document.body.appendChild(select);
+
+			expect(option.closest('select') === select).toBe(true);
+		});
 	});
 
 	describe('querySelectorAll()', () => {
--- a/packages/happy-dom/test/nodes/node/Node.test.ts
+++ b/packages/happy-dom/test/nodes/node/Node.test.ts
@@ -255,6 +255,40 @@ describe('Node', () => {
 			expect(htmlElement.parentNode).toBe(document);
 			expect(htmlElement.parentElement).toBe(null);
 		});
+
+		it('Returns HTMLFormElement when it is inserted with child nodes already appended to it (issue #2253).', () => {
+			const form = document.createElement('form');
+			const div = document.createElement('div');
+
+			form.appendChild(div);
+			document.body.appendChild(form);
+
+			expect(div.parentNode).toBe(form);
+			expect(div.parentElement).toBe(form);
+		});
+
+		it('Returns HTMLSelectElement when it is inserted with child nodes already appended to it (issue #2253).', () => {
+			const select = document.createElement('select');
+			const option = document.createElement('option');
+
+			select.appendChild(option);
+			document.body.appendChild(select);
+
+			expect(option.parentNode).toBe(select);
+			expect(option.parentElement).toBe(select);
+		});
+
+		it('Returns HTMLFormElement when it is inserted into a node that is not connected to the document (issue #2253).', () => {
+			const parent = document.createElement('div');
+			const form = document.createElement('form');
+			const div = document.createElement('div');
+
+			form.appendChild(div);
+			parent.appendChild(form);
+
+			expect(div.parentNode).toBe(form);
+			expect(div.parentElement).toBe(form);
+		});
 	});
 
 	describe('get baseURI()', () => {
@@ -387,6 +421,44 @@ describe('Node', () => {
 			expect(form.contains(input)).toBe(true);
 			expect(form.contains(div)).toBe(true);
 			expect(div.contains(input)).toBe(true);
+		});
+
+		it('Returns "true" for nested elements within an HTMLFormElement that is inserted after its child nodes have been appended (issue #2253).', () => {
+			const form = document.createElement('form');
+			const div = document.createElement('div');
+			const input = document.createElement('input');
+
+			form.appendChild(div);
+			div.appendChild(input);
+			document.body.appendChild(form);
+
+			expect(form.contains(input)).toBe(true);
+			expect(form.contains(div)).toBe(true);
+		});
+
+		it('Returns "true" for nested elements within an HTMLSelectElement that is inserted after its child nodes have been appended (issue #2253).', () => {
+			const select = document.createElement('select');
+			const optgroup = document.createElement('optgroup');
+			const option = document.createElement('option');
+
+			select.appendChild(optgroup);
+			optgroup.appendChild(option);
+			document.body.appendChild(select);
+
+			expect(select.contains(option)).toBe(true);
+			expect(select.contains(optgroup)).toBe(true);
+		});
+
+		it('Returns "false" for a node that is not a descendant of an inserted HTMLFormElement (issue #2253).', () => {
+			const form = document.createElement('form');
+			const div = document.createElement('div');
+			const sibling = document.createElement('input');
+
+			form.appendChild(div);
+			document.body.appendChild(form);
+			document.body.appendChild(sibling);
+
+			expect(form.contains(sibling)).toBe(false);
 		});
 	});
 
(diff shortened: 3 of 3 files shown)


diegohaz:
The CI failure here looks unrelated to this change — it's `@happy-dom/server-renderer` › `ServerRendererBrowser.test.ts`, which asserts on the exact async-task trace recorded when a 100 ms render timeout fires against a recursive `setTimeout(t, 10)`. The diff is an extra pending `Task #1` (the page navigation) that hadn't settled yet, which is what you'd expect from a loaded runner.

I re-ran the identical commit through the same `pull_request.yml` workflow on my fork, and it passes on all three Node versions:

| job | upstream (this PR) | fork, same commit |
| --- | --- | --- |
| build (20) | ❌ | ✅ 2m49s |
| build (22) | ❌ | ✅ 2m57s |
| build (24) | ❌ | ✅ 2m49s |

https://github.com/diegohaz/happy-dom/actions/runs/30150519553

Two other reasons I don't think this change can be causal:

1. The fix is a strict no-op for anything that isn't a proxied `<form>`/`<select>` — `this[PropertySymbol.proxy]` is `undefined` for every other node, so `self` is just `this` and the assignment is byte-for-byte what it was. `ServerRendererBrowser.test.ts` renders `<script>function t(){ setTimeout(t, 10) } t()</script>` with no form or select anywhere.
2. `@happy-dom/server-renderer` passes locally on this branch — I ran it 5 times, 83/83 each time.

Timing-sensitive tests seem to misfire on this runner fairly regularly (the docs-only #2252 branch, for instance, went red on a `Fetch.test.ts` cache-revalidation case and passed on re-run). A re-run should be enough here, but let me know if you'd like me to dig into it further.

marciomazza:
I'm having the same issue and from what I look this PR will solve it.

marciomazza:
This PR's CI failure isn't caused by this diff — it's a known flaky test on `master`: `ServerRendererBrowser.test.ts`'s debug-trace race (`"N tasks did not end in time"` mismatch), unrelated to what's changed here. Fixed in #2270.

Note there's a second unrelated flake still on `master` too — a `SyncFetch.test.ts`/`Fetch.test.ts` cache-revalidation issue, fixed in #2266 — so a rebuild/rerun here could hit that one instead before both land. No action needed on this PR's own diff either way.

diegohaz:
CI is passing now.

capricorn86 (approved):
Thank you for your contribution

Issue #2257, opened 2026-07-27
querySelector stops throwing for an invalid selector after the first call (regression in 20.4.0)


janosh:
**Describe the bug**

An invalid selector makes `querySelector` throw a `DOMException` the first time. Every later call with the same string returns `null`, as if the failed parse were cached as a successful empty result. `closest()` keeps throwing for that same string, so the two APIs disagree after the first call.

The cache is per-`Window`, so in test runners that reuse one `Window` per file, the first test sees the throw and every later one silently gets `null`.

**To Reproduce**

```js
import { Window } from 'happy-dom'

const document = new Window().document

for (const attempt of [1, 2, 3]) {
  try {
    console.log(`attempt ${attempt} -> returned ${document.querySelector(`[`)}`)
  } catch (err) {
    console.log(`attempt ${attempt} -> threw ${err.constructor.name}`)
  }
}
// attempt 1 -> threw DOMException
// attempt 2 -> returned null   <- expected to throw
// attempt 3 -> returned null   <- expected to throw
```

**Expected behavior**

Every call throws. The [DOM spec](https://dom.spec.whatwg.org/#scope-match-a-selectors-string) requires a `SyntaxError` `DOMException` whenever the selector fails to parse, with no allowance for caching the failure. jsdom and browsers throw consistently.

**Device and details:** macOS, Node 24.18.0, happy-dom 20.11.1

**Additional context**

Bisected: 20.3.0 throws on every call, 20.4.0 does not, and 20.11.1 still does not. Not caused by the recent #2228, since 20.11.0 predates that merge and already shows it.

The practical failure mode is validating a user-supplied selector by calling `querySelector` inside a `try`/`catch`. The first invalid selector is caught and reported, a second one passes validation, and the error resurfaces later from `closest()` far from its cause.

Kjubikstronk:
Taking this one.

PR #2258, opened 2026-07-28
fix: [#2257] Removes cached null result after querySelector parse failure

mixelburg:
When querySelector caches a pending result before parsing the selector string, an invalid selector that causes the parser to throw leaves behind a cached entry with element: null. Subsequent calls with the same invalid selector return null instead of throwing DOMException.

The fix wraps the parsing and matching logic in a try-catch that removes the cached entry on error, ensuring every call with an invalid selector throws consistently.


mixelburg:

Issue #2256, opened 2026-07-27
var() with a fallback is dropped from inline style declarations


janosh:
**Describe the bug**

An inline declaration whose value contains `var()` with a fallback is discarded, and the CSSOM reports an empty value. The same value works in a stylesheet, so the two paths disagree:

| Surface | `color: var(--nope, red)` |
| :-- | :-- |
| stylesheet + `getComputedStyle` | `"red"` (correct) |
| inline `style.cssText` | `""` (dropped) |

`var()` without a fallback works inline, and plain `calc()` works, so the trigger is the fallback argument, including when nested inside `calc()`.

Inline, this affects `style.cssText`, `style.setProperty()` and `style.width = ...` alike. With `setAttribute('style', ...)` the attribute string survives but the CSSOM still reports `""`, so `getAttribute('style')` and `style.width` disagree.

**To Reproduce**

```js
import { Window } from 'happy-dom'

const document = new Window().document
const show = (css) => {
  const el = document.createElement(`div`)
  el.style.cssText = css
  console.log(`${JSON.stringify(css).padEnd(30)} -> ${JSON.stringify(el.style.cssText)}`)
}

show(`width: calc(100% - 1px)`)      // "width: calc(100% - 1px);"
show(`width: var(--x)`)              // "width: var(--x);"
show(`width: var(--x, 10px)`)        // ""  <- expected "width: var(--x, 10px);"
show(`width: calc(var(--x) - 1px)`)  // ""  <- expected to round-trip
```

**Expected behavior**

All four round-trip. The second argument to `var()` is a valid declaration-value fallback ([CSS Variables](https://www.w3.org/TR/css-variables-1/#using-variables)), and browsers and jsdom preserve it inline.

**Device and details:** macOS, Node 24.18.0, happy-dom 20.11.1

**Additional context**

Reproduces on 15.11.7, 18.0.1, 20.10.6 and 20.11.1, so it is not a recent regression. #1308 fixed fallbacks for the stylesheet path; the inline path looks like it was never covered.

jesusgalhub:
This is affecting us as well. Any plan to merge the current active PR? @mixelburg 
Beforehand: thanks for your time!

PR #2259, opened 2026-07-28
fix: [#2256] Preserves var() fallback and calc() containing var() in inline styles

mixelburg:
Two regex fixes in CSSStyleDeclarationValueParser:

1. CSS_VARIABLE_REGEXP now accepts `var()` with a fallback argument (e.g. `var(--x, 10px)`). Previously only `var(--x)` without fallback was recognized.

2. CALC_REGEXP now handles nested parentheses from `var()` calls inside `calc()` (e.g. `calc(var(--x) - 1px)`).


mixelburg:
Closing in favor of #2280, which uses a more robust balanced-parentheses parser approach for the same issue. The newer PR handles nested parentheses in var() fallback values correctly even in edge cases like `var(--x, rgb(0,0,0))`.

PR #2260, opened 2026-07-29
fix: [#2227] Treat out-of-range index as null in HTMLSelectElement.add()

### LINKED ISSUE #2227: HTMLSelectElement.add() throws when `before` is an out-of-range index instead of appending
**Summary**

When `add(element, before)` is called with a numeric `before` that is greater than the number of options in the select (an index that does not exist), happy-dom throws a `DOMException`. Per the [HTML specification](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#dom-htmloptionscollection-add) and [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/add), an out-of-range index must be treated as `null`, i.e. the element is appended to the end. Browsers and jsdom append; happy-dom throws.

**Reproduction Code**

```js
document.body.innerHTML = '<select><option>A</option></select>'; // 1 option, index 0
const select = document.querySelector('select');

select.add(document.createElement('option'), null); // appends ✅
select.add(document.createElement('option'), 2);    // index 2 does not exist
// Expected: appended (treated as null)
// Actual: throws DOMException "Failed to execute 'add' on 'HTMLFormElement':
//         The node before which the new node is to be inserted before is not a child of this node."
```

**Root Cause**

In `HTMLSelectElement.add()`, for a numeric `before` the implementation does `const optionsElement = options[before]` and throws when it is `undefined`. An index beyond the current option count yields `undefined`, so it throws rather than falling back to append.

**Suggested Solution**

When `before` is a number for which no option exists (out of range), treat it as `null` and append — matching the spec, browsers, and jsdom — instead of throwing.

**Environment**

- happy-dom 20.10.6 (present in master)
- Vitest 3.2.4

mixelburg:
Fixes #2227

When HTMLSelectElement.add(element, before) receives a numeric before index that is out of range (>= options.length), the HTML spec says it should be treated as null (i.e., appended to the end). Browsers and jsdom append; happy-dom was throwing a DOMException.

Added an early check that sets before to undefined when the index is out of range, which falls through to the existing append logic.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2400,9 +2400,6 @@
 				"arm64"
 			],
 			"dev": true,
-			"libc": [
-				"glibc"
-			],
 			"license": "MIT",
 			"optional": true,
 			"os": [
@@ -2420,9 +2417,6 @@
 				"arm64"
 			],
 			"dev": true,
-			"libc": [
-				"musl"
-			],
 			"license": "MIT",
 			"optional": true,
 			"os": [
@@ -2440,9 +2434,6 @@
 				"ppc64"
 			],
 			"dev": true,
-			"libc": [
-				"glibc"
-			],
 			"license": "MIT",
 			"optional": true,
 			"os": [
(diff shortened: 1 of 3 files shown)


mixelburg:

PR #2261, opened 2026-07-29
fix: [#2238] Normalize font-weight keywords to numeric in computed style

### LINKED ISSUE #2238: Elements `font-weight` returned by `getComputedStyle` is not normalized
**Describe the bug**
In a browser `window.getComputedStyle` returns some kind of a normalized style, e.g. if a `div` has `font-weight: bold` on it, `getComputedStyle` will return `font-weight: 700`.

**To Reproduce**
Run the following script:
```typescript
const window = new Window();
const { document } = window;

const div = document.createElement("div");
div.style.fontWeight = "bold";
document.body.appendChild(div);

console.log(window.getComputedStyle(div).fontWeight);
```

It will print `bold`.

**Expected behavior**
Running the script in a browser (without the first three lines) results in `700` being printed to console.

**Device and details:**

- OS: Linux 7.1.3-200.fc44.x86_64
- Node version: v20.20.0
- Package version: 20.10.6

**Additional context**
Similar issues: https://github.com/capricorn86/happy-dom/issues/1829

mixelburg:
Fixes #2238

In browsers, `getComputedStyle(el).fontWeight` returns numeric values for keyword font-weights (e.g. `bold` → `700`, `normal` → `400`). happy-dom was returning the raw keyword value.

Added normalization in `CSSStyleDeclarationComputedStyle` so computed styles resolve font-weight keywords to their numeric equivalents. `element.style.fontWeight` continues to preserve the author's keyword value.

### DIFF
--- a/packages/happy-dom/src/query-selector/QuerySelector.ts
+++ b/packages/happy-dom/src/query-selector/QuerySelector.ts
@@ -264,35 +264,40 @@ export default class QuerySelector {
 			(node[PropertySymbol.ownerDocument] || node)[PropertySymbol.affectsCache].push(cachedItem);
 		}
 
-		let bestMatch: DocumentPositionAndElement | null = null;
-		const matchesMap: Map<string, boolean> = new Map();
-		const scope =
-			node[PropertySymbol.nodeType] === NodeTypeEnum.documentNode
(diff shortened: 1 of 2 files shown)


mixelburg:

PR #2262, opened 2026-07-30
fix: [#2226] Copy labels array to prevent mutation of cached querySelectorAll result

### LINKED ISSUE #2226: HTMLInputElement.labels returns duplicated entries when read more than once
First off, thanks for happy-dom — it's been great to work with, and we're using it to migrate a large test suite. While doing that I ran into a small `.labels` behaviour that I was able to narrow down to a specific cause, so I wanted to write it up in case it's helpful, and I'm happy to open a PR if you'd like.

## Summary

Reading `.labels` on a labelable element (input/select/textarea/…) that has an `id` returns a list that grows by one each time it's accessed — one entry on the first read, two on the second, and so on. The first read is correct; the duplication only appears on subsequent reads.

## Reproduction (no testing-library required)

```js
document.body.innerHTML =
  '<label><input id="r" type="radio" name="g"><span>Alpha</span></label>';
const input = document.querySelector('#r');

console.log([...input.labels].length); // 1  ✅ expected
console.log([...input.labels].length); // 2  unexpected
console.log([...input.labels].length); // 3  unexpected
```

It only reproduces when the control has an `id`; without one, the result is correct on every read.

## What I found

I traced it to `HTMLLabelElementUtility.getAssociatedLabelElements()` ([source](https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElementUtility.ts)). In the `id` branch, the `labels` variable is assigned the internal items array of the `querySelectorAll` result, and the enclosing-`<label>` walk then `push()`es onto that same array:

```ts
labels = <HTMLLabelElement[]>rootNode.querySelectorAll(`label[for="${id}"]`)[PropertySymbol.items];
// …
labels.push(<HTMLLabelElement>parent);
```

Since `querySelectorAll` caches its result, the pushed entry appears to persist on the cached array and accumulates on each subsequent read.

## Possible related reports

I'm not certain, but this may be the same underlying cause behind a couple of existing issues — happy to be corrected if not:

- #1410 (clicking a wrapping `<label>` triggers the input twice) — a duplicated label entry could cause the activation to fire twice.
- #1220 (`getByLabelText` behaving differently from jsdom).

In our case it surfaces through `@testing-library`, whose `getByRole(role, { name })` / `getByLabelText` read `.labels` to compute accessible names, so a duplicated label can make those queries match more than one control (we hit it with MUI `FormControlLabel` radio groups).

## Suggested fix

Copying the array before pushing (rather than mutating the cached one) resolves it in our testing. Something like:

```ts
const labels: HTMLLabelElement[] =
  id && element[PropertySymbol.isConnected]
    ? [...rootNode.querySelectorAll(`label[for="${id}"]`)[PropertySymbol.items]]
    : [];
// the enclosing-<label> walk then pushes onto the fresh copy
```

I'd be glad to put up a PR with this plus a small regression test if that's helpful.

## Environment

- happy-dom: reproduced on `20.10.6` (current `latest`); the relevant code also appears unchanged on `master` and in earlier lines I checked (18.x, 19.x, 20.0.0).
- Test runner: Vitest 3.2.4, `happy-dom` environment.

Thanks again for maintaining this — glad to help however is most useful.

mixelburg:
Fixes #2226. The labels getter assigned the internal items array of a cached querySelectorAll result directly, so the enclosing-label push mutated the cached array and caused duplicates on repeated reads. Spread into a fresh array instead.

### DIFF
--- a/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElementUtility.ts
+++ b/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElementUtility.ts
@@ -24,11 +24,9 @@ export default class HTMLLabelElementUtility {
 			const rootNode =
 				<Document | ShadowRoot>element[PropertySymbol.rootNode] ||
 				element[PropertySymbol.ownerDocument];
-			labels = [
-				...(<HTMLLabelElement[]>(
-					rootNode.querySelectorAll(`label[for="${id}"]`)[PropertySymbol.items]
-				))
-			];
+			labels = <HTMLLabelElement[]>(
+				rootNode.querySelectorAll(`label[for="${id}"]`)[PropertySymbol.items].slice()
+			);
 		} else {
 			labels = [];
 		}
--- a/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElementUtility.ts
+++ b/packages/happy-dom/src/nodes/html-label-element/HTMLLabelElementUtility.ts
@@ -24,9 +24,11 @@ export default class HTMLLabelElementUtility {
 			const rootNode =
 				<Document | ShadowRoot>element[PropertySymbol.rootNode] ||
 				element[PropertySymbol.ownerDocument];
-			labels = <HTMLLabelElement[]>(
-				rootNode.querySelectorAll(`label[for="${id}"]`)[PropertySymbol.items]
-			);
+			labels = [
+				...(<HTMLLabelElement[]>(
+					rootNode.querySelectorAll(`label[for="${id}"]`)[PropertySymbol.items]
+				))
+			];
 		} else {
 			labels = [];
 		}
--- a/packages/happy-dom/test/nodes/html-label-element/HTMLLabelElement.test.ts
+++ b/packages/happy-dom/test/nodes/html-label-element/HTMLLabelElement.test.ts
@@ -209,5 +209,31 @@ describe('HTMLLabelElement', () => {
 			expect(changeFiredCount).toBe(1);
 			expect(div.querySelector('input')?.checked).toBe(true);
 		});
+
+		describe('get labels()', () => {
+			it('Does not duplicate entries on repeated reads.', () => {
+				const input = document.createElement('input');
+				input.id = 'test-input';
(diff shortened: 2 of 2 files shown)


packages/happy-dom/src/nodes/html-label-element/HTMLLabelElementUtility.ts
```diff
@@ -24,9 +24,11 @@ export default class HTMLLabelElementUtility {
...
 				<Document | ShadowRoot>element[PropertySymbol.rootNode] ||
 				element[PropertySymbol.ownerDocument];
-			labels = <HTMLLabelElement[]>(
-				rootNode.querySelectorAll(`label[for="${id}"]`)[PropertySymbol.items]
-			);
+			labels = [
+				...(<HTMLLabelElement[]>(
+					rootNode.querySelectorAll(`label[for="${id}"]`)[PropertySymbol.items]
+				))
+			];
```
capricorn86: Slice is faster.

```suggestion
			labels = <HTMLLabelElement[]>(
				rootNode.querySelectorAll(`label[for="${id}"]`)[PropertySymbol.items].slice()
			);
```

capricorn86 (changes requested):
Thank you for your contribution @mixelburg!

Found one thing to optimize performance.

mixelburg (commented):
Done — replaced the spread with  as suggested. Thanks for the review!

mixelburg (commented):
Rebased on latest master and the `.slice()` optimization from the review is applied. Ready for re-review!

mixelburg (commented):
All review feedback has been addressed: the spread operator was replaced with `.slice()` as suggested, and the PR has been rebased on latest master. All CI checks are passing. Ready for re-review!

mixelburg:
Hi @capricorn86! All review feedback has been addressed \u2014 the spread operator was replaced with `.slice()` as suggested. Ready for re-review whenever you have a moment!

capricorn86 (approved):
Nice!

PR #2263, opened 2026-07-30
feat: [#2213] Expose waitForNavigation() on window.happyDOM

### LINKED ISSUE #2213: Expose `waitForNavigation()` on `window.happyDOM`

mixelburg:
Fixes #2213. Exposes `waitForNavigation()` on `DetachedWindowAPI` (accessible via `window.happyDOM`), delegating to the browser frame's existing implementation. This matches the pattern already used by `waitUntilComplete()` and allows Vitest users to await navigation without polling.

### DIFF
--- a/packages/happy-dom/src/window/DetachedWindowAPI.ts
+++ b/packages/happy-dom/src/window/DetachedWindowAPI.ts
@@ -46,6 +46,15 @@ export default class DetachedWindowAPI {
 	}
 
 	/**
+	 * Waits for the frame to navigate and the response HTML to be written to the document.
+	 *
+	 * @returns Promise.
+	 */
+	public waitForNavigation(): Promise<void> {
+		return this.#browserFrame.waitForNavigation();
+	}
+
+	/**
 	 * Waits for all async tasks to complete.
 	 *
 	 * @deprecated Use waitUntilComplete() instead.
--- a/packages/happy-dom/test/window/DetachedWindowAPI.test.ts
+++ b/packages/happy-dom/test/window/DetachedWindowAPI.test.ts
@@ -288,6 +288,12 @@ describe('DetachedWindowAPI', () => {
 		});
 	});
 
+	describe('waitForNavigation()', () => {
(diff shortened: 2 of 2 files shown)


capricorn86 (changes requested):
I'm not sure it's a good idea to expose this method on the `happyDOM` object. It's intentional not to expose it.

An instance of `Window` is considered detached from a Browser and can't be navigated. Navigating would destroy the `Window` instance and the user will not have a reference to the new `Window`. This is why the user needs to create a `Browser` to keep track of the navigation and get access the new `Window` instance.

mixelburg:
Thanks for the detailed explanation — that makes sense. Navigating a detached Window would destroy the instance, and the Browser API already handles this correctly. Closing this one. Appreciate the review!

PR #2266, opened 2026-07-31
fix: [#2373] `ResponseCache` must not evict responses revalidatable via `Last-Modified`

### LINKED ISSUE #2373: Cached responses with only `Last-Modified` are evicted on expiry instead of being revalidated
### Bug report

A cached response that carries `Last-Modified` but no `ETag` is deleted from the
cache the moment it expires, so the next request is a plain `GET` instead of a
conditional `If-Modified-Since` revalidation.

```js

// Server sends: Cache-Control: max-age=0, Last-Modified: <date>  (no ETag)
const window = new Window({ url: 'https://example.com/' });
await window.fetch('/data');          // stored, immediately expired
await window.fetch('/data');          // happy-dom: uncached GET (200)
                                      // browsers: conditional GET -> 304
```

`ResponseCache.get()` already handles this correctly: an expired entry with a
`Last-Modified` date is kept as `stale` for revalidation. `ResponseCache.add()`
does not — its eviction check looks at `ETag` only, so it purges the entry
before it can ever be revalidated.

### Expected

Same as `get()`: keep an expired entry as long as it has an `ETag` **or** a
`Last-Modified` to revalidate against.

marciomazza:
### Description

Resolves #2373

`ResponseCache.get()` keeps an expired entry that has a `Last-Modified` date as
`stale` so the next request can revalidate it with `If-Modified-Since`.
`ResponseCache.add()` didn't: its eviction check only looked at `ETag`, so a
response with `Cache-Control: max-age=0` + `Last-Modified` and no `ETag` (a
common "always revalidate" setup) was purged immediately, turning every
follow-up request into an uncached `GET` instead of a conditional one.

```js

const window = new Window({ url: 'https://example.com/' });
await window.fetch('/data');   // Last-Modified, no ETag, max-age=0
await window.fetch('/data');   // happy-dom: plain GET 200 — browsers: 304 revalidation
```

Root cause: `add()`'s expiry cleanup dropped any expired entry without an
`ETag`, ignoring `lastModified` — out of sync with the check in `get()`.

Fix: add the same `!cachedResponse.lastModified` term to `add()`'s eviction
condition, so an entry is only dropped when it has neither validator.

### Tests

Added 2 tests (`test/fetch/Fetch.test.ts`, `test/fetch/SyncFetch.test.ts`): an
already-expired `Last-Modified` response with no `ETag` and no `Cache-Control`
must stay cached so the next request is a conditional `If-Modified-Since` `GET`,
not a plain refetch. Both fail on `master` (the second request goes out
unconditional) and pass with this change. Full `npm test` passes.

<details>
<summary>Verification</summary>

The new tests use an `Expires` date in the past rather than a sub-millisecond
`max-age`, so they fail deterministically on `master` instead of only under
scheduling latency — this same eviction is the cause of the intermittent
`SyncFetch.test.ts` / `Fetch.test.ts` cache-revalidation failures on `master`.
`npm test` (happy-dom package) — 302 files, 7721 tests green.
Behavior checked against HTTP caching semantics (RFC 9111 §4.3): a stored
response with a validator must be revalidated on expiry, not discarded.
</details>

### AI

I used **Claude Code** to write this, over multiple iterations. I guided and reviewed
it myself at every step.

### Tests

  - _Added 2 deterministic regression tests (`Fetch.test.ts`, `SyncFetch.test.ts`). Not runnable in a browser console (server-side response cache); verified against RFC 9111 §4.3 conditional-request semantics instead._

### Title

### AI tools


marciomazza:
@capricorn86 

The flaky CI is biting me a lot. I try to keep all my fix/feature branches rebased and conflict free, so this issue shows up all the time. There's another CI flakiness fix in #2270.

Please take a look at them.

Issue #2199, opened 2026-06-11
Performance: every Node eagerly allocates 8 query-selector cache Maps it usually never uses, inflating parse memory and allocation cost


TrevorBurnham:
**Describe the bug**

The base `Node` class eagerly initializes its query-selector cache as a class-field initializer, allocating **seven `new Map()` instances plus a wrapper object on every single node construction**:

```ts
// packages/happy-dom/src/nodes/node/Node.ts (master, lines 85–103)
public [PropertySymbol.cache]: {
  querySelector: Map<string, ICachedQuerySelectorResult>;
  // ...8 typed members...
} = {
  querySelector: new Map(),
  querySelectorAll: new Map(),
  matches: new Map(),
  elementsByTagName: new Map(),
  elementsByTagNameNS: new Map(),
  elementByTagName: new Map(),
  elementById: new Map(),
  computedStyle: null
};
```

Because this is a field initializer on the base `Node` class (`export default class Node extends EventTarget {`, master line 33), it is emitted into the constructor and runs for **every** node instance — `Element`, `Text`, `Comment`, `Attr`, `Document`, `DocumentFragment`, etc. — regardless of subclass.

In practice only a small fraction of nodes are ever the receiver of a `querySelector` / `querySelectorAll` / `matches` / `getElementsByTagName*` / `getElementById` / `getComputedStyle` call. Text nodes, comment nodes, attribute nodes, and never-queried elements never read or write their own cache, yet each pays for 7 `Map` allocations at construction time. This is wasted heap and wasted allocation/GC work, most visible during HTML parsing where the document is dominated by text and attribute nodes.

**To Reproduce**

Parse a representative page and count `Map` allocations during the parse, and measure retained heap per held document.

```js
// Count Map allocations during one parse (after a warmup parse)
const RealMap = globalThis.Map;
let mapCount = 0;
globalThis.Map = new Proxy(RealMap, {
  construct(t, a, nt) { mapCount++; return Reflect.construct(t, a, nt); }
});
mapCount = 0;
new DOMParser().parseFromString(html, 'text/html'); // html = a ~193KB real-world page
globalThis.Map = RealMap;
console.log(mapCount); // master: ~76,355 ; lazy: ~31,149
```

```js
// Retained heap per held parsed document (run node with --expose-gc)
const settle = () => { for (let i = 0; i < 6; i++) globalThis.gc(); };
[...]

PR #2271, opened 2026-08-03
fix: [#2199] Allocates the Node cache and EventTarget listeners lazily

mtfranchetto:
### Description

Every `Node` allocates 11 empty `Map`s in its constructor: 7 query cache maps on `Node`, plus 2 + 2 listener tables on `EventTarget`. An empty `Map` is around 193 bytes in V8 because the hash table backing store is allocated up front, so that's roughly 2.3kB per node before it holds anything, and most nodes never use either one. `Text`, `Comment` and `Attr` are `Node`s too and can never be the receiver of a query.

This makes all three lazy. Each one keeps its existing symbol as an accessor that allocates on first access, backed by a new nullable `*Store` symbol, so none of the ~118 existing read sites change. `Node[PropertySymbol.cache]` also gets a setter, because the custom element upgrade path transfers the cache between two elements.

Parsing 500 `<p>` elements, each with a text node and a comment:

| | master | this branch |
| --- | --- | --- |
| `Map`s during the parse | 29,515 | 2,011 |
| `Map`s during `happyDOM.close()` | 52 | 19 |

Where I hit this: I work at [Ceros](https://www.ceros.com), and [Flex](https://www.ceros.com/flex/), our interactive content platform, uses happy-dom to server-render published pages of web components in Node. A 1,175-element document becomes around 46,000 happy-dom nodes, most of them Lit marker comments and attributes. A heap snapshot attributed 530,903 `Map` objects and 95MB to these fields, and peak RSS for one render dropped from 2,600MB to 1,162MB, which moved our OOM ceiling by about 4x.

We have been running the same change in production on Flex for a week now, as a patched dependency, with no new issues, across every published page our customers render.

This supersedes #2200. The `Node` half implements the design @TrevorBurnham described in #2199, written fresh against current master since that PR has been sitting a while with a red build. The `EventTarget` half, the extra guards and the tests are new.

A handful of call sites have to read the store directly instead of the accessor, otherwise they re-materialise exactly what the laziness avoids. Those are commented inline, and I broke each one to confirm a test catches it.

`npm run compile`, `npm run lint` and `npm test` pass locally: 297 files, 7,634 tests. Happy to split the `EventTarget` half into its own PR if you'd rather take these separately.

---

AI disclosure, per the contribution guidelines: I used Claude Code on this, for the investigation and to draft the change. I reviewed and tested everything myself and I'm responsible for the content.


mtfranchetto:
Hey @capricorn86, sorry for the ping! Before you spend time on a proper review, would you be open to this direction in
principle? Not asking about the implementation, just whether this is something you'd consider for happy-dom at all.

Asking now because we're running this as a patch in our codebase and we can't keep it forever. If it's not a good fit for the project that's completely fair, we'd just need to know early. And if it is, happy to split the EventTarget part out or rework it however you prefer. Thanks for all the work on happy-dom :pray:

PR #2272, opened 2026-08-05
fix: [#2264] MutationObserver callback being GC'd due to orphaned WeakRef

### LINKED ISSUE #2264: MutationObserver callbacks can be silently GC'd (orphaned `WeakRef`)
## Describe the bug

`MutationObserverListener`'s constructor wraps each listener's mutation-delivery callback in a `WeakRef`:

```ts
// packages/happy-dom/src/mutation-observer/MutationObserverListener.ts
this.mutationListener = {
    options: init.options,
    callback: new WeakRef((record: MutationRecord) => this.report(record))
};
```

This `mutationListener` object is stored on the observed `Node` (`Node[PropertySymbol.mutationListeners]`, via `[PropertySymbol.observeMutations]`) and consulted at mutation time:

```ts
// packages/happy-dom/src/nodes/node/Node.ts, [PropertySymbol.reportMutation] (~line 822)
const callback = mutationListener.callback.deref();
if (callback) {
    ...
    callback(record);
}
```

The problem: the arrow function `(record) => this.report(record)` is a **freshly allocated closure with no strong referent anywhere in the object graph** except this one `WeakRef`. It is never stored as a property of `this` (`MutationObserverListener`) or anywhere else. Meanwhile, everything *around* it — the `MutationObserver`, its `MutationObserverListener`, and the target `Node` — can all still be perfectly alive and strongly referenced (e.g. by application code holding `this.observer = new MutationObserver(...)`), and it doesn't matter: nothing on that live object graph points at the closure itself, so V8 is free to collect it the moment a GC cycle runs, at any time after `observe()` returns. Once collected, `deref()` returns `undefined` forever — mutation delivery for that listener silently and permanently stops. No error, no warning, and `disconnect()`/re-`observe()` doesn't help, since each `observe()` call allocates a new equally-orphaned closure.

This is not a "the observer went out of scope, so weak semantics correctly let it die" situation — the *observer itself* can be alive and referenced for the entire lifetime of the program and still lose delivery, because the thing being weakly held was never reachable from the observer in the first place.

`MutationObserver` already keeps a strong `#listeners: MutationObserverListener[]` array (`MutationObserver.ts`), so the `MutationObserverListener` instance itself *is* reachable for as long as its `MutationObserver` is. The natural fix that matches this existing design is to make the `WeakRef` point at `this` (already correctly kept alive elsewhere) rather than at an orphan closure — see "Suggested fix" under Additional context.

## To Reproduce

No UI steps — this is a Node-side GC timing bug. Minimal reproduction (stock `Window`, no other libraries involved):

```js
// node --expose-gc repro.mjs

const window = new Window();
const target = document.createElement("div");
document.body.appendChild(target);

let fired = false;
new window.MutationObserver(() => {
    fired = true;
}).observe(target, { childList: true });

// The MutationObserver, its listener, and the target are all still alive and
// referenced here -- yet the delivery closure has no strong referent.
await new Promise((resolve) => setTimeout(resolve, 0));
global.gc();

target.appendChild(document.createElement("span"));
await new Promise((resolve) => setTimeout(resolve, 0));

console.log(fired ? "PASS: callback fired" : "FAIL: callback silently dropped");
```

1. Save as `repro.mjs` next to a `node_modules` containing `happy-dom`.
2. Run `node --expose-gc repro.mjs`.
3. See `FAIL: callback silently dropped` instead of `PASS`.

Reliably prints `FAIL` (15/15 runs against happy-dom 20.11.1 / Node 26.5.0). The `await new Promise(r => setTimeout(r, 0))` before `global.gc()` matters: a microtask-only yield (`queueMicrotask`/bare `Promise.resolve().then()`) is not enough to reproduce, only an actual macrotask-boundary tick (`setTimeout`/`setImmediate`) is — plausibly because something in happy-dom's own internal async bootstrapping holds a transient reference until that tick completes. This isn't a contrived requirement: any real application does many macrotask-boundary things (a real HTTP round-trip, `setTimeout`, etc.) between setting up an observer and its next relevant mutation, so this reproduces the real-world failure mode rather than a synthetic one.

## Expected behavior

`fired` should be `true` — a `MutationObserver` callback must keep firing for as long as the observer itself (and its `observe()` target) are reachable, regardless of how many GC cycles run in between, per [the spec](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) and every real browser's behavior.

## Device and details

- OS: Linux (not OS-specific — pure V8/GC timing bug, expected to reproduce on any platform)
- Node version: 26.5.0
- Package version: happy-dom 20.11.1 (current `master` as of this writing — the relevant code is unchanged since the file was introduced, see "History" below)

### History (why does this exist?)

Checked upstream commit history for `MutationObserverListener.ts` — the `WeakRef` was present from the file's very first commit ([`6f99ffd5`](https://github.com/capricorn86/happy-dom/commit/6f99ffd5), "BREAKING CHANGE: [#1332] Implements remaining HTML elements (#1359)", 2024-08-20), a 305-file mechanical PR with an unrelated purpose (adding basic support for dozens of missing HTML element classes) and zero PR/issue discussion. The file it replaced, `MutationListener.ts`, held the same callback as a plain strong `#callback` field with no `WeakRef` at all — the weak-wrapping was introduced incidentally while splitting the dispatch payload into the new `IMutationListener` interface, not as a deliberate, discussed tradeoff. We did not find any commit message, linked issue, or PR review comment anywhere in the repo's history that explains an intended tradeoff for this `WeakRef`, so this report doesn't attempt to preserve one — there doesn't appear to be one documented to preserve.

That said, there's a plausible (unconfirmed — inferred, not found in any commit) *design goal* behind reaching for `WeakRef` here at all: real `MutationObserver` is specifically designed to be GC-transparent — an observer that the application no longer references should become eligible for collection without requiring an explicit `disconnect()` call, and an observed node shouldn't be forced to stay alive just because something is watching it. That's a deliberate part of why `MutationObserver` replaced the old Mutation Events API, and `WeakRef` is exactly the primitive you'd reach for to implement "only notify while someone external still cares." If that was the goal, the bug is that the `WeakRef` wraps the *wrong* object: to implement that property correctly, the weakly-held thing needs to be unreachable through every path *except* "someone external still cares" (e.g. the `MutationObserverListener` behind `MutationObserver#listeners`). Instead it wraps a closure that was never reachable through that path in the first place, so it can't distinguish "still cared about" from "abandoned" at all — it fails identically in both cases. The suggested fix below happens to also correctly implement this GC-transparency property, if that was the intent — it isn't just a bug fix, it's what actually makes the weak reference meaningful for the first time.

### Suggested fix

Point the `WeakRef` at the `MutationObserverListener` instance itself (already kept alive via `MutationObserver#listeners`), not an orphan closure — and rename the field `callback` → `listener`, since it no longer holds a callback:

```diff
 // IMutationListener.ts
 export default interface IMutationListener {
 	options: IMutationObserverInit;
-	callback: WeakRef<(record: MutationRecord) => void>;
+	listener: WeakRef<MutationObserverListener>;
 }
```

```diff
 // MutationObserverListener.ts constructor
 this.mutationListener = {
 	options: init.options,
-	callback: new WeakRef((record: MutationRecord) => this.report(record))
+	listener: new WeakRef(this)
 };
```

```diff
 // Node.ts -- 4 read sites, all `.callback.deref()` -> `.listener.deref()`.
 // 3 are truth-tests only (appendChild/removeChild/insertBefore), e.g.:
-if (mutationListener.options?.subtree && mutationListener.callback.deref()) {
+if (mutationListener.options?.subtree && mutationListener.listener.deref()) {

 // The 4th, in [PropertySymbol.reportMutation], is the one that invokes it
 // (all 3 switch branches get the same swap):
-const callback = mutationListener.callback.deref();
-if (callback) {
+const listener = mutationListener.listener.deref();
+if (listener) {
   switch (record.type) {
     case MutationTypeEnum.childList:
       if (mutationListener.options.childList) {
-        callback(record);
+        listener.report(record);
       }
       break;
     // attributes / characterData branches: same swap
   }
 }
```

No other files reference this field. Bonus: this also happens to correctly implement the GC-transparency property speculated about under "History" — the weakly-referenced object now has a real strong-reference path exactly when its `MutationObserver` is genuinely still alive.

**Verified**: applied the functional part (`WeakRef(this)`, `.report()` call) to a patched scratch copy of the compiled `lib/` output and reran the repro — 10/10 `PASS` vs. 15/15 `FAIL` unpatched (happy-dom 20.11.1). The `callback`→`listener` rename has no runtime effect, so wasn't separately re-tested.

### Impact

Any application that: (1) creates a `MutationObserver`, (2) calls `observe()`, and (3) does anything at all involving a macrotask tick (essentially always true — `setTimeout`, a network request, etc.) before the next mutation, is at risk of the observer permanently and silently going dead, with no error surfaced anywhere. We hit this via Alpine.js's reactivity engine, which disconnects/reconnects its single global `MutationObserver` on every DOM write — each reconnect allocates a fresh, equally at-risk closure, making the failure effectively constant under any real usage pattern, not a rare edge case.

### Local workaround already in place

We're currently unblocking ourselves by wrapping `MutationObserver.prototype.observe` at runtime to temporarily swap in a strong-reference `WeakRef`-shaped shim for the duration of each `observe()` call, restoring the real `WeakRef` in a `finally`. This is scoped narrowly (only while `observe()` runs) so it doesn't affect happy-dom's other, legitimate `WeakRef` uses (query-result caching). This is a workaround, not a substitute for the upstream fix — it doesn't change the underlying weak-reference target, so it wouldn't be an appropriate patch to submit as-is; it's just what we're using to unblock ourselves pending a real fix.

### References

- Root cause: `packages/happy-dom/src/mutation-observer/MutationObserverListener.ts`, constructor.
- Dispatch site: `packages/happy-dom/src/nodes/node/Node.ts`, `[PropertySymbol.reportMutation]` (~line 822).
- Introducing commit: [`6f99ffd5`](https://github.com/capricorn86/happy-dom/commit/6f99ffd5) (PR [#1359](https://github.com/capricorn86/happy-dom/pull/1359), no related discussion).
- Existing correct pattern for the same "proxy identity" class of mistake, already used elsewhere in this same file's neighborhood: `MutationObserver#listeners` keeping `MutationObserverListener` instances strongly alive.

---

*Drafted with AI assistance; root cause, repro, and fix were verified directly against the happy-dom source and compiled output (see "Verified, not just proposed" above) before writing this up.*

mixelburg:
## What

Fixes #2264 — MutationObserverListener wraps the mutation delivery callback in a WeakRef without a strong reference, allowing V8 to GC the callback at any time. When collected, deref() returns undefined permanently and mutation delivery silently stops.

## How

Added a #listenerCallback private field to hold a strong reference to the arrow function. The strong reference is cleared in destroy(). This maintains the original weak semantics between listener and node while preventing premature GC.

## Changes

- Store listener callback as private field alongside WeakRef
- Clear in destroy()
- Test: verifies mutations are delivered when observer is externally dereferenced

*AI assisted: Claude Code*

### DIFF
--- a/packages/happy-dom/src/mutation-observer/MutationObserverListener.ts
+++ b/packages/happy-dom/src/mutation-observer/MutationObserverListener.ts
@@ -18,6 +18,7 @@ export default class MutationObserverListener {
 	#records: MutationRecord[] = [];
 	#destroyed = false;
 	#microtaskQueued = false;
+	#listenerCallback: (record: MutationRecord) => void;
 
 	/**
 	 * Constructor.
@@ -38,9 +39,10 @@ export default class MutationObserverListener {
 	}) {
 		this.options = init.options;
 		this.target = init.target;
+		this.#listenerCallback = (record: MutationRecord) => this.report(record);
 		this.mutationListener = {
 			options: init.options,
-			callback: new WeakRef((record: MutationRecord) => this.report(record))
+			callback: new WeakRef(this.#listenerCallback)
 		};
 		this.#window = init.window;
 		this.#observer = init.observer;
@@ -103,6 +105,7 @@ export default class MutationObserverListener {
 		this.#destroyed = true;
 		this.options = null!;
 		this.target = null!;
+		this.#listenerCallback = null!;
 		this.mutationListener = null!;
 		this.#window = null!;
 		this.#observer = null!;
--- a/packages/happy-dom/test/mutation-observer/MutationObserver.test.ts
+++ b/packages/happy-dom/test/mutation-observer/MutationObserver.test.ts
@@ -1,6 +1,6 @@
 import Window from '../../src/window/Window.js';
 import type Document from '../../src/nodes/document/Document.js';
-import MutationObserver from '../../src/mutation-observer/MutationObserver.js';
+import type MutationObserver from '../../src/mutation-observer/MutationObserver.js';
 import type MutationRecord from '../../src/mutation-observer/MutationRecord.js';
 import { beforeEach, describe, it, expect } from 'vitest';
 
@@ -327,6 +327,25 @@ describe('MutationObserver', () => {
 					type: 'attributes'
 				}
 			]);
+		});
+
+		it('Continues delivering mutations after the listener closure is no longer externally referenced.', async () => {
+			const records: MutationRecord[] = [];
+			const div = document.createElement('div');
+			let observer: MutationObserver | null = new window.MutationObserver((mutationRecords) =>
+				records.push(...mutationRecords)
+			);
+			observer.observe(div, { attributes: true });
+			// Dereference observer — only internal MutationObserverListener holds the callback
+			observer = null;
+			div.setAttribute('attr1', 'value1');
+			div.setAttribute('attr2', 'value2');
+
+			await new Promise((resolve) => setTimeout(resolve, 1));
+
+			expect(records.length).toBe(2);
+			expect(records[0].attributeName).toBe('attr1');
+			expect(records[1].attributeName).toBe('attr2');
 		});
 	});
 


marciomazza:
Comparing this fix to the approach suggested in #2264:

**This PR:** keeps the `WeakRef` pointed at the callback closure, and adds a new private field `#listenerCallback` to hold a separate strong reference to that same closure (cleared in `destroy()`). `Node.ts`'s dispatch site is unchanged — it still does `callback(record)` after deref.

**Suggested in #2264:** point the `WeakRef` at `this` (the `MutationObserverListener` instance) instead of at the closure, and call `listener.report(record)` at the dispatch site. No extra field, no extra closure — relies on `MutationObserver#listeners` already keeping the instance alive, the same pattern already used elsewhere in this file.

Both fix the GC bug. The difference is where the strong reference lives: a new field pointing at the closure (this PR) vs. removing the closure and pointing the `WeakRef` at the already-alive instance (#2264's suggestion).

mixelburg:
Thanks for the detailed comparison @marciomazza! 

The approach suggested in #2264 (pointing the WeakRef at `this` and calling `listener.report(record)`) is indeed cleaner — it removes the extra closure and relies on the already-existing strong reference from `MutationObserver#listeners`, which is the same pattern already used elsewhere in the file. No extra `#listenerCallback` field needed.

Would you prefer I refactor this PR to use the #2264 approach? Happy to make that change.

capricorn86 (approved):
Thank you for your contribution @mixelburg!

@marciomazza I think I prefer the fix with using the report callback as strong reference, as using a reference to itself as a strong reference opens up for more scenarios.

capricorn86 (approved):
Thank you for your contribution @mixelburg!

@marciomazza I think I prefer the fix with using the report callback as strong reference, as using a reference to itself as a strong reference opens up for more scenarios.

marciomazza:
Awesome. That allows me to remove my local patch. Thanks to you both.

PR #2274, opened 2026-08-06
fix: [#2273] Preserves case in XMLDocument.createElement

### LINKED ISSUE #2273: XMLDocument.createElement() uppercases element names
**Describe the bug**

`XMLDocument.createElement()` uppercases element names even though XML element names are case-sensitive.

When parsing a document as `text/xml`, calling `createElement('myNode')` results in an element whose `nodeName` and `tagName` are `"MYNODE"` instead of `"myNode"`.

[StackBliz](https://stackblitz.com/edit/stackblitz-starters-jfsffl66?file=index.js)

**To Reproduce**

1. Create an XML document:

   ```ts
   const doc = new DOMParser().parseFromString('<root/>', 'text/xml');
   ```
2. Create an element:

   ```ts
   const element = doc.createElement('myNode');
   ```
3. Inspect the element:

   ```ts
   console.log(element.nodeName);
   console.log(element.tagName);
   ```

Current output:

```text
MYNODE
MYNODE
```

**Expected behavior**

The element name should be preserved exactly as provided when using an `XMLDocument`.

Expected output:

```text
myNode
myNode
```

This is the behavior defined for XML documents, where element names are case-sensitive.

**Device and details:**

* OS: Ubuntu 20.04.0 LTS
* Node version: 22.22.3
* Package version: 20.11.1

**Additional context**

This appears to be HTML `createElement()` behavior being applied to `XMLDocument`. Parsing existing XML preserves the original case correctly, but creating new elements programmatically does not.

mixelburg:
XMLDocument.createElement() was uppercasing element names because it inherited Document.createElement() which lowercases and then createElementNS() which uppercases. For XML documents element names are case-sensitive per spec.

Fix: Override createElement() in XMLDocument to skip lowercasing, and preserve original case in createElementNS() when the document is an XML document (checked via contentType).

Closes #2273

### DIFF
--- a/packages/happy-dom/src/nodes/document/Document.ts
+++ b/packages/happy-dom/src/nodes/document/Document.ts
@@ -1958,9 +1958,15 @@ export default class Document extends Node {
 					options && options.is ? String(options.is) : qualifiedName
 				);
 
+				const isXMLDocument =
+					this[PropertySymbol.contentType] === 'application/xml' ||
+					this[PropertySymbol.contentType] === 'text/xml';
+
 				if (customElementDefinition) {
 					const element = new customElementDefinition.elementClass();
-					element[PropertySymbol.tagName] = StringUtility.asciiUpperCase(qualifiedName);
+					element[PropertySymbol.tagName] = isXMLDocument
+						? qualifiedName
+						: StringUtility.asciiUpperCase(qualifiedName);
 					element[PropertySymbol.localName] = localName;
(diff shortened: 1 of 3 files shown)


mixelburg:
Hi! Just checking in — is there anything else needed from my side on this one?

mixelburg:
Closing for now as this hasn't received any feedback. Happy to reopen if there's interest.

PR #2275, opened 2026-08-06
fix: [#2232] Prevent infinite loop on cyclic CSS variable references

### LINKED ISSUE #2232: getComputedStyle infinite-loops on cyclic custom property (var()) references
Note: I used Claude to help file this issue, but I've reviewed and agree with the findings.

**Describe the bug**

When two custom properties reference each other (`--a: var(--b)` / `--b: var(--a)`), calling `window.getComputedStyle(element)` — or reading any property value off the returned declaration — hangs forever. In a real browser, cyclic `var()` references resolve to the property's "guaranteed-invalid value" (effectively falling back to initial/inherited value) rather than looping infinitely.

This was found while migrating a test suite from jsdom to happy-dom: a test asserting graceful handling of cyclic `var()` references (which passed under jsdom, since jsdom returns the raw stored value without attempting resolution) hung indefinitely under happy-dom.

**To Reproduce**

```js
import { Window } from "happy-dom"

const window = new Window()
const { document } = window
document.documentElement.style.setProperty("--a", "var(--b)")
document.documentElement.style.setProperty("--b", "var(--a)")

// hangs forever
const styles = window.getComputedStyle(document.documentElement)
console.log(styles.getPropertyValue("--a"))
```

Root cause looks to be in `CSSStyleDeclarationComputedStyle.parseCSSVariablesInValue`:

```js
parseCSSVariablesInValue(value, cssVariables) {
    let newValue = value;
    let match;
    while ((match = newValue.match(SINGLE_CSS_VARIABLE_REGEXP)) != null) {
        newValue = newValue.replace(match[0], cssVariables[match[1]] || '');
    }
    ...
}
```

When `cssVariables["--a"]` is `"var(--b)"` and `cssVariables["--b"]` is `"var(--a)"`, the `while` loop alternates between `"var(--a)"` and `"var(--b)"` forever — there's no cycle/visited-set guard or iteration cap.

**Expected behavior**

`getComputedStyle` should not hang. Ideally it would detect the cycle and treat the value as invalid (per spec), but at minimum it should cap substitution iterations so a cyclic reference fails fast instead of hanging the process.

**Environment**
- happy-dom: 20.10.6
- Node: 20.19.4
- Reproduced standalone (no test runner) and inside a vitest 4.1 suite

mk-techi:
### Description

`getComputedStyle()` hangs forever when two custom properties reference each other (`--a: var(--b)` / `--b: var(--a)`).

**Root cause:** `CSSStyleDeclarationComputedStyle.parseCSSVariablesInValue()` substitutes `var()` references in a loop until no more matches are found. With a cyclic reference the substitution alternates between `var(--a)` and `var(--b)` indefinitely — there is no cycle guard or iteration cap.

**Fix:** Added an iteration cap (`MAX_CSS_VARIABLE_ITERATIONS = 100`) to both substitution loops. When the cap is exceeded the value resolves to an empty string, matching the "guaranteed-invalid value" behaviour described in the CSS spec, and consistent with how the existing code already handles undefined variables. Both loops are guarded, since the fallback-value loop (`var(--a, red)`) is vulnerable to the same cycle.

Resolves #2232

### Tests
- [x] Make sure to add tests for your changes. Added two regression tests covering cyclic references with and without fallback values. Verified in Chrome that a real browser resolves cyclic `var()` references to an empty value rather than hanging.
- [x] Run the tests with `npm test` locally to make sure that all tests pass before submitting the PR. The 5 tests in `CSSStyleDeclarationComputedStyle.test.ts` pass. A few unrelated tests fail on my Windows machine (line-ending and path-resolution issues in `BrowserFrame`, `ModuleURLUtility`, `ResponseCacheFileSystem`) — they fail identically on a clean checkout and are untouched by this change.

### Title

### AI tools
- [x] I used an AI assistant to help diagnose the issue and draft the fix and tests. I reviewed the code, reproduced the bug locally before the change, and verified the fix and the test suite myself.

### DIFF
--- a/packages/happy-dom/src/css/declaration/computed-style/CSSStyleDeclarationComputedStyle.ts
+++ b/packages/happy-dom/src/css/declaration/computed-style/CSSStyleDeclarationComputedStyle.ts
@@ -199,12 +199,17 @@ export default class CSSStyleDeclarationComputedStyle {
 
 			Object.assign(cssProperties, rulesAndProperties.properties);
 
+			const resolvedCSSVariables = this.resolveCSSVariables(cssProperties);
+
 			for (const { name, value, important } of rules) {
 				if (
 					(<any>CSSStyleDeclarationElementInheritedProperties)[name] ||
 					parentElement === targetElement
 				) {
-					const parsedValue = this.parseCSSVariablesInValue(value.trim(), cssProperties);
+					// CSS variables have already been resolved, which is important as a variable taking part in a cyclic reference is invalid and therefore not allowed to use its fallback value.
+					const parsedValue = name.startsWith('--')
+						? resolvedCSSVariables[name]
+						: this.parseCSSVariablesInValue(value.trim(), resolvedCSSVariables);
 
 					if (parsedValue && (!propertyManager.get(name)?.important || important)) {
 						propertyManager.set(name, parsedValue, important);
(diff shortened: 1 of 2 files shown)


packages/happy-dom/src/css/declaration/computed-style/CSSStyleDeclarationComputedStyle.ts
```diff
@@ -460,13 +461,23 @@ export default class CSSStyleDeclarationComputedStyle {
 	private parseCSSVariablesInValue(value: string, cssVariables: { [k: string]: string }): string {
 		let newValue = value;
 		let match: RegExpMatchArray | null;
+		let iterations = 0;
 
 		while ((match = newValue.match(SINGLE_CSS_VARIABLE_REGEXP)) != null) {
+			if (++iterations > MAX_CSS_VARIABLE_ITERATIONS) {
```
capricorn86: As it's not really possible to set a max for the amount of iterations (someone could technically use 100 variables), it's probably better to resolve the variables and check for cyclic references.

It's not super easy to solve this as a value could contain many of the same variable without the variable having cycling references (e.g. `position: var(--zero) var(--zero) var(--zero) var(--zero)`).

This could be one solution:

```typescript
    private resolveCSSVariables(cssVariables: Record<string, string>): Record<string, string> {
          const resolvedCSSVariables: Record<string, string> = {};

          for(const key of Object.keys(cssVariables)) {
                const usedValues = new Set();
                let newValue = cssVariables[key];
                // Without fallback value - E.g. var(--my-var)
                 while ((match = value.match(SINGLE_CSS_VARIABLE_REGEXP)) != null) {
                        const name = match[1]];
                        // Check for cyclic reference
                        if(usedValues.has(name) {
                              newValue = '';
                              break;
                        }
                        usedValues.add(name);
                        newValue = newValue.replace(match[0], cssVariables[name] || '');
                 }
                // Fallback value - E.g. var(--my-var, #FFFFFF)
                 while ((match = value.match(SINGLE_CSS_VARIABLE_REGEXP)) != null) {
                        const name = match[1] || match[2];
                        // Check for cyclic reference
                        if(usedValues.has(name)) {
                              newValue = '';
                              break;
                        }
                        usedValues.add(name);
                        newValue = newValue.replace(match[0], cssVariables[name] || '');
                 }
                 resolvedCSSVariables[key] = newValue;
          }

          return resolvedCSSVariables;
    }
```
// At line 200 to resolve the variable object before it's used
```typescript
      Object.assign(cssProperties, rulesAndProperties.properties);
      const resolvedCSSVariables = resolveCSSVariables(cssProperties);
```
```typescript
	private parseCSSVariablesInValue(value: string, cssVariables: { [k: string]: string }): string {
         return newValue.replace(SINGLE_CSS_VARIABLE_REGEXP, (_, p) => cssVariables[p] || '')).replace(CSS_VARIABLE_REGEXP, (_, p1, p2) => cssVariables[p1 || p2] || '');
	}
```

capricorn86 (changes requested):
Thank you for your contribution @mk-techi! 

I found one thing we need to consider.

packages/happy-dom/src/css/declaration/computed-style/CSSStyleDeclarationComputedStyle.ts
mk-techi: Thanks for the review! Reworked using the pre-resolution approach. Three deviations from your snippet, all with tests:

1. Reference chain instead of a per-variable set of names. A set of names breaks on the exact case you flagged: String.replace(match[0], …) only replaces the first occurrence, so with var(--zero) var(--zero) the name --zero is seen twice and the value collapses to ''. I track the variables currently being resolved (removed once resolved) plus memoization, which also handles the diamond case (--top: var(--zero), --left: var(--zero), --position: var(--top) var(--left)).

2. parseCSSVariablesInValue keeps its loop. A non-global .replace() only substitutes the first occurrence, and a single pass doesn't resolve nesting like var(--a, var(--b)), covered by the existing "parses nested variables correctly" test. The loop can no longer run forever since injected values are pre-resolved and contain no substitutable var().

3. Custom properties bypass re-parsing (line 209). They also go through rules, so --a was re-parsed and cssVariables['--b'] || 'red' resolved to red, regressing the cyclic-with-fallback test. Per spec a variable inside the cycle is invalid and does not use its fallback, while a normal property referencing a cyclic variable does, both behaviours are now preserved.

Added 3 tests: repetition inside a variable value, two variables pointing to the same variable, and repetition inside a property value. 8 tests in the file and all 451 CSS tests pass.

PR #2279, opened 2026-08-07
fix: [#2255] Keep strong ref to MutationObserverListener callback

### LINKED ISSUE #2255: MutationObserver silently stops after garbage collection: internal listener callback is only held via WeakRef
**Describe the bug**

`MutationObserverListener` stores its internal dispatch callback **only inside a `WeakRef`** — no strong reference to the closure exists anywhere:

```js
// lib/mutation-observer/MutationObserverListener.js (v20.11.1)
this.mutationListener = {
    options: init.options,
    callback: new WeakRef((record) => this.report(record))
};
```

The inline arrow function is reachable only through the `WeakRef`, so any garbage-collection pass may reclaim it. Every dispatch and subtree-maintenance site guards with `mutationListener.callback.deref()` (e.g. in `lib/nodes/node/Node.js`), so once the closure is collected, an **active, connected observer silently stops receiving mutations** — subtree bookkeeping even prunes the listener from nodes afterwards.

We hit this in CI: a `MutationObserver`-based test failed intermittently, but only under V8 coverage instrumentation on one runner (higher allocation pressure → more frequent GC). Once the callback is collected, no amount of waiting recovers the observer. This is very likely the same root cause as #1878 — Bun's GC just reclaims the weakly-held closure sooner.

**To Reproduce**

```js
// repro.mjs — run: node --expose-gc repro.mjs

const window = new Window()
document.body.innerHTML = '<p>hello</p>'

let calls = 0
const observer = new window.MutationObserver(() => {
  calls += 1
})
observer.observe(document.body, {
  characterData: true,
  childList: true,
  subtree: true,
})

document.querySelector('p').firstChild.data = 'first'
await new Promise(resolve => setTimeout(resolve, 20))
console.log('after first mutation:', calls) // 1

global.gc()
await new Promise(resolve => setTimeout(resolve, 20))

document.querySelector('p').firstChild.data = 'second'
await new Promise(resolve => setTimeout(resolve, 20))
console.log('after gc + second mutation:', calls) // still 1 — observer is dead
```

Output on happy-dom 20.11.1 / Node 24.16.0:

```
after first mutation: 1
after gc + second mutation: 1
```

Without the `global.gc()` call the second mutation is delivered (`calls === 2`). In real code the same loss happens whenever a natural GC pass runs between mutations, which is why it surfaces as flaky tests rather than a hard failure.

**Expected behavior**

A connected `MutationObserver` keeps receiving mutations regardless of garbage collection. Keeping one strong reference to the closure on the listener instance fixes it while preserving the existing `WeakRef` plumbing:

```js
this.reportCallback = (record) => this.report(record);
this.mutationListener = {
    options: init.options,
    callback: new WeakRef(this.reportCallback)
};
```

We are running exactly this as a `pnpm patch` in our repo and the intermittent failures disappeared (verified with the `--expose-gc` repro above: fails before, passes after). Happy to open a PR if this shape works for you.

**Device:**

- OS: reproduced on macOS 15 (arm64) and GitHub Actions `windows-latest`
- Node.js: 24.16.0 (also observed on the 22/24/26 CI matrix; `--expose-gc` makes it deterministic)
- happy-dom: 20.11.1

mixelburg:
MutationObserverListener stores its dispatch callback only inside a WeakRef, so GC can collect it and silently stop the observer. Changed  private field to public  to keep a strong reference while preserving the existing WeakRef plumbing. Fixes #2255.

### DIFF
--- a/packages/happy-dom/src/mutation-observer/MutationObserverListener.ts
+++ b/packages/happy-dom/src/mutation-observer/MutationObserverListener.ts
@@ -12,13 +12,13 @@ export default class MutationObserverListener {
 	public target: Node;
 	public options: IMutationObserverInit;
 	public mutationListener: IMutationListener;
+	public reportCallback: (record: MutationRecord) => void;
 	#window: BrowserWindow;
 	#observer: MutationObserver;
 	#callback: (record: MutationRecord[], observer: MutationObserver) => void;
 	#records: MutationRecord[] = [];
 	#destroyed = false;
 	#microtaskQueued = false;
-	#listenerCallback: (record: MutationRecord) => void;
 
 	/**
 	 * Constructor.
@@ -39,10 +39,10 @@ export default class MutationObserverListener {
 	}) {
 		this.options = init.options;
 		this.target = init.target;
-		this.#listenerCallback = (record: MutationRecord) => this.report(record);
+		this.reportCallback = (record: MutationRecord) => this.report(record);
 		this.mutationListener = {
 			options: init.options,
-			callback: new WeakRef(this.#listenerCallback)
+			callback: new WeakRef(this.reportCallback)
 		};
 		this.#window = init.window;
 		this.#observer = init.observer;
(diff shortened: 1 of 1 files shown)


mixelburg:
Hi! Wanted to check whether this one needs any further changes from me.

mixelburg:
Closing since there's been no feedback on this. Happy to reopen if there's interest.

PR #2280, opened 2026-08-08
fix: [#2256] Preserve var() fallback values in inline style declarations

### LINKED ISSUE #2256: var() with a fallback is dropped from inline style declarations

mixelburg:
Fixes #2256.

The CSS variable parser in `CSSStyleDeclarationValueParser.getVariable()` only matched `var(--name)` without fallback values. When a value like `var(--x, 10px)` was set via inline style, the regex didn't match and the entire declaration was dropped.

Fixed by adding a balanced-parentheses parser fallback when the regex doesn't match, preserving the full `var()` expression including any fallback values with nested parentheses (e.g. `var(--primary, rgb(255, 0, 0))`).

Added two tests for fallback preservation.

### DIFF
--- a/packages/happy-dom/src/css/declaration/property-manager/CSSStyleDeclarationValueParser.ts
+++ b/packages/happy-dom/src/css/declaration/property-manager/CSSStyleDeclarationValueParser.ts
@@ -429,10 +429,32 @@ export default class CSSStyleDeclarationValueParser {
 	 * @returns Parsed value.
 	 */
 	public static getVariable(value: string): string | null {
+		// Fast path: simple var(--name) without fallback
 		const cssVariableMatch = value.match(CSS_VARIABLE_REGEXP);
 		if (cssVariableMatch) {
 			return `var(${cssVariableMatch[1]})`;
 		}
+
+		// Slow path: var() with fallback or nested parens in value (e.g. var(--x, 10px), var(--x, rgb(0,0,0)))
+		if (value.startsWith('var(')) {
+			let depth = 0;
+			for (let i = 0; i < value.length; i++) {
+				if (value[i] === '(') {
+					depth++;
+				} else if (value[i] === ')') {
+					depth--;
+					if (depth === 0) {
+						// Verify there's a valid variable name between 'var(' and ',' or ')'
+						const inner = value.slice(4, i);
+						if (/^\s*--[^),]/.test(inner)) {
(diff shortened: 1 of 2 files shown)


mixelburg:
Hi! Anything else I can do to move this one forward?

mixelburg:
Closing this — issue #2256 was closed as *completed* upstream yesterday. The CSS value parser was refactored and `var()` fallback handling moved into `CSSVariableFormatter`, which now round-trips the inline `var(--x, 10px)` case this PR was targeting. Thanks for the time on this one — happy to reopen or send a fresh PR against the new architecture if the nested-fallback case is still wanted.

PR #2283, opened 2026-08-12
fix: allow explicit element types for querySelector

### LINKED ISSUE #2282: Compiler error for `querySelector<K extends keyof IHTMLElementTagNameMap>`
**Describe the bug**
`document.querySelector<HTMLInputElement>('selector')` gives compiler error

**To Reproduce**
```
import { HTMLInputElement, Window } from 'happy-dom';

async function main() {
  const window = new Window({ url: 'https://...' });
  await window.happyDOM.waitUntilComplete();
  document.querySelector<HTMLInputElement>('selector').value = 'xx';
}
```
gives the following error:
```
Type HTMLInputElement does not satisfy the constraint
keyof ISVGElementTagNameMap
. (ts 2344)
```

**Expected behavior**
This should typecheck.

**Device and details:**

- OS: mac
- Node version: v23.7.0
- Package version: 20.11.2

**Additional context**
tsconfig.json: `"strict": true`

cyphercodes:
## Summary

Fixes #2282

Allows explicit element type arguments on string-selector overloads, so calls like `document.querySelector<HTMLInputElement>("#id")` type-check instead of requiring the generic to be a tag-name key.

## Testing

- `npx vitest run test/query-selector/QuerySelector.test.ts`
- `npm run compile`
- `git diff --check HEAD~1..HEAD`

I also ran local TypeScript smoke checks while developing the overload change.

### DIFF
--- a/packages/happy-dom/src/nodes/document-fragment/DocumentFragment.ts
+++ b/packages/happy-dom/src/nodes/document-fragment/DocumentFragment.ts
@@ -146,7 +146,7 @@ export default class DocumentFragment extends Node {
 	 * @param selector CSS selector.
 	 * @returns Matching elements.
 	 */
-	public querySelectorAll(selector: string): NodeList<Element>;
+	public querySelectorAll<E extends Element = Element>(selector: string): NodeList<E>;
 
 	/**
 	 * Query CSS selector to find matching elments.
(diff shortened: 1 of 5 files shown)


cyphercodes:
I inspected the failing build logs. The failed Node 20 run was in an unrelated @happy-dom/server-renderer timeout-debug expectation, while the Node 22/24 jobs were cancelled after the Node 20 failure. I reran the local PR branch checks and they now pass without code changes:\n\n- npm ci\n- npm run compile\n- npm run lint\n- npm run test\n- git diff --check\n\nI also verified current upstream master passes npm run test locally in the same checkout, so I’m treating the CI failure as transient/no PR code change needed.

capricorn86 (approved):
Thank you for your contribution @cyphercodes!

PR #2285, opened 2026-08-12
feat: [#2284] Adds support for cascade layers (@layer)

### LINKED ISSUE #2284: Rules inside @layer blocks are parsed and silently dropped
`CSSParser` has no case for `@layer`, so a layer block falls through to the unknown-rule arm, which consumes the block without adding its children to `cssRules`. Every rule wrapped in a cascade layer is silently invisible to `getComputedStyle`, while the same rule outside a layer resolves.

```js

const window = new Window();
const { document } = window;

document.head.innerHTML =
  "<style>@layer a { .x { color: rgb(255, 0, 0); } }</style>";
document.body.innerHTML = '<div class="x"></div>';

const color = window.getComputedStyle(document.body.children[0]).color;
console.log(color); // "" — expected "rgb(255, 0, 0)"
// The control: the same rule without @layer resolves to rgb(255, 0, 0).
```

Layered CSS is now the default shape of Tailwind 4 output and common in design systems, so a page's whole stylesheet can vanish from computed styles under test.

Two possible shapes for a fix. The full one is a real `CSSLayerBlockRule` with layer priority in the cascade. The minimal one treats a layer block as a transparent group: nested rules attach where the layer's parent would attach them, which makes layered rules apply while leaving layer *priority* unmodeled. I'll put up a PR with the minimal shape; happy to rework it toward the full rule if that's the direction you want.

Possibly related: #1862 hits the same missing `@layer` support through `insertRule`.

Gamote:
Fixes #2284.

This adds support for cascade layers, both the block form and the statement form, and makes `getComputedStyle` respect the layer order.

There are two new rule classes. `CSSLayerBlockRule` extends `CSSGroupingRule` and carries `name`, and `CSSLayerStatementRule` extends `CSSRule` and carries `nameList`. Both return 0 from `type`, the same as `CSSScopeRule` and `CSSContainerRule` do for rules with no legacy type number.

In `CSSParser` the block form follows the shape of the `@scope` arm above it. The statement form (`@layer base, theme;`) needed the tokenizer to match `;` as well, and to consume the text before it only when that text starts with `@`, since a semicolon ends an at-rule and nothing else. A stray semicolon still invalidates the rule after it, as `CSSParserInput` expects, and `@charset` and `@import` now stop swallowing the rule that follows them.

In `CSSStyleDeclarationComputedStyle` the layers form a tree, numbered so that a layer's own rules rank after the rules of its sublayers and unlayered rules rank last, and the existing sort by specificity takes the layer index as its first key. `!important` inverts the layer order, so it gets a second pass over the sources that contain `!important`, with the style attribute appended again so an inline `!important` still wins. That pass only runs when more than one matched source carries `!important`.

The tests cover both parser forms, unnamed and nested layers, a layer inside `@media`, a layer ignored inside `@keyframes`, a file for each new rule class, `insertRule` for both forms, and six cases in `getComputedStyle`: a layered rule applying, an unlayered rule beating a layered one, order between two layers, a statement fixing that order, a layer beating its own sublayer, and `!important` reversing the order. Following the contributing guidelines I ran those six in Chrome 151 as well, taking the CSS and the expected values from the test file so the two can't drift apart, and it agrees with all six. Timing for stylesheets that use no layers is unchanged.

`@import ... layer()` and `revert-layer` aren't covered here.

A few things I'd like your opinion on:

1. `type` returning 0 follows `CSSScopeRule`, but it leaves `instanceof` as the only discriminator in `parseCSSRules`, which now has three cases behind one number. Happy to add an enum member if you'd prefer.
2. Layer order is worked out while walking the rules that match, so a layer declared only inside a media query that doesn't match never takes its place in the order. Walking every rule separately would close that, at the cost of an extra pass.
3. The second parse for `!important` is the part I'm least comfortable with, since `getComputedStyle` is a hot path.
4. Layers are per encapsulation context in the spec, and this keeps one tree per document, so a shadow root shares the document's layer order.

### DIFF
--- a/packages/happy-dom/src/PropertySymbol.ts
+++ b/packages/happy-dom/src/PropertySymbol.ts
@@ -96,6 +96,7 @@ export const clientWidth = Symbol('clientWidth');
 export const clientLeft = Symbol('clientLeft');
 export const clientTop = Symbol('clientTop');
 export const name = Symbol('name');
+export const nameList = Symbol('nameList');
 export const specified = Symbol('specified');
 export const adoptedStyleSheets = Symbol('adoptedStyleSheets');
 export const implementation = Symbol('implementation');
--- a/packages/happy-dom/src/css/declaration/computed-style/CSSStyleDeclarationComputedStyle.ts
+++ b/packages/happy-dom/src/css/declaration/computed-style/CSSStyleDeclarationComputedStyle.ts
@@ -20,6 +20,8 @@ import MediaQueryList from '../../../match-media/MediaQueryList.js';
 import WindowBrowserContext from '../../../window/WindowBrowserContext.js';
 import type CSSSupportsRule from '../../rules/CSSSupportsRule.js';
 import CSSScopeRule from '../../rules/CSSScopeRule.js';
+import CSSLayerBlockRule from '../../rules/CSSLayerBlockRule.js';
+import CSSLayerStatementRule from '../../rules/CSSLayerStatementRule.js';
 import type CSSStyleSheet from '../../CSSStyleSheet.js';
 
 const CSS_MEASUREMENT_REGEXP = /[0-9.]+(px|rem|em|vw|vh|%|vmin|vmax|cm|mm|in|pt|pc|Q)/g;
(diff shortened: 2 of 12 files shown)


capricorn86 (changes requested):
Thank you for your contribution @Gamote!

I think we need to implement the full solution with `CSSLayerBlockRule` and `CSSLayerStatementRule`. We would also need to handle this in `CSSStyleDeclarationComputedStyle.`.

Gamote:
> I think we need to implement the full solution with `CSSLayerBlockRule` and `CSSLayerStatementRule`. We would also need to handle this in `CSSStyleDeclarationComputedStyle.`.

@capricorn86 Makes sense, I've reworked it that way. The PR now adds `CSSLayerBlockRule` and `CSSLayerStatementRule`, handles both the block and the statement form in the parser, and orders the cascade by layer in `CSSStyleDeclarationComputedStyle`, including the way `!important` inverts that order. Tests come with it, and I ran the six cascade cases in Chrome as well to check the expectations against a real browser.

There are a few questions at the end of the description where I wasn't sure which way you'd want it, mainly the `type` value and the extra pass for `!important`.

PR #2286, opened 2026-08-12
fix: [#1490] Different ports/parent domains are different origins

### LINKED ISSUE #1490: CORS "same origin" check is incorrect (ports + subdomains)
**Describe the bug**
Determining if a request is cross-origin is [currently implemented](https://github.com/capricorn86/happy-dom/blob/e6f91276d7b4549d17f5d67bd2f909723f4940b3/packages/happy-dom/src/fetch/utilities/FetchCORSUtility.ts#L21-L25) by checking the `hostname` of the current page and the remote URL. This is not in accordance with the specification, where the origin is considered to be scheme + domain + port. In addition, the check currently allows subdomains - which is also not in accordance with it.

**To Reproduce**
1. Do a request from `http://localhost:1234` to `http://localhost:9876`
2. Request should have triggered cross-origin behavior (port differs), but currently does not
4. Do a request from `http://sub.some.host` to `http://some.host`
5. Request should have triggered cross-origin behavior (hostname differs), but currently does not
**Expected behavior**
Should follow browser behavior for cross-origin requests

rexxars:
### Description

The CORS check is currently wrong about at least two scenarios:

- A request to the same hostname but different protocol **should** be treated as "not the same origin", but is not
- A request from a subdomain to a parent **should** be treated as "not the same origin", but is not

This PR uses the `origin` property on the URL for a cheap and simple comparison.

Resolves #1490

> [!WARNING]
> **Theoretically** this could be seen as a breaking change, but it not follows the spec. I'm not sure whether or not to consider that breaking, so will leave the decision up to you.

### Tests

### Title

### AI tools

### DIFF
--- a/packages/happy-dom/src/fetch/utilities/FetchCORSUtility.ts
+++ b/packages/happy-dom/src/fetch/utilities/FetchCORSUtility.ts
@@ -18,10 +18,6 @@ export default class FetchCORSUtility {
 			return false;
 		}
 
-		return (
-			(originURL.hostname !== targetURL.hostname &&
-				!originURL.hostname.endsWith(targetURL.hostname)) ||
-			originURL.protocol !== targetURL.protocol
-		);
+		return originURL.origin !== targetURL.origin;
 	}
 }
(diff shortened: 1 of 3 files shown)


capricorn86 (approved):
Thank you for your contribution @rexxars!

PR #2303, opened 2026-08-21
fix: [#2257] Keeps throwing for an invalid selector after the first call

### LINKED ISSUE #2257: querySelector stops throwing for an invalid selector after the first call (regression in 20.4.0)

Kjubikstronk:
### Description

`querySelector()` added its cache entry before parsing the selector. The entry starts out as `{ result: { element: null } }`, the parse then throws for an invalid selector, and the entry is left behind looking exactly like a successful search that matched nothing. Every later call for that string reads it and returns `null`.

`match()` had the same order, which is why `closest()` and `matches()` behave the same way after their first call.

Both now parse before the entry is added, which is what `querySelectorAll()` already did — it computes `groups` first, and that is why it was unaffected.

```js

for (const attempt of [1, 2, 3]) {
  try { document.querySelector('[') } catch { /* threw */ }
}
// before: threw, then null, then null
// after:  throws every time
```

Resolves #2257

The cache is per-`Window`, so the practical shape of this was a test file where the first invalid selector threw and every later one silently passed validation.

### Tests

- [x] Make sure to add tests for your changes.

The added test covers `querySelector`, `querySelectorAll`, `matches` and `closest` over three calls each. The rest of the suite has the same failures before and after this change.

### AI disclosure

Claude assisted with this change. I reviewed and tested it.


Kjubikstronk:
The first red run was mine — a line in the added test was too long for prettier, and `--max-warnings 0` turns that into a failed build. Fixed in the last commit.

The second one looks unrelated to this change. `build (22)` failed in `@happy-dom/server-renderer` on "Handles debug error caused while waiting for the page to complete", where the expectation is `1 task did not end in time` and it got `2`, the extra one being the navigation task from `BrowserWindow`. Builds 20 and 24 were cancelled when 22 failed, so there's only the one data point on that run.

This change reorders selector parsing relative to the cache write in `QuerySelector`. For a selector that parses, the behaviour is identical; the only difference is that one that doesn't parse now throws every time instead of once. That test doesn't use selectors, and it passed 12/12 over three local runs against this branch.

For what it's worth, #2285 also failed on `build (22)` in the same window, on `Fetch` cache revalidation, which is a different timing-sensitive test. Happy to rebase or push again if you'd rather see a clean run before looking at it.

PR #2335, opened 2026-08-29
feat: Adds support for Element.animate

### LINKED ISSUE #502: `animate()` is missing on Element
Hi, using happy-dom with vitest and @testing-library/preact. When testing a component that calls `Element.animate()`, I get a `TypeError: animate is not a function`

https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/element/Element.ts doesn't have `animate()` - could this be added please?

capricorn86:


diegohaz:
Hey @capricorn86! Just a heads-up: I think this feature may have broken animations somehow. See https://github.com/ariakit/ariakit/pull/7389 (failing job: https://github.com/ariakit/ariakit/actions/runs/33586635693/job/100112147384)

I can investigate it later (maybe it's a problem with my code). I’m just leaving this here in case you have time to look at it first. Thanks!

PR #2341, opened 2026-09-01
fix: [#2256] Fixes issue when parsing CSS fallback variables

### LINKED ISSUE #2256: var() with a fallback is dropped from inline style declarations

capricorn86:
### Description

Resolves #issue-number

- [ ] Read the [contributing guidelines](https://github.com/capricorn86/happy-dom/blob/master/CONTRIBUTING.md).
- [ ] It's really useful if your PR references an issue where it is discussed ahead of time.
- [ ] Please check [Allow edits by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to make review process faster. Note that this option is not available for repositories that are owned by Github organizations.

### Tests

- [ ] Make sure to add tests for your changes. Run your test in a real browser to make sure that the test tests what a real browser would do (e.g. by running the code in the browser console).
- [ ] Run the tests with `npm test` locally to make sure that all tests pass before submitting the PR.

### Title

- [ ] The title of the pull request should be in the format of "type: [#issue] description". The type can be `feat`, `fix`, `chore` or `BREAKING CHANGE`. The issue is optional and can be omitted if the pull request does not relate to an issue.
- [ ] The title should be concise and descriptive. The title will be used when generating release notes. Make sure that the title is easily understood by users of the library.

### AI tools

- [ ] Please disclose in the PR description that you used AI tools to generate code. This is important for transparency and to ensure that the generated code meets the quality standards of the project.


cossssmin:
Hi there,

FYI this one broke our CI in [maizzle/framework](https://github.com/maizzle/framework/pull/1862) when Dependabot bumped `happy-dom` from `v20.11.2` to `v20.13.2`.

Since `v20.12.1`, `CSSStyleDeclaration` validates property names against the Chromium list and silently drops anything it doesn't know. `mso-font-width` is one of those - it's an Outlook-only property that we (and pretty much everyone doing HTML email) rely on, and in our usage Vue's `mount()` sets inline styles through `style.cssText`, so it never makes it into the DOM anymore:

```js
el.style.cssText = 'display: inline-block; mso-font-width: 100%'
el.getAttribute('style')
// v20.11.2: "display: inline-block; mso-font-width: 100%;"
// v20.13.2: "display: inline-block;"
```

To be fair, real browsers drop unknown properties from the CSSOM too, so I'm not saying this is a bug on your end really, just thought I'd write in since maybe others woke up to something similar and would find the explanation useful. We've worked around it by asserting against SSR output instead.

Thanks for the great work on happy-dom :)

PR #2360, opened 2026-09-03
fix: [#2306] Return an ArrayBuffer as XMLHttpRequest response for res…

### LINKED ISSUE #2306: XMLHttpRequest response is a Uint8Array instead of an ArrayBuffer when responseType is "arraybuffer"
**Describe the bug**

When `responseType` is set to `"arraybuffer"`, the `response` property returns a `Uint8Array` instead of an `ArrayBuffer`. Code that checks `response instanceof ArrayBuffer`, or passes the response to APIs that expect an `ArrayBuffer`, behaves differently than in browsers.

**To Reproduce**

1. Run the following code in a Happy DOM environment against any URL:

```javascript
const request = new XMLHttpRequest();
request.responseType = 'arraybuffer';
request.open('GET', '/some/path', true);
request.addEventListener('load', () => {
  console.log(request.response instanceof ArrayBuffer); // false, logs a Uint8Array
  console.log(request.response.constructor.name); // "Uint8Array"
});
request.send();
```

**Expected behavior**

`request.response` should be an `ArrayBuffer`, as defined by the XHR specification (https://xhr.spec.whatwg.org/#the-response-attribute) and as implemented in browsers.

**Device and details:**

- OS: macOS
- Node version: 20.17.0
- Package version: master (d43c1065)

**Additional context**

In `XMLHttpRequestResponseDataParser.parse()`, the `arraybuffer` case creates an `ArrayBuffer`, copies the data into a `Uint8Array` view, and then returns the view instead of the buffer:

https://github.com/capricorn86/happy-dom/blob/d43c1065c1ec2eb0d3a1c1a5e1489f3ec4103c5c/packages/happy-dom/src/xml-http-request/XMLHttpRequestResponseDataParser.ts#L31-L36

Returning `newAB` instead of `view` should fix it. Note that one existing test decodes the response with `TextDecoder`, which accepts both types, so it doesn't catch this.

This was found with the help of Claude Code while working on #2293.

Agnik47:
### Description

`XMLHttpRequestResponseDataParser.parse()` allocated an `ArrayBuffer`, copied the response data into a `Uint8Array` view of it, and then returned the view instead of the buffer:

```js
const request = new XMLHttpRequest();
request.responseType = 'arraybuffer';
request.open('GET', '/some/path', true);
request.addEventListener('load', () => {
  request.response instanceof ArrayBuffer; // false, was a Uint8Array
});
request.send();
```

As a result, `XMLHttpRequest.response` was a `Uint8Array` when `responseType` was set to `"arraybuffer"`. This broke code that checks `response instanceof ArrayBuffer` or passes the response to APIs expecting an `ArrayBuffer`.

Note that `parse()` already declared `ArrayBuffer` in its return type, so the implementation contradicted its own signature. TypeScript did not catch this because `Uint8Array` is assignable to the `object` member of that union.

The buffer is now returned, as defined by the [XMLHttpRequest specification](https://xhr.spec.whatwg.org/#the-response-attribute) and as implemented in browsers.

Resolves #2306

### Tests

Rather than adding a new test, the existing one was strengthened. It was already named "Returns ArrayBuffer when `responseType` is set to `arrayBuffer`", but never asserted the type — `TextDecoder.decode()` accepts both an `ArrayBuffer` and a `Uint8Array`, so it passed either way. It now asserts the response type, fails on `master`, and passes with this change.

### Title

### AI tools

Claude Code was used to assist with this change. I reviewed the specification behaviour, ran the test suite locally, and verified that the strengthened test fails on `master` before submitting.

### DIFF
--- a/packages/happy-dom/src/xml-http-request/XMLHttpRequestResponseDataParser.ts
+++ b/packages/happy-dom/src/xml-http-request/XMLHttpRequestResponseDataParser.ts
@@ -33,7 +33,7 @@ export default class XMLHttpRequestResponseDataParser {
 				const newAB = new ArrayBuffer(options.data.length);
 				const view = new Uint8Array(newAB);
 				view.set(options.data);
-				return view;
+				return newAB;
 			case XMLHttpResponseTypeEnum.blob:
 				try {
 					return new options.window.Blob([new Uint8Array(options.data)], {
--- a/packages/happy-dom/test/xml-http-request/XMLHttpRequest.test.ts
+++ b/packages/happy-dom/test/xml-http-request/XMLHttpRequest.test.ts
@@ -161,6 +161,7 @@ describe('XMLHttpRequest', () => {
 				request.open('GET', REQUEST_URL, true);
 
 				request.addEventListener('load', () => {
+					expect(request.response instanceof ArrayBuffer).toBe(true);
 					expect(responseText).toBe(new TextDecoder().decode(<ArrayBuffer>request.response));
 					resolve(null);
 				});


Agnik47:
The `build (20)` failure looks unrelated to this change.

It fails in `@happy-dom/server-renderer`:

```
FAIL test/ServerRendererPage.test.ts > render() > Handles debug error caused while waiting for the page to complete in browser mode.
AssertionError: expected { …(9) } to deeply equal { …(9) }
  "error": "Error: The maximum time was reached for \"waitUntilComplete()\"."
```

A few reasons I don't think this PR caused it:

- The `@happy-dom/server-renderer` package does not reference `XMLHttpRequest`, `responseType` or `arraybuffer` anywhere, so the changed branch in `XMLHttpRequestResponseDataParser.parse()` is not reachable from that test.
- The test is timing sensitive by construction: it sets `browser.settings.debug.traceWaitUntilComplete = 100` against a page running `setTimeout(t, 10)` recursively, and then asserts on the resulting trace.
- #2359 is branched from the same base commit (eac5a380) and passed on Node 20, 22 and 24.
- `build (22)` and `build (24)` were cancelled after `build (20)` failed, so they never reported a result.

Could you re-run the job when you get a chance? I do not have permission to re-run it from my side.

PR #2361, opened 2026-09-04
fix: [#2356] Ignore null and undefined when setting input.files

### LINKED ISSUE #2356: `HTMLInputElement.files = null` breaks `.value` getter on file inputs
**Describe the bug**

Setting `input.files = null` (or `undefined`) on a `type="file"` input is stored as-is. Reading `.value` afterwards throws:

```
TypeError: Cannot read properties of null (reading 'length')
    at HTMLInputElement.get value (HTMLInputElement.ts:821)
```

Per the [HTML spec](https://html.spec.whatwg.org/multipage/input.html#dom-input-files), setting `files` to `null` must be ignored. Browsers keep the existing `FileList`.

Real-world trigger: Svelte 5 `bind:files` runs `input.files = get()` on mount, which is `undefined` when nothing is bound yet. Any component using `bind:files` + `bind:value` on a file input crashes on mount under happy-dom.

**To Reproduce**

```js

const input = new Window().document.createElement('input');
input.type = 'file';
input.files = null;
input.value; // throws
```

**Expected behavior**

`input.files = null` is a no-op. `input.value` returns `''`.

**Suggested fix**

`HTMLInputElement.ts` `set files`: early return when `files` is null/undefined (spec says also ignore when type is not `file`).

**Device**

- happy-dom 20.13.2
- Bun 1.x / Node 22, macOS

mixelburg:
Setting `input.files = null` (or `undefined`) on a file input is stored as-is, so reading `.value` afterwards throws. Per spec these assignments are no-ops, and browsers keep the existing `FileList` — this also matches jsdom, which ignores the write when the input isn't a file type.

The setter now returns early for null/undefined and for non-file inputs. Added tests covering both cases.

Closes #2356

### DIFF
--- a/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
+++ b/packages/happy-dom/src/nodes/html-input-element/HTMLInputElement.ts
@@ -145,6 +145,9 @@ export default class HTMLInputElement extends HTMLElement {
 	 * @param files Files.
 	 */
 	public set files(files: FileList) {
+		if (files === null || files === undefined || this.type !== 'file') {
+			return;
+		}
 		this[PropertySymbol.files] = files;
 	}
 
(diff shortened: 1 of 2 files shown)


mixelburg:
Hi! The `build (20)` failure here looks unrelated to this change — it's in `@happy-dom/server-renderer` (`ServerRendererBrowser.test.ts` / `ServerRendererPage.test.ts`), specifically the timing-sensitive `waitUntilComplete()` tests: `"2 tasks did not end in time"` vs the expected `"1 task"`, plus a null `.error` in the debug-error case. This PR only touches the `HTMLInputElement.files` setter and its tests.

The sibling PR #2362 ran the same server-renderer tests on Node 20 the same day and passed, which points to a flake rather than a real regression. Would you mind re-running the `build (20)` job? Happy to dig in further if it turns out not to be a flake.

PR #2364, opened 2026-09-05
fix: [#2363] Avoids cloning all properties in CSSPropertyManager.toString()

### LINKED ISSUE #2363: CSSStyleDeclaration.setProperty() is O(n) per call, so building a declaration is O(n²)
## Describe the bug

Every `setProperty()` / `removeProperty()` on an element's `style` reserializes the whole declaration to update the `style` attribute. `CSSStyleDeclarationPropertyManager.toString()` starts with `this.clone()`, which deep-copies all properties via `JSON.parse(JSON.stringify(...))`, then walks the shorthand table and builds the string. Cost per call grows with the number of declared properties, so setting n properties costs O(n²).

This hits any code that sets many CSS custom properties on one element, e.g. a design-system theme inlining a few hundred `--*` variables. React sets `style` props one `setProperty` at a time, so a component rendering `style={themeVariables}` pays the full quadratic cost on every mount. In our suite one 15-test file spent 9 of its 10 seconds inside `clone` + `toString` (CPU profile: 6.4s in `clone`, 1.9s in `toString`).

## To Reproduce

```js
import {Window} from "happy-dom";
const {document} = new Window();
for (const n of [100, 200, 400, 800, 1600]) {
  const el = document.createElement("div");
  const start = performance.now();
  for (let i = 0; i < n; i++) {
    el.style.setProperty(`--v${i}`, `rgb(${i % 255} 0 0)`);
  }
  console.log(`${n}\t${(performance.now() - start).toFixed(0)} ms`);
}
```

happy-dom 20.7.0, Node 24.20.0, Apple Silicon:

```
100	5 ms
200	13 ms
400	44 ms
800	163 ms
1600	848 ms
```

Doubling n roughly quadruples the time. Chromium does the same 1600 calls in well under a millisecond.

## Expected behavior

`setProperty` should be roughly O(1), or at least not deep-clone the declaration on every call.

## Where the time goes

`lib/css/declaration/property-manager/CSSStyleDeclarationPropertyManager.js`:

```js
clone() {
    const clone = new _class();
    clone.properties = JSON.parse(JSON.stringify(this.properties));   // ~70% of the cost
    ...
}

toString() {
    const result = [];
    const clone = this.clone();   // only needed so the shorthand loop can call clone.remove()
    ...
}
```

`CSSStyleDeclaration.setProperty()` / `removeProperty()` then call `propertyManager.toString()` and `element.setAttribute('style', ...)` unconditionally.

## Possible fixes

1. **Lazy clone in `toString()`** (small, behaviour-preserving). `remove()` only deletes keys, so a shallow copy is enough, and it's only needed if a shorthand (margin, padding, border, …) actually matches:

   ```js
   let clone = this;
   ...
   if (clone === this) { clone = this.clone(); }
   clone.remove(shorthandProperty);
   ```

   Measured locally: the 10s test file drops to 2.1s. Still O(n²) but with a much smaller constant.

2. **Lazy attribute reflection** (bigger). Mark the declaration dirty in `setProperty` and serialize the `style` attribute only when it is read (`getAttribute('style')`, `outerHTML`, `cssText`, MutationObserver delivery). Makes `setProperty` O(1).

Happy to send a PR for option 1 if you're open to it.

## Related

`CSSStyleDeclarationComputedStyle` collects every `style, link[rel="stylesheet"]` element and ignores `HTMLStyleElement.disabled`, `CSSStyleSheet.disabled` and the `media` attribute, so a disabled sheet still contributes to `getComputedStyle()`. Separate bug, can file separately if useful.

erictheswift:
### Description

`CSSPropertyManager.toString()` started with `this.clone()`, a `structuredClone` of every property, on each call. `CSSStyleDeclaration.setProperty()` / `removeProperty()` call `toString()` to refresh the `style` attribute, so each call was O(n) in declared properties and filling a declaration with n properties was O(n²).

The clone exists only so the shorthand loop can call `clone.remove()` without mutating the real declaration. `remove()` only deletes keys, so the clone is now created lazily, the first time a shorthand (margin, padding, border, …) actually matches. Declarations without shorthands, e.g. a theme setting hundreds of `--*` variables, are serialized without any cloning. Output is unchanged.

`setProperty()` of n custom properties on one element, Node 24, Apple Silicon:

| n | before | after |
|---|---|---|
| 400 | 55 ms | 13 ms |
| 800 | 211 ms | 42 ms |
| 1600 | 799 ms | 164 ms |

Still O(n²) since the attribute is reserialized per call; making that lazy is a bigger change and out of scope here.

Resolves #2363

### Tests

### Title

### AI tools

Written with Claude Code assistance (profiling, patch, test); reviewed and measured by hand.

### DIFF
--- a/packages/happy-dom/src/css/declaration/property-manager/CSSPropertyManager.ts
+++ b/packages/happy-dom/src/css/declaration/property-manager/CSSPropertyManager.ts
@@ -576,8 +576,10 @@ export default class CSSPropertyManager {
 	 */
 	public toString(): string {
 		const result = [];
-		const clone = this.clone();
 		const properties: { [k: string]: ICSSPropertyValue } = {};
+		// Cloning is only needed when a shorthand matches, as remove() mutates the properties.
+		// Skipping it keeps toString() cheap for declarations with many custom properties.
+		let clone: CSSPropertyManager = this;
 
 		for (const shorthandPropertyGroup of TO_STRING_SHORTHAND_PROPERTIES) {
 			for (const shorthandProperty of shorthandPropertyGroup) {
@@ -587,6 +589,9 @@ export default class CSSPropertyManager {
 						const property = clone.get(childShorthandProperty);
 						if (property) {
 							properties[childShorthandProperty] = property;
+							if (clone === this) {
+								clone = this.clone();
+							}
 							clone.remove(childShorthandProperty);
 							isMatch = true;
 						}
@@ -598,6 +603,9 @@ export default class CSSPropertyManager {
 					const property = clone.get(shorthandProperty);
 					if (property) {
 						properties[shorthandProperty] = property;
+						if (clone === this) {
+							clone = this.clone();
+						}
 						clone.remove(shorthandProperty);
 						break;
 					}
--- a/packages/happy-dom/test/css/declaration/CSSStyleDeclaration.test.ts
+++ b/packages/happy-dom/test/css/declaration/CSSStyleDeclaration.test.ts
@@ -3082,6 +3082,24 @@ describe('CSSStyleDeclaration', () => {
 			expect(element.getAttribute('style')).toBe('top: 0px;');
 		});
 
+		it('Sets many custom properties.', () => {
+			const declaration = new CSSStyleDeclaration(PropertySymbol.illegalConstructor, window, {
+				element
+			});
+			const count = 2000;
+
+			for (let i = 0; i < count; i++) {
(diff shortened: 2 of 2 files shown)


capricorn86 (dismissed):
Thank you for your contribution @erictheswift!

capricorn86 (approved):

capricorn86:
I removed the unit test as it doesn't seem to add any value except slowing down the test suite, and it seems like it's causing timeouts.

PR #2367, opened 2026-09-07
fix: [#2366] Invalidate the computed style cache for the whole subtree

### LINKED ISSUE #2366: Computed style cache is only invalidated one level deep, descendants keep stale inherited values (regression in 20.14.0)
**Describe the bug**

Since 20.14.0 (`Improve computed style cache` #2357), changing an element's class or `style` no longer invalidates the computed style cache for its whole subtree, only the element itself and its direct children are recomputed. Anything deeper keeps the previously cached value of inherited properties.

20.13.2 is correct.

**To Reproduce**

```js

const win = new Window()
const doc = win.document
doc.body.innerHTML = `<style>.red { color: rgb(255, 0, 0); }</style>
  <div id="d0" class="red"><div id="d1"><div id="d2"><div id="d3"><span id="d4">x</span></div></div></div></div>`

const ids = ['d0', 'd1', 'd2', 'd3', 'd4']
const color = id => win.getComputedStyle(doc.getElementById(id)).color || '<empty>'

console.log('with .red:', ids.map(id => `${id}=${color(id)}`).join(' '))

doc.getElementById('d0').classList.remove('red')

console.log('.red off :', ids.map(id => `${id}=${color(id)}`).join(' '))
```

20.13.2:

```
with .red: d0=rgb(255, 0, 0) d1=rgb(255, 0, 0) d2=rgb(255, 0, 0) d3=rgb(255, 0, 0) d4=rgb(255, 0, 0)
.red off : d0=<empty> d1=<empty> d2=<empty> d3=<empty> d4=<empty>
```

20.14.0:

```
with .red: d0=rgb(255, 0, 0) d1=rgb(255, 0, 0) d2=rgb(255, 0, 0) d3=rgb(255, 0, 0) d4=rgb(255, 0, 0)
.red off : d0=<empty> d1=<empty> d2=rgb(255, 0, 0) d3=rgb(255, 0, 0) d4=rgb(255, 0, 0)
```

**Expected behavior**

`d2`, `d3` and `d4` inherit `color` from `d0`, so all five should report the same value after the class is removed, as they do in 20.13.2.

**Device and details:**

- OS: macOS 26.6.2
- Node version: 24.20.0
- Package version: 20.14.0 (20.13.2 is fine)

matmannion:
### Description

A cached computed style holds the inherited properties of every ancestor it was computed from, but since #2358 only the mutated element and its direct children are invalidated, so anything deeper keeps stale inherited values.

`CSSComputedStyle` registers each element's cached result on its direct parent's `affectsCache`, and `Node[PropertySymbol.clearCache]()` clears the node's own `cache.computedStyle` plus that `affectsCache`. Between them, invalidation reaches exactly one level down.

Each cached result now records the ancestor result it inherited from:

```ts
export default interface ICachedComputedStyleResult extends ICachedResult {
	result: WeakRef<CSSPropertyManager> | null;
	parent: ICachedComputedStyleResult | null;
}
```

and `getCachedPropertyManager()` walks that chain, refusing a cached result if any ancestor result it was derived from has been invalidated. Nothing is added to the invalidation path, so there is no subtree walk on mutation; the extra cost is following a few pointers on a cache read.

Resolves #2366

The change and its test were written with Claude Code, and reviewed and verified by me before submitting.

### Tests

### Title

### AI tools

### DIFF
--- a/packages/happy-dom/src/css/declaration/computed-style/CSSComputedStyle.ts
+++ b/packages/happy-dom/src/css/declaration/computed-style/CSSComputedStyle.ts
@@ -24,6 +24,7 @@ import type CSSStyleSheet from '../../CSSStyleSheet.js';
 import CSSVariableFormatter from '../property-manager/utilities/CSSVariableFormatter.js';
 import type HTMLElement from '../../../nodes/html-element/HTMLElement.js';
 import type Node from '../../../nodes/node/Node.js';
+import type ICachedComputedStyleResult from '../../../nodes/node/ICachedComputedStyleResult.js';
 
 const CSS_MEASUREMENT_REGEXP = /[0-9.]+(px|rem|em|vw|vh|%|vmin|vmax|cm|mm|in|pt|pc|Q)/g;
 
@@ -275,8 +276,11 @@ export default class CSSComputedStyle {
 					}
 				}
 
-				const cachedResult = {
-					result: new WeakRef(propertyManager)
+				const cachedResult: ICachedComputedStyleResult = {
+					result: new WeakRef(propertyManager),
+					parent: previousParent
+						? previousParent.element![PropertySymbol.cache].computedStyle
+						: null
 				};
 				parentElement.element![PropertySymbol.cache].computedStyle = cachedResult;
 				parentElement.element![PropertySymbol.ownerDocument][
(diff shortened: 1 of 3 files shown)


capricorn86 (approved):
Thank you for your contribution @matmannion!

I refactored the solution to work with how cache is managed in happy-dom. Otherwise this may fail in other scenarios.

PR #2395, opened 2026-09-08
fix: [#2344] Resolve every var() in a value, not only the first one

### LINKED ISSUE #2344: Regression in 20.12.1: only the first var() in a value is resolved, shorthand with multiple var() is dropped from getComputedStyle()
**Describe the bug**
Since v20.12.1, when a declaration value contains **two or more `var()` references**, only the first one is resolved. The remaining `var()` stays unresolved in the computed value. For shorthand properties (`padding`, `margin`, …) the partially resolved value then fails validation and the whole declaration is dropped, so `getComputedStyle()` returns `""` for the shorthand **and** for all its longhands.

v20.12.0 resolves all of them correctly. This is a regression introduced by #2341 (the fix for #2256).

This breaks any MUI (Material UI) test suite that uses `cssVariables: true`: MUI's `theme.spacing(a, b)` emits `calc(a * var(--mui-spacing)) calc(b * var(--mui-spacing))`.

**To Reproduce**

```js

const window = new Window();

document.body.innerHTML = `
  <style>:root { --x: 16px; }</style>
  <style>
    .a { padding: calc(0.375 * var(--x)) calc(0.75 * var(--x)); }
    .b { padding: var(--x) var(--x); }
    .c { margin: calc(0.375 * var(--x)) calc(0.75 * var(--x)); }
    .d { padding: calc(0.375 * var(--x)); }
  </style>
  <div class="a"></div><div class="b"></div><div class="c"></div><div class="d"></div>
`;

for (const cls of ["a", "b", "c", "d"]) {
  const cs = window.getComputedStyle(document.querySelector("." + cls));
  console.log(cls, JSON.stringify({ padding: cs.padding, paddingTop: cs.paddingTop, margin: cs.margin }));
}
```

Output with **v20.12.1 / v20.12.2**:

```
a {"padding":"","paddingTop":"","margin":""}
b {"padding":"16px var(--x)","paddingTop":"16px","margin":""}
c {"padding":"","paddingTop":"","margin":""}
d {"padding":"calc(0.375 * 16px)","paddingTop":"calc(0.375 * 16px)","margin":""}
```

Output with **v20.12.0** (and v20.11.x):

```
a {"padding":"calc(0.375 * 16px) calc(0.75 * 16px)","paddingTop":"calc(0.375 * 16px)","margin":""}
b {"padding":"16px","paddingTop":"16px","margin":""}
c {"padding":"","paddingTop":"","margin":"calc(0.375 * 16px) calc(0.75 * 16px)"}
d {"padding":"calc(0.375 * 16px)","paddingTop":"calc(0.375 * 16px)","margin":""}
```

Case `b` shows the root cause directly: the second `var(--x)` is left untouched. The same happens with `element.style.cssText = "padding: calc(0.375 * var(--x)) calc(0.75 * var(--x))"` (the inline declaration is dropped as well).

**Expected behavior**
Every `var()` in the value is resolved, as in v20.12.0: `.a` → `calc(0.375 * 16px) calc(0.75 * 16px)`, `.b` → `16px`, `.c` margin → `calc(0.375 * 16px) calc(0.75 * 16px)`.

**Device and details:**

- OS: macOS 26 (Darwin 25.5.0)
- Node version: 24.16.0
- Package version: 20.12.1, 20.12.2 (20.12.0 works)

**Additional context**
The cause is in `CSSVariableFormatter.resolveVariables()`: after the first `var()` is substituted, the tail of the string is appended verbatim instead of being resolved recursively. Applying `resolveVariables` to the tail fixes all cases above; with that one-line change our full test suite (~2000 tests, MUI v9 with CSS variables) passes on v20.12.2:

```diff
- return `${value.substring(0, variable.index)}${this.resolveVariables(variableValue || fallbackValue, cssVariables)}${value.substring(match.index! + match[0].length)}`;
+ return `${value.substring(0, variable.index)}${this.resolveVariables(variableValue || fallbackValue, cssVariables)}${this.resolveVariables(value.substring(match.index! + match[0].length), cssVariables)}`;
```

### LINKED ISSUE #2351: getComputedStyle() resolves only the first var() occurrence in a declaration
### Description

When a CSS declaration contains more than one `var()` occurrence, `getComputedStyle()` only resolves the first one. Subsequent variables remain unresolved, producing a partially transformed value.

This is separate from #2256. In Happy DOM 20.13.2 the inline declaration now round-trips correctly through `element.style`, but the computed-style path still resolves only one variable.

### Reproduction

```js

const window = new Window();
const element = window.document.createElement('div');

window.document.body.append(element);
element.style.setProperty('--column-width', '10px');
element.style.setProperty('--gutter', '5px');
element.style.width =
  'calc(3 * var(--column-width) + 2 * var(--gutter) + 20px)';

console.log(element.style.width);
console.log(window.getComputedStyle(element).width);
```

### Actual result

```text
calc(3 * var(--column-width) + 2 * var(--gutter) + 20px)
calc(3 * 10px + 2 * var(--gutter) + 20px)
```

Only `--column-width` is resolved. The same happens when the same variable occurs twice:

```js
element.style.width = 'calc(var(--column-width) + var(--column-width))';
console.log(window.getComputedStyle(element).width);
// calc(10px + var(--column-width))
```

### Expected result

All `var()` occurrences should be resolved. Chromium returns `60px` for the first example. Even if Happy DOM intentionally preserves the `calc()` expression rather than evaluating it, both variables should be substituted rather than returning a partially resolved declaration.

### Environment

- Happy DOM: 20.13.2
- Node.js: 24.14.1
- OS: macOS

### Additional context

`CSSVariableFormatter.resolveVariables()` appears to return immediately after replacing the first variable and only recursively processes that variable's replacement value, not the complete resulting declaration:

https://github.com/capricorn86/happy-dom/blob/v20.13.2/packages/happy-dom/src/css/declaration/property-manager/utilities/CSSVariableFormatter.ts#L39-L74

klaesra:
### Description

Since v20.12.1 `CSSVariableFormatter.resolveVariables()` substitutes the first `var()` it finds and returns, leaving the rest of the value untouched. `calc(5 * var(--box) + 6 * var(--gap))` computes to `calc(5 * 20px + 6 * var(--gap))`, and a shorthand such as `padding: calc(0.375 * var(--x)) calc(0.75 * var(--x))` fails validation and is dropped from `getComputedStyle()` together with its longhands.

The remainder of the value is now passed through `resolveVariables()` as well, so every `var()` is substituted; nested fallbacks are unchanged. Browsers substitute all occurrences at computed-value time (Chromium returns the resolved length, as noted in #2351).

Resolves #2344
Resolves #2351

- [x] References issues discussed ahead of time (#2344, #2351).
- [x] Allow edits by maintainers.

### Tests

- [x] Two tests in `CSSComputedStyle.test.ts`: a `calc()` with two variables (and the same variable twice), and a shorthand with `var()` fallbacks after an earlier `var()`. Both fail on master and pass with the fix.
- [x] `npm test` run locally; the only failures are pre-existing Windows-environment tests (virtual server, module loader, timers, file cache) that fail identically on master.

### AI tools

- [x] Written with Claude Code (Anthropic), reviewed and verified by the submitter.

### DIFF
--- a/packages/happy-dom/src/css/declaration/property-manager/utilities/CSSVariableFormatter.ts
+++ b/packages/happy-dom/src/css/declaration/property-manager/utilities/CSSVariableFormatter.ts
@@ -66,7 +66,8 @@ export default class CSSVariableFormatter {
 			if (variable && variable.parentheses === parentheses) {
 				const fallbackValue = value.substring(variable.fallbackIndex, match.index).trim();
 				const variableValue = cssVariables[variable.name];
-				return `${value.substring(0, variable.index)}${this.resolveVariables(variableValue || fallbackValue, cssVariables)}${value.substring(match.index! + match[0].length)}`;
+				// The rest of the value may hold further var() references (e.g. "calc(var(--a) + var(--b))").
+				return `${value.substring(0, variable.index)}${this.resolveVariables(variableValue || fallbackValue, cssVariables)}${this.resolveVariables(value.substring(match.index! + match[0].length), cssVariables)}`;
 			}
 		}
 
--- a/packages/happy-dom/test/css/declaration/computed-style/CSSComputedStyle.test.ts
+++ b/packages/happy-dom/test/css/declaration/computed-style/CSSComputedStyle.test.ts
@@ -98,6 +98,32 @@ describe('CSSComputedStyle', () => {
 			expect(propertyManager.get('background-color')?.value).toBe('rgb(255, 219, 0)');
 		});
 
+		it('Resolves every var() in a value, not only the first one.', () => {
+			document.body.appendChild(element);
+			element.setAttribute(
+				'style',
+				`--box: 20px; --gap: 2px; width: calc(5 * var(--box) + 6 * var(--gap)); height: calc(var(--gap) + var(--gap));`
+			);
+
+			const computedStyle = new CSSComputedStyle(element);
+			const propertyManager = computedStyle.getComputedStyle();
+
+			expect(propertyManager.get('width')?.value).toBe('calc(5 * 20px + 6 * 2px)');
+			expect(propertyManager.get('height')?.value).toBe('calc(2px + 2px)');
+		});
+
+		it('Resolves var() with fallback after an earlier var() in a shorthand value.', () => {
+			document.body.appendChild(element);
+			element.setAttribute('style', `--x: 16px; margin: var(--x) var(--x, 4px) var(--y, 8px);`);
+
+			const computedStyle = new CSSComputedStyle(element);
+			const propertyManager = computedStyle.getComputedStyle();
+
+			expect(propertyManager.get('margin-top')?.value).toBe('16px');
+			expect(propertyManager.get('margin-right')?.value).toBe('16px');
+			expect(propertyManager.get('margin-bottom')?.value).toBe('8px');
+		});
+
 		it('Ignores invalid CSS variable fallback value.', () => {
 			document.body.appendChild(element);
 			element.setAttribute('style', `width: var(--my-width, invalid);`);


capricorn86 (approved):
Thank you for your contribution @klaesra!

PR #2408, opened 2026-09-10
fix: [#2407] End comments at the first comment end tag when it overlaps a comment start tag

### LINKED ISSUE #2407: Comment end tag is not detected when it overlaps a comment start tag
**Describe the bug**

If the data of a comment ends with a comment open sequence that overlaps the close sequence, the parser never terminates the comment and the rest of the document is silently dropped. This breaks the downlevel-revealed conditional comments used in HTML email markup (see below).

**To Reproduce**

```js

const window = new Window();
window.document.write('<!--[if !mso]><!--><div>hello</div>');

console.log(window.document.querySelectorAll('div').length);
// 0
console.log(window.document.documentElement.outerHTML);
// "<html><head></head><body></body></html>"
```

In a browser the div is parsed:

```js
new DOMParser()
      .parseFromString('<!--[if !mso]><!--><div>hello</div>', 'text/html')
      .querySelectorAll('div').length;
// 1
```

**Expected behavior**

The comment ends at the first close sequence, followed by a live div:

```html
<!--[if !mso]><!--><div>hello</div>
parses as: comment with data "[if !mso]><!", then <div>hello</div>
```

Per the tokenizer spec, a comment open sequence inside comment data leads to the [comment less-than sign bang dash dash state](https://html.spec.whatwg.org/multipage/parsing.html#comment-less-than-sign-bang-dash-dash-state), which reconsumes the greater-than sign in the [comment end state](https://html.spec.whatwg.org/multipage/parsing.html#comment-end-state) and emits the comment.

The cause seems to be that the markup regexp in HTMLParser matches the inner open sequence as a new comment opener while in the comment read state, skipping past the two hyphens that belong to the closer.

**Device and details:**

- OS: macOS
- Node version: 24.15.0
- Package version: 20.14.3

**Additional context**

I have a small fix with tests and will open a PR 🙂

hampustagerud:
### Description

Comments were never terminated when the comment end tag overlapped a comment start tag (as in the downlevel-revealed conditional comments used in HTML email markup), which silently dropped the rest of the document. The markup regexp matched the inner comment start tag while in the comment read state, moving the scan position past the two hyphens that belonged to the end tag. The parser now resumes the scan at those hyphens, so the end tag is detected and the comment is terminated where a browser terminates it.

Tests cover the conditional comment case from the issue, the alternative end tag, and that comments still end at the first end tag when a comment start tag appears earlier in the data.

AI disclosure: I used Claude Code to help validate the fix and the tests. I have reviewed the changes and verified the behavior against the HTML specification and a real browser.

Resolves #2407

### Tests

### AI tools

### DIFF
--- a/packages/happy-dom/src/html-parser/HTMLParser.ts
+++ b/packages/happy-dom/src/html-parser/HTMLParser.ts
@@ -292,6 +292,9 @@ export default class HTMLParser {
 
 					if (match[4]) {
 						this.parseComment(html.substring(this.startTagIndex, match.index));
+					} else if (match[3]) {
+						// The "--" can be part of the end tag (e.g. "<!-->").
+						this.markupRegExp.lastIndex -= 2;
 					}
 					break;
 				case MarkupReadStateEnum.documentType:
--- a/packages/happy-dom/test/html-parser/HTMLParser.test.ts
+++ b/packages/happy-dom/test/html-parser/HTMLParser.test.ts
@@ -123,6 +123,42 @@ describe('HTMLParser', () => {
 			expect((<HTMLElement>result.childNodes[0]).id).toBe('æøåÆØÅ');
 		});
 
+		it('Parses comments containing "<!--" in their data.', () => {
+			const result = new HTMLParser(window).parse(
+				'<!--[if !mso]><!--><div>hello</div><!--<![endif]-->'
+			);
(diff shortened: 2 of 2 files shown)


capricorn86 (approved):
Thank you for your contribution!

PR #2418, opened 2026-09-14
fix: [#2396] Ignore click() while the element's click event is being dispatched

### LINKED ISSUE #2396: HTMLElement.click() is missing the spec's click-in-progress flag — re-entrant click() recurses to "Maximum call stack size exceeded"
**Describe the bug**

`HTMLElement.click()` dispatches its click event unconditionally — it does not implement the HTML spec's *click in progress flag* ([HTML spec, `click()` steps](https://html.spec.whatwg.org/multipage/interaction.html#dom-click)), which makes a re-entrant `click()` during the element's own click dispatch a no-op in real browsers. As a result, the very common "dropzone" pattern — a hidden file input inside a container whose click handler calls `input.click()` — recurses container → input → container until `RangeError: Maximum call stack size exceeded`.

With the default `errorCapture: 'tryAndCatch'` the overflow is converted into a window `error` event and a `console.error`, so under a test runner it silently prints `[RangeError: Maximum call stack size exceeded]` on stderr while the test **passes** — which is how we found it (vitest + `@testing-library/user-event` `upload()` against a React dropzone).

**To Reproduce**

```js

const window = new Window()
document.body.innerHTML = `<div id="zone"><input id="file" type="file"></div>`

const zone = document.getElementById('zone')
const input = document.getElementById('file')

// The common dropzone pattern: a click anywhere in the zone opens the picker.
zone.addEventListener('click', () => input.click())

input.click()
```

Observed (instrumented with a run counter):

- happy-dom 18.0.1: listener runs ~1,200 times, then `RangeError: Maximum call stack size exceeded` (thrown with `errorCapture: 'disabled'`; captured into a window `error` event with the default `'tryAndCatch'`).
- happy-dom 20.14.0: same.
- Chromium (same DOM and listener, served over HTTP): listener runs **exactly once**, nothing thrown.

**Expected behavior**

Per the spec's `click()` steps — step 2 is "If this element's click in progress flag is set, then return." (WHATWG HTML §6.4.2) — the nested `click()` issued from inside the input's own click dispatch should be a no-op, so the listener runs once, as it does in Chromium and Firefox. `click()` on master currently has no such guard:

https://github.com/capricorn86/happy-dom/blob/master/packages/happy-dom/src/nodes/html-element/HTMLElement.ts

A per-element boolean (set before dispatch, cleared after, early-return when already set) would restore the browser behaviour.

**Device and details:**

- OS: Windows 11
- Node version: 24.18.0
- Package version: 18.0.1 and 20.14.0 (both reproduce; the `click()` on master is unguarded too)

**Additional context**

Downstream shape of the failure, in case others search for it: a vitest suite using `@testing-library/user-event`'s `upload()` on a React dropzone (`<div onClick={() => inputRef.current?.click()}><input type="file" …/></div>`) prints

```
stderr | …
[RangeError: Maximum call stack size exceeded]
```

against a green test, because the overflow lands in an event listener and `'tryAndCatch'` swallows it. Workaround we applied: guard the container handler with `if (event.target !== inputRef.current) inputRef.current?.click()`.

mixelburg:
### Description

`HTMLElement.click()` dispatches its event unconditionally, so a `click()` issued from inside the element's own click dispatch recurses until the stack overflows. The dropzone pattern (`zone.addEventListener('click', () => input.click())`, triggered by `input.click()`) hits `RangeError: Maximum call stack size exceeded` — silently, since the default `errorCapture: 'tryAndCatch'` turns it into a window error event while the test still passes.

The spec's `click()` steps return early when the element's click in progress flag is set, which is what browsers do here (the listener runs once). The flag is now set before dispatching and cleared in a `finally` block.

Resolves #2396

- [x] Read the contributing guidelines.

### Tests

- [x] Added tests for re-entrant `click()`, the dropzone pattern, and a subsequent independent `click()`.
- [x] `npm test` in `packages/happy-dom` passes (7730 tests, 303 files).

### AI tools

I used an AI coding assistant to help write this change. I reviewed the diff and ran the tests locally.


mixelburg:
CI failed on the Node 24 leg in `Fetch > send() > Updates cache after a failed revalidation with a "If-Modified-Since" request for a GET response with "Cache-Control" set to a "max-age"` (`expected 'some text' to be 'some new text'`, `test/fetch/Fetch.test.ts:3977`). That test doesn't go near `HTMLElement.click()`, and it passes on this exact head outside CI:

- `npx vitest run test/fetch/Fetch.test.ts` → 103 passed on Node 22
- same file on Node 24.21.0 in a clean container → 103 passed

The test sleeps 300ms against a `max-age=0.0001` cache entry, so it looks timing-sensitive. Could the run be re-triggered? Happy to dig further if it reproduces.

mixelburg:
CI is green now — the re-run passed on all three Node legs (`build (20)`, `build (22)`, `build (24)`), so the earlier red was indeed the timing flake rather than anything in this change.

For the record it's a known one: the same `Fetch` cache-revalidation family is tracked in #2013 (open, "Cache revalidation race condition"), and #1985 was the `SyncFetch` variant fixed for one test only. Happy to send a separate PR stabilising the timing-sensitive ones if that's useful.

PR #2421, opened 2026-09-16
fix: [#2420] fix isInclusiveAncestor hang when an element defines its own .host property

### LINKED ISSUE #2420: Hangs indefinitely in isInclusiveAncestor when an element defines its own `host` property
`NodeUtility.isInclusiveAncestor()` walks up a node's ancestor chain to check whether one node contains another, and to avoid inserting a node into its own descendent. When it hits a node with no parentNode it falls back to reading the node's `.host` property, but as this isn't a reserved property name there's nothing stopping users or libraries (like Stencil) from defining their own value for `.host`. Case in point: Stencil-compiled web components declare an `@Element()` ref named `host` which points at itself. When `isInclusiveAncestor()` gets to this node we get stuck in a synchronous loop and our test hangs indefinitely.

**To Reproduce**
Steps to reproduce the behaviour:

1. Use a Stencil-compiled web component library
2. Point test-runner at happy-dom
3. Render a Stencil component in a test, e.g. document.body.appendChild(myStencilElement) or through a test-utils render helper.
4. Component triggers DOM mutation (e.g. appendChild or insertBefore), and as `includeShadowRoots` is set to true, isInclusiveAncestor looks for `(<ShadowRoot>parent).host` while the element is detached ( `parentNode===null`) but user/library has defined its own `host` prop on the element pointing at itself.
5. Test hangs indefinitely with cpu pinned at 100% as the isInclusiveAncestor loop is synchronous.  

[Stackblitz repro](https://stackblitz.com/edit/stackblitz-starters-uahgr2ly?file=index.js)

**Expected behavior**
`isInclusiveAncestor` should only follow the `.host` for genuine shadowRoot instances. An otherwise-defined `host` property should be ignored and the function should return normally instead of hanging.

sbays:
### Description

`isInclusiveAncestor()` walks up a node's ancestor chain. When it hits a node with no parentNode it falls back to reading the node's `.host` property, but as this isn't a reserved property name there's nothing stopping users or libraries (like Stencil) from defining their own value for `.host`. 

As an example, Stencil-compiled web components declare an `@Element()` ref named `host `which points at itself. When `isInclusiveAncestor()` gets to this node it gets stuck in a synchronous loop and our test hangs indefinitely with CPU at 100%.

Changed the fallback to read `PropertySymbol.host` instead of the public `.host` property (to match `PropertySymbol` use elsewhere in the function). Only genuine ShadowRoot instances have that symbol, so avoids the walk-up being hijacked by user-defined code.

Added tests to assert that user-defined self-referential .host doesn't loop and that real shadow-root host traversal still works.

Claude was utilised in isolating the original bug, when adding the fix & writing the tests. All verified by me and tested locally.

Resolves #2420 

### Tests

### Title

### AI tools


sbays:
Had to re-run CI as Fetch.test.ts cache-revalidation failure is a [known issue](https://github.com/capricorn86/happy-dom/issues/2013)  [with a fix in review](https://github.com/capricorn86/happy-dom/pull/2201/changes#diff-c050d5d142f91dd064874a9a4b4eb8a17e0891079498a6aa916d3a0425acea11) unrelated to this

PR #2427, opened 2026-09-20
fix: [#2353] Clone socket-backed Response bodies via stream tee

### LINKED ISSUE #2353: Response.clone() silently returns empty body for real socket responses
**Describe the bug**
Response.clone() silently returns an empty body when cloning a response from a real network request (socket-backed), while successfully cloning synthetic/in-memory responses.

**To Reproduce**

1. Setup a real HTTP server (or use a localhost echo server)
2. fetch with a json body
3. Test 1: Synthetic response (in-memory) ✅ 
```
const synthetic = new Response(JSON.stringify({ test: 'synthetic' }));
const clonedSynthetic = synthetic.clone();
await clonedSynthetic.json();
```
output : { test: 'synthetic' }

4. Test 2: Real network response ❌ 
```
const real = await fetch('http://localhost:3000/data.json');
const clonedReal = real.clone();
await clonedReal.json(); 
```
output: {} (empty)

**Actual Behavior**
Real socket responses: clone().body is empty
Synthetic responses: clone().body contains the full content
Original response's body also becomes empty after calling .clone() on a real socket response

**Expected Behavior**
Both should return the same content. Cloning a response should produce an independent readable copy regardless of whether the original response came from a real network socket or was constructed in-memory.

**Device and details:**
happy-dom: 20.12.0 (tested; also reproduced on earlier versions)
Node.js: 18.x+
Only affects responses from real network calls (node streams underneath)

**Additional context**
Dig the code and seems the issue is in `FetchBodyUtility.cloneBodyStream()` in Fetch.ts. When a Response is built from a real incoming socket, the implementation re-consumes the underlying Node stream:
1. Response construction (nodeToWebStream): A 'data' listener is attached to the raw Node stream, which switches it to flowing mode:
2. Data arrival: For small/local responses, all data arrives and drains within 1–2 event loop ticks. The stream ends.
3. Later: clone(): The code tries to pipe the same Node stream again to two new PassThroughs:
4. Node doesn't replay past 'data' events to late listeners, so both stream1 and stream2 are empty.
Result: Both the clone and the original end up with empty bodies.

Maybe we can use the .tee() on the already constructed WeReadableStream ? We did this on our side to patch it in a response.clone() polyfill.

Happy to test this fix or propose a PR if you want.

dyk1454683243-sudo:
### Description

**Fixes #2353**

`Response.clone()` on a socket-backed fetch response re-piped a Node stream that `nodeToWebStream()` had already put into flowing mode. Node does not replay past `data` events, so both the clone and the original body could become empty.

Stream-backed bodies are now cloned with `ReadableStream.tee()`, which splits queued and future bytes on the existing Web stream. Buffer-backed / synthetic `Response.clone()` is unchanged.

### Tests

- [x] Make sure to add tests for your changes.
- [x] Regression coverage: drained `nodeToWebStream` body, real Node HTTP `fetch` + `clone()`, and the existing delayed-chunk stream clone (now using the same construction as Fetch).
- [x] Ran `vitest` for `Response.test.ts`, `Request.test.ts`, and `Fetch.test.ts` (all green).
- [x] Ran the full `packages/happy-dom` suite: 7728 tests passed.

### Title

### AI tools

- [x] This change was authored with assistance from Cursor.


dyk1454683243-sudo:

PR #2429, opened 2026-09-20
fix: [#2310] Continue NodeIterator after current node is removed

### LINKED ISSUE #2310: NodeIterator stops iterating after the current node is removed from the document
When the node a `NodeIterator` currently references is removed from the
document, iteration ends: the next `nextNode()` returns `null`, even though
there are still nodes after it in document order. Per the DOM standard, a
`NodeIterator` has pre-removing steps that move its reference node so that
iteration continues from the right place, so removing the current node should
not end the walk.

This is what turns DOMPurify into a text extractor on happy-dom. Its
`KEEP_CONTENT` path hoists an unwanted element's children to just after that
element and then removes the element itself; on happy-dom the walk ends there, so
everything after the first removed element is never inspected. Consequences we
measured with `ALLOWED_TAGS` that excludes `div` and `KEEP_CONTENT: true`:

| input | happy-dom 20.8.9 | jsdom 28.1.0 and Chromium 151 |
|---|---|---|
| `<div class="a"><div class="b"><strong>x</strong></div></div>` | `<div class="b"><strong>x</strong></div>` | `<strong>x</strong>` |
| `<div class="o1">a</div><div class="o2">b</div>` | `a<div class="o2">b</div>` | `ab` |
| `<p>a<script>alert(1)</script>b</p>` | `a<script>alert(1)</script>b` | `<p>ab</p>` |

The third row is the one to look at: the sanitizer returns a live `<script>`
element. A single level of nesting does NOT show the problem, because "stops
after the first removed element" and "removes every disallowed tag" produce the
same output there.

Repro without DOMPurify, mirroring only what its `KEEP_CONTENT` branch does.
Versions: happy-dom 20.8.9, jsdom 28.1.0, Node 22.22.3.

```js

const document = new Window().document;
document.body.innerHTML =
  '<div id="outer"><div id="inner"><strong id="leaf">x</strong></div></div>';

const iterator = document.createNodeIterator(document.body, 1 /* SHOW_ELEMENT */);
const visited = [];
let node;
while ((node = iterator.nextNode())) {
  visited.push(node.id || node.nodeName);
  if (node.id === 'outer') {
    // What DOMPurify does for a disallowed element with KEEP_CONTENT: move the
    // children to just after the element, then remove the element.
    const children = node.childNodes;
    for (let i = children.length - 1; i >= 0; --i) {
      node.parentNode.insertBefore(children[i], node.nextSibling);
    }
    node.parentNode.removeChild(node);
  }
}
console.log(JSON.stringify(visited));
```

Actual (happy-dom 20.8.9):

```
["BODY","outer"]
```

Expected, and what both jsdom 28.1.0 and Chromium 151 produce:

```
["BODY","outer","inner","leaf"]
```

Note that the hoist direction matters for reproducing this. Inserting the
children BEFORE the element (into already-visited territory) ends the walk on
jsdom and in Chromium too, which is correct: those nodes are behind the
iterator. DOMPurify inserts them after the element, which is why a spec-compliant
`NodeIterator` reaches them.

dyk1454683243-sudo:
### Description

Fixes #2310.

`NodeIterator` is no longer a thin `TreeWalker` wrapper. It now keeps a spec `reference` / `pointerBeforeReferenceNode` and runs the DOM **pre-removing steps** when a node is removed (`removeChild` / `replaceChild`).

That matches jsdom and Chromium: if the current node is removed after its children are inserted *after* it, iteration continues to the still-in-document nodes that follow.

This is what turns DOMPurify `KEEP_CONTENT` into a text extractor on happy-dom today. The sanitizer hoists a disallowed element's children to just after that element and then removes the element; without pre-removing, the walk stops there, so later nodes are never inspected. With `ALLOWED_TAGS` that excludes `div`, that can leave a live `<script>` in the output (see the issue).

Focused change only — NodeIterator + document iterator registry + `removeChild` hook + regression tests. Not stacked on Response.clone (#2353/#2427) or MathML/#1629.

### Tests

Local verification: `packages/happy-dom` `tsc --noEmit` is clean; NodeIterator / TreeWalker / Document tests pass; full happy-dom vitest suite is green (7734 tests).

### Title

### AI tools

AI assistance: Cursor was used while implementing and testing this change. I reviewed the spec/jsdom pre-removing behavior and the resulting diff.


dyk1454683243-sudo:

Issue #1165, opened 2023-11-20
Expose properties on Node prototype to support DOMPurify anti-clobbering


luxaritas:
**Is your feature request related to a problem? Please describe.**
I am [currently working on](https://github.com/cure53/DOMPurify/pull/878) happy-dom support in DOMPurify. In order to protect against DOM clobbering, it uses the `Node` prototype in order to access the following:
* cloneNode
* nextSibling
* childNodes
* parentNode

In order to do this, all of these properties need to be either functions or getters (not raw properties) and cannot be overridden in child classes of `Node`.

**Describe the solution you'd like**
* Move full implementation of `cloneNode` into Node
* Implement `parentNode` as a getter
* Add unit tests to ensure the implementation of these functions maintains consistent
* Audit potential opportunities for DOM clobbering of node/node subclass properties. While it appears happy-dom has protections for this already, this would probably be a good opportunity to do a full audit of instances where properties are dynamically set and add unit tests specifically to protect against vulnerabilities of this kind.
* While not necessary for DOMPurify, it may be desirable for other ecosystem users to avoid overrides for all methods of Node (and subclasses), so that they can be referenced via prototype when specifically intending to avoid DOM clobbering.

**Describe alternatives you've considered**
* Removing prototype __lookupGetter__ usage in DOMPurify: This would be unsafe and against the point of the library.
* Specifically for parentNode, fall back to property when getter is not available. This however would weaken security guarantees in DOMPurify.
* No DOMPurify support

**Additional context**
I appreciate that this could require some nontrivial architecture changes. However, DOMPurify is an important security library in the ecosystem, and I believe support would be highly valuable.

Additional discussion is available in https://github.com/cure53/DOMPurify/pull/878

mcous:
This issue appears to be the cause of https://github.com/testing-library/svelte-testing-library/issues/319 and https://github.com/sveltejs/svelte/issues/10358.

The upcoming Svelte v5 relies on calling [various methods through `Node.prototype`](https://github.com/sveltejs/svelte/blob/6b7f1162adac0e9c86fa58a4fce4acff5843eb60/packages/svelte/src/internal/client/dom/operations.js#L53-L58) as well as `Element.prototype`. For example, this minimal snippet, based on Svelte's initial mounting code, doesn't work in Happy DOM:

```ts
import { Window } from "happy-dom";

const window = new Window({ url: "http://localhost:3000" });
const document = window.document;
[...]

PR #2430, opened 2026-09-20
fix: [#2182] Node.prototype.nodeName returns real name for subclass instances

### LINKED ISSUE #2182: `Node.prototype.nodeName` getter returns `''` when invoked on an element (breaks DOMPurify ≥ 3.4.8 anti-clobbering)

dyk1454683243-sudo:
### Description

**Fixes #2182**

`Node.prototype.nodeName` always returned `''`. Normal access (`el.nodeName`) still worked because `Element` (and other subclasses) override the getter, but libraries that extract the un-clobberable prototype getter do not see those overrides:

```js
Object.getOwnPropertyDescriptor(Node.prototype, 'nodeName').get.call(element)
// was '' — expected 'P' for a <p>
```

DOMPurify ≥ 3.4.8 reads tag names this way as part of anti-clobbering (GHSA-x4vx-rjvf-j5p4 / GHSA-4w3q-35jp-p934). Under happy-dom the empty name makes allow/block logic misfire (allowed tags dropped, XSS vectors retained). This continues the #1165 anti-clobbering work.

`Node.prototype.nodeName` now dispatches on `nodeType` (WHATWG `nodeName`):

- Element → `tagName`
- Attr → qualified name
- Text → `#text`
- CDATA → `#cdata-section`
- ProcessingInstruction → `target`
- Comment → `#comment`
- Document → `#document`
- DocumentType → `name`
- DocumentFragment → `#document-fragment`

Subclass getters are left in place. Prior attempt #2183 sketched this approach and was closed without merge; this is a fresh PR from current `master`.

### Tests

- [x] Regression test extracts `Node.prototype.nodeName` and `.call()`s it on Element (`<p>` → `P`), Text, Comment, Document, DocumentFragment, DocumentType, ProcessingInstruction, and Attr. Normal `el.nodeName` stays green.
- [x] Ran Node / Element / Text / Comment / Document / DocumentFragment tests: 669 passed.
- [x] Ran the full `packages/happy-dom` suite: 7727 tests passed.
- [x] Minimal issue repro against compiled `lib`: `get.call(el) === 'P' && el.nodeName === 'P'`.

### Title

### AI tools

- [x] This change was authored with assistance from Cursor.


dyk1454683243-sudo:

